PR #2, opened 2013-12-06
Add login styles

carpeliam:
This has two separate commits; the first one adds styles to login-related pages, the second one adds a CSS class to the patient builder to keep its changes from affecting other pages. If we decide not to handle the patient builder this way, that commit is pretty isolated and is easy to get rid of, however we will need to handle bleed-through in some way, and I'd recommend doing it this way at least temporarily, if not long term.


andrequina:
are you looking for review of this?  If so maybe Pete should take a look since there is a patient builder related change.

carpeliam:
Yep, I just didn't want to mess up the patient builder. I'm also thinking that a pull request format with peer review might be a good practice for us.

PR #3, opened 2013-12-06
Develop

rdingwell:
Added rake task to update patient source data criteria with title and description from measure source data criteria.

### DIFF
--- a/lib/tasks/bonnie.rake
+++ b/lib/tasks/bonnie.rake
@@ -107,6 +107,26 @@ namespace :bonnie do
 
   namespace :patients do
 
+    desc "Updated source_data_criteria to include title and description from measure(s)"
+    task :update_source_data_criteria=> :environment do
+
+      Record.where({}).each do |patient|
+        measures = Measure.where({measure_id: {"$in" => patient.measure_ids}})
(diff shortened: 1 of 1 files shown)


pkmitre:
I think that this task should be written to run after using the new DB reset task (rake bonnie:db:reset), which changes record measure_ids from NQF ids to hqmf_set_ids, so

```
measures = Measure.where({measure_id: {"$in" => patient.measure_ids}})
```

should probably be

```
measures = Measure.where({hqmf_set_id: {"$in" => patient.measure_ids}})
```

Possibly of interest, I think that the "in" above can be simplified a bit using Mongoid, so:

```
measures = Measure.in(hqmf_set_id: patient.measure_ids)
```

Also, 

```
Record.where({}).each do |patient|
```

can be written as just

```
Record.each do |patient|
```

rdingwell:
So the patient records are being associated to the version neutral measure id's?

pkmitre:
That's right, that way when a measure developer uploads a newer version of the same measure they keep the existing patient associations.

rdingwell:
Allright.  I'm not sure I agree with that but if Andre is on board ok.   I just think that there will probably be some issues with that approach in practice for updated valuesets and redefinition of data criteria.

PR #4, opened 2013-12-06
Tweaked styling and patient builder views for measures with 1 population

jtferns:
I tweaked the builder.less styling to match Mockups v3.02 for the Expected Values section of the patient builder view.

I also addressed issues with the population tab when building a patient from a measure with just 1 population, both in the Expected Values and the NQF sections.

I updated the Expected Value selections to read from Thorax.Models.Measure.allPopulationCodes to retain an order across all patient building, and I updated the text to display readable text based on the mockups.


pkmitre:
It turns out there's an issue with this pull request: a test fails, plus it seems like expected values don't correctly save

PR #5, opened 2013-12-09
Patient builder expected values update + bugfixes

jtferns:
- https://www.pivotaltracker.com/story/show/61762566
  - Removed single tab when rendering patient builder from a measure with one population
  - Removed population sub-id in NQF logic view when rendering patient builder from a measure with one population
- https://www.pivotaltracker.com/story/show/61306338
  - Updated expected value view/template to show ordered, readable population selections with some preliminary styling
- https://www.pivotaltracker.com/story/show/61306678
  - Removed standing divs from dashboard and individual measure views for upload/update dialogs


app/assets/javascripts/templates/patient_builder/expected_value.hbs
```diff
@@ -1,9 +1,10 @@
 <div class="expected-values">
+  <h2 for="fname">EXPECTED VALUE</h2>
```
carpeliam: can we remove the `for` attribute?

jtferns: I pulled it from the expected_values.hbs template, and I don't think it is required for any specific functionality, so I will go ahead and remove it

app/assets/stylesheets/builder.less
```diff
@@ -130,7 +130,11 @@
       .clearfix;
     }
     .sub-measure-tabs {
+      padding-top: 15px;
       padding-left: 15px;
+      .make-sm-column(12);
+      .make-md-column(12);
```
carpeliam, line 136: If you make an `sm` column, you shouldn't need to create an `md` or `lg` column. You also might be able to avoid doing this altogether if it's going to take up all 12 columns. We can take a look at this when we clean up the builder stylesheets, though.

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -19,7 +19,7 @@ class Thorax.Views.PatientBuilder extends Thorax.View
       measure: @measure
       edit: true
       values: _.extend({}, @model.get('expected_values'))
-    @populationLogicView = new Thorax.Views.BuilderPopulationLogic
+    @populationLogicView = new Thorax.Views.BuilderPopulationLogic(measure: @measure)
```
carpeliam: Models should be able to reference each other, rather than passing references through the views. A population has a reference to its measure already through `population.collection.parent`.

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -262,14 +263,30 @@ class Thorax.Views.ExpectedValueView extends Thorax.View
...
+      IPP: 'INITIAL PATIENT POP.'
+      DENOM: 'DENOMINATOR'
+      NUMER: 'NUMERATOR'
+      DENEXCEP: 'EXCEPTION'
+      DENEX: 'EXCLUSION'
+      MSRPOPL: 'MSRPOPL'
+      OBSERV: 'OBSERVATION'
+    @currentCriteria = {}
+    cIndex = 0;
+    for mc in matchingCriteria
```
carpeliam, line 276: I mentioned in a separate comment that you can probably get rid of `index`, but in future cases, CoffeeScript loops take an optional second argument for the index. You can remove `cIndex` altogether, and this becomes:

```
for mc, i in matchingCriteria
  @currentCriteria[mc] =
    key: mc
    displayName: criteriaMap[mc]
    index: i     
```

app/assets/javascripts/templates/patient_builder/expected_value.hbs
```diff
@@ -1,9 +1,10 @@
...
   <form class="form-inline" role="form">
     {{#each currentCriteria}}
-      <div class="checkbox">
-        <label>
-          <input type="checkbox" id="{{this}}"> {{this}}
+      <div class="checkbox {{#if index}}checkbox-divider{{/if}}">
```
carpeliam: It looks like index is equal to the index of this item in `currentCriteria`. Handlebars gives you `@index`, which is available within an `each` block, which you can use instead.

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -262,14 +263,30 @@ class Thorax.Views.ExpectedValueView extends Thorax.View
   initialize: ->
     c = @criteria
     p = @population
-    @currentCriteria = (pop for pop in Object.keys(c) when pop in Object.keys(p))
+    matchingCriteria = (criteria for criteria in Thorax.Models.Measure.allPopulationCodes when criteria in _.keys(p))
+    criteriaMap =
```
carpeliam, line 267: It looks like we might be doing this kind of mapping in a few places. And if `Thorax.Models.Measure.allPopulationCodes` ever changed, we'd have to change these places too. We should probably investigate some way to centralize this kind of information at some point.

carpeliam:
Hi @jtferns, I see you've been busy again :) And this time, on the weekend! I hope you weren't working the whole time. I left some comments about some minor things, nothing too major; I haven't had a chance to test, but it looks like this accomplishes what you were setting out to do in these stories. Either I or somebody else will be able to test this in the morning.

app/assets/javascripts/templates/patient_builder/expected_value.hbs
pkmitre: Does @index work within an if?

jtferns: My understanding is that it does work within an <code>#if</code>, but only if you are using the <code>#each</code> helper with an array; when I switched the data that I'm iterating over to Objects, **@index** did not appear to work as expected.

carpeliam: Ah, I see, I missed that earlier. In that case, you'd need to convert it
into an array in order to accomplish the ordering, as [JavaScript doesn't
guarantee key order in an object](http://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order).

app/assets/stylesheets/builder.less
jtferns, line 136: I'm a bit too unfamiliar with the current styling of the patient builder, and I didn't want to tweak it too much sans guidance, so I made simple changes to get it to match the mockup. When I remove the column declarations under <code>.sub-measure-tabs</code> or <code>.expected-value</code> I get odd behavior so I left them unchanged for now.

app/assets/javascripts/views/patient_builder.js.coffee
jtferns: This makes sense; I updated the view to no longer pass the measure reference but rather pull it down from the population model instead.

app/assets/javascripts/views/patient_builder.js.coffee
jtferns, line 276: Ah yes, that would be a more elegant solution. I'm removing this because I switched to arrays of objects in the <code>#each</code> comments below.

app/assets/javascripts/templates/patient_builder/expected_value.hbs
jtferns: Using an array of objects worked, I can grab the index within the helper as expected.

PR #6, opened 2013-12-09
updated rationale fixing for specifics to handle multiple parents and to...

andrequina:
... handle derived data criteria

### DIFF
--- a/app/assets/javascripts/models/result.js.coffee
+++ b/app/assets/javascripts/models/result.js.coffee
@@ -15,7 +15,7 @@ class Thorax.Models.Result extends Thorax.Model
         occurrences = @getOccurrences(@measure.get('population_criteria')[code]) # FIXME can we get these from the population instead
         # get the good and bad specifics
         occurrenceResults = @checkSpecificsForRationale(specifics, occurrences, @measure.get('data_criteria'))
-        parentMap = @buildParentMap(@measure.get('population_criteria')[code])
+        parentMap = @buildParentMap(@measure.get('population_criteria')[code], @measure.get('data_criteria'))
 
         # check each bad occurrence and remove highlights marking true
         for badOccurrence in occurrenceResults.bad
@@ -54,6 +54,10 @@ class Thorax.Models.Result extends Thorax.Model
       else
         if child.specific_occurrence
           occurrences.push child.key
+      if (child.temporal_references?.length > 0)
+        for temporal_reference in child.temporal_references
+          dataCriteria = @measure.get('data_criteria')[temporal_reference.reference]
+          occurrences = occurrences.concat @getOccurrences(dataCriteria)
     return occurrences
 
   # get good and bad specific occurrences referenced in this part of the measure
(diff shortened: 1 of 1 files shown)


andrequina:
Essentially this updates the specific occurrence rationale fixing code to deal with a few scenarios that were not originally covered.  The first scenario is multiple parents when walking up the tree.  In 0435: Discharged on Antithrombotic Therapy, we have an encounter (OccurrenceAInpatientEncounter1_precondition_73) that has 2 parents in the tree (both of the stroke diagnoses). The second scenario can be seen in the first population of 0004: Initiation and Engagement of Alcohol and Other Drug Dependence Treatment.  Occurrence B: Diagnosis, Active: Alcohol and Drug Dependence starts during multiple encounters that are specific occurrences.  If one of the child specific occurrences is bad the logic has to walk up the tree through the derived data criteria.

app/assets/javascripts/models/result.js.coffee
```diff
@@ -15,7 +15,7 @@ class Thorax.Models.Result extends Thorax.Model
         occurrences = @getOccurrences(@measure.get('population_criteria')[code]) # FIXME can we get these from the population instead
         # get the good and bad specifics
         occurrenceResults = @checkSpecificsForRationale(specifics, occurrences, @measure.get('data_criteria'))
-        parentMap = @buildParentMap(@measure.get('population_criteria')[code])
+        parentMap = @buildParentMap(@measure.get('population_criteria')[code], @measure.get('data_criteria'))
```
pkmitre: Doesn't buildParentMap() know about @measure already? Why do we need to pass in a piece of it, rather than referencing it from within the function? Could this be simplified even further by just passing in "code"?

app/assets/javascripts/models/result.js.coffee
```diff
@@ -75,29 +79,58 @@ class Thorax.Models.Result extends Thorax.Model
...
+  updateLogicTreeChildren: (updatedRationale, rationale, code, parents, orCounts, parentMap) ->
+    return unless parents
+    for parent in parents
+      if parent.id? 
+        parentKey = "precondition_#{parent.id}"
+      else
+        if parent.key?
+          parentKey = parent.key
+        else
+          parentKey = parent.type
```
pkmitre: This nested if could be written as

```
if parent.id? 
  parentKey = "precondition_#{parent.id}"
else if parent.key?
  parentKey = parent.key
else
  parentKey = parent.type
```

app/assets/javascripts/models/result.js.coffee
```diff
@@ -75,29 +79,58 @@ class Thorax.Models.Result extends Thorax.Model
...
       negated = parent.negation
-      parent = null
       if updatedRationale[code][parentKey] != false && !negated
         # if this is an OR then remove a true increment since it's a bad true
         orCounts[parentKey]-- if orCounts[parentKey]?
         # if we're either an AND or we're an OR and the count is zero then switch to false and move up the tree
-        if ((!orCounts[parentKey]? || orCounts[parentKey] == 0) && rationale[parentKey])
-          updatedRationale[code][parentKey] = false
-          parent = parentMap[parentKey]
+        if ((!orCounts[parentKey]? || orCounts[parentKey] == 0) && (!!rationale[parentKey] == true || rationale[parentKey] == undefined))
```
pkmitre, line 95: This is a bit hard to parse; would

```
if ((!orCounts[parentKey]? || orCounts[parentKey] == 0) && (rationale[parentKey]) || !rationale[parentKey]?)
```

be any clearer?

app/assets/javascripts/models/result.js.coffee
carpeliam: or even `parentKey = if parent.id? then "precondition_#{parent.id}" else parent.key || parent.type`, if you want to get really terse. But I do personally tend to like putting the `parentKey =` on the outside of the `if`.

andrequina:
updated based on comments

pkmitre:
Merged via cherry-pick to simplify history; sent out suggested pull-request process to team

PR #9, opened 2013-12-11
added ability to drop a data criteria on another in the patient builder ...

andrequina:
...to set the start and end time

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
+++ b/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
@@ -91,7 +91,7 @@
   <div class="time-line">
     <div class="timeline-title">Patient History Relative to Measures</div>
   </div>
-  <div class="record droppable">
+  <div class="record patient-data-list">
     {{view editCriteriaCollectionView}}
   </div>
   <div class="measure-view-side">
--- a/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
+++ b/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
@@ -91,7 +91,7 @@
   <div class="time-line">
     <div class="timeline-title">Patient History Relative to Measures</div>
   </div>
-  <div class="record droppable">
+  <div class="record">
     {{view editCriteriaCollectionView}}
   </div>
   <div class="measure-view-side">
--- a/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
+++ b/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
@@ -91,7 +91,7 @@
   <div class="time-line">
     <div class="timeline-title">Patient History Relative to Measures</div>
   </div>
-  <div class="record">
+  <div class="record droppable">
     {{view editCriteriaCollectionView}}
   </div>
   <div class="measure-view-side">
--- a/app/assets/javascripts/views/patient_builder.js.coffee
+++ b/app/assets/javascripts/views/patient_builder.js.coffee
@@ -39,9 +39,12 @@ class Thorax.Views.PatientBuilder extends Thorax.View
   events:
     rendered: ->
       @$('.draggable').draggable revert: 'invalid', helper: 'clone', zIndex: 10
-      @$('.droppable').droppable accept: '.ui-draggable'
-      # TODO move event handling up into events object, if possible
-      @$('.droppable').on 'drop', _.bind(@drop, this)
+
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/templates/patient_builder/patient_builder.hbs
```diff
@@ -91,7 +91,7 @@
   <div class="time-line">
     <div class="timeline-title">Patient History Relative to Measures</div>
   </div>
-  <div class="record droppable">
+  <div class="record">
```
pkmitre: Why not keep "droppable"? It's semantically clearer.

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -39,9 +39,12 @@ class Thorax.Views.PatientBuilder extends Thorax.View
...
     rendered: ->
       @$('.draggable').draggable revert: 'invalid', helper: 'clone', zIndex: 10
-      @$('.droppable').droppable accept: '.ui-draggable'
-      # TODO move event handling up into events object, if possible
-      @$('.droppable').on 'drop', _.bind(@drop, this)
+
+      @$('.record').droppable greedy: true, accept: '.ui-draggable', drop: _.bind(@drop, this)
+      @$('.record').on 'drop', '.patient-data', _.bind(@drop, this)
+
+      # @$('.record').bind('DOMNodeInserted', @registerPatientDataCriteria)
```
pkmitre: We can take out this commented line of code.

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -70,9 +73,15 @@ class Thorax.Views.PatientBuilder extends Thorax.View
     @expectedValuesView.serialize(children: false)
 
   drop: (e, ui) ->
-    measureDataCriteria = $(ui.draggable).model()
-    @model.get('source_data_criteria').add measureDataCriteria.toPatientDataCriteria()
+    console.log(e.target)
```
pkmitre: Remove debugging code like console.log

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -70,9 +73,15 @@ class Thorax.Views.PatientBuilder extends Thorax.View
...
+    patientDataCriteria = $(ui.draggable).model().toPatientDataCriteria()
+    dropTargetModel = $(e.target).model()
+    if dropTargetModel instanceof Thorax.Models.PatientDataCriteria
+      patientDataCriteria.set('start_date', dropTargetModel.get('start_date'))
+      patientDataCriteria.set('end_date', dropTargetModel.get('end_date'))
+    @model.get('source_data_criteria').add patientDataCriteria
     @materialize()
+    false
```
pkmitre, line 83: Without this false, does the event get handled twice?

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -146,6 +155,9 @@ class Thorax.Views.EditCriteriaView extends Thorax.View
     serialize: (attr) ->
       attr.start_date = moment(attr.start_date, 'L LT').format('X') * 1000 if attr.start_date
       attr.end_date = moment(attr.end_date, 'L LT').format('X') * 1000 if attr.end_date
+    rendered: ->
+      @$el.find('.patient-data').droppable greedy: true, accept: '.ui-draggable', hoverClass: 'drop-target-highlight'
```
pkmitre: Why @$el.find('.patient-data') vs @$('.patient-data')?

Also, mostly just curious, how is is the drop event getting bound for new elements as they get added? I assume it's up in the toplevel view, but I can't see what change is making it work. Could you walk me through it?

app/assets/javascripts/views/patient_builder.js.coffee
andrequina, line 83: yes, it will still bubble even though the drop is set to greedy... it will cause multiple instance of the dropped item to get added

The greedy appears to change the order to have the first event be the DC, but it still bubbles.  Without greedy the drop is handled in the order of the DOM.

app/assets/javascripts/views/patient_builder.js.coffee
andrequina: // make the parent droppable
      @$('.record').droppable greedy: true, accept: '.ui-draggable', drop: _.bind(@drop, this)

// make existing and new data criteria droppable, we are adding a .patient-data selector to the parent to make the children with .patient-data selectable
      @$('.record').on 'drop', '.patient-data', _.bind(@drop, this)

PR #10, opened 2013-12-11
adds negations to patient builder data criteria editing

andrequina:
- changes to view
  - changes to patient materiaization to fix string vs boolean

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
@@ -56,6 +56,21 @@
             </div>
           </div>
         </div>
+        <div class="form-group" style="margin: 15px 0px 15px 15px;">
+          <div class="checkbox">
+            <div style="margin: 0px 0px 0px 20px;"> 
+              <label>
+                <input type="checkbox" name="negation" class="negation-select checkbox" data-toggle="negation-select"> NOT PERFORMED
+              </label>
+            </div>
+          </div>  
+          <select name="negation_code_list_id" class="form-control negation-code-list {{#unless negation}}hide{{/unless}}">
+              <option value="">--</option>
+              {{#each codes tag="select" name="negation_code_list_id"}}
+                <option value="{{oid}}">{{display_name}}</option>
+              {{/each}}
+          </select>
+        </div>
         <div class="form-group minimize">
           {{#button "closeDetails" class="btn btn-primary pull-right"}}CLOSE{{/button}}
         </div>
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -139,6 +139,7 @@ class Thorax.Views.EditCriteriaView extends Thorax.View
     _(super).extend
       start_date: moment(@model.get('start_date')).format('L LT') if @model.get('start_date')
       end_date: moment(@model.get('end_date')).format('L LT') if @model.get('end_date')
+      codes: @measure.get('value_sets').map (vs) -> vs.toJSON()
```
pkmitre, line 142: "codes" seems like a bit of a generic variable name.

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -56,6 +56,21 @@
...
+          <div class="checkbox">
+            <div style="margin: 0px 0px 0px 20px;"> 
+              <label>
+                <input type="checkbox" name="negation" class="negation-select checkbox" data-toggle="negation-select"> NOT PERFORMED
+              </label>
+            </div>
+          </div>  
+          <select name="negation_code_list_id" class="form-control negation-code-list {{#unless negation}}hide{{/unless}}">
+              <option value="">--</option>
+              {{#each codes tag="select" name="negation_code_list_id"}}
```
pkmitre: It looks like you might have started this using the value_sets collection (hence the tag and name references in each), but switched to each because you needed a default value... so the tag and name can probably be removed?

lib/measures/patient_builder.rb
```diff
@@ -101,7 +101,7 @@ def self.check_data_criteria!(record, data_criteria)
     def self.derive_entry(data_criteria,value, value_sets)
       return nil if data_criteria.type == :characteristic && data_criteria.patient_api_function.nil? 
       time = derive_time_range(value)
-      data_criteria.negation = value["negation"] == "true"
+      data_criteria.negation = value["negation"] == "true" || value["negation"]
```
pkmitre: This code is ambiguous... if value["negation"] is set to "false" (which it might be for existing data?) this expression will still evaluate to true.

Would it make more sense to have the front end normalize the negation checkbox to true/false rather than have the backend try to be flexible?

app/assets/javascripts/views/patient_builder.js.coffee
andrequina, line 142: that's the same name used in the child views so it's at least consistent

PR #12, opened 2013-12-13
NQF Highlighting bugfix

jtferns:
**Story:** https://www.pivotaltracker.com/story/show/62175964

For measures, such as 0031, sometimes the logic highlighting throws a TypeError for undefined length.

I updated the highlighting logic in result.js.coffee to check if there are existing preconditions to recurse on.
### Error on patient-select

original_typeerror
### Post bugfix

post_bugfix

I am unsure if this affects the highlighting logic in a specific way, so that is the primary feedback I am looking for, etc.


pkmitre:
When closing a pull request without merging it, could we add a comment to indicate what's going on?

PR #13, opened 2013-12-13
Refactored expected values to be represented using Thorax models

pkmitre:


app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -223,70 +221,63 @@ class Thorax.Views.ExpectedValuesView extends Thorax.View
...
-    '0': ['unchecked', 'danger', 1]
-    '1': ['check', 'success', 0]
+
+  options:
+    populate: { context: true }
+
+  events:
+    serialize: (attr) ->
+      for pc in @model.populationCriteria()
+        attr[pc] = 1 if attr[pc] == true
```
carpeliam: Couldn't this be `attr[pc] = if attr[pc] then 1 else 0 for pc in @model.populationCriteria()`? Maybe two lines for readability, but the `== true` part should be trimmed here.

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -223,70 +221,63 @@ class Thorax.Views.ExpectedValuesView extends Thorax.View
...
+        attr[pc] = 0 unless attr[pc]
+
+  context: ->
+    context = super
+    for pc in @model.populationCriteria()
+      unless @measure.get('episode_of_care') || (@measure.get('continuous_variable') && pc == 'OBSERV')
```
carpeliam, line 275: Should this be `has` instead of `get`?

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -84,6 +86,20 @@ class Thorax.Models.Patient extends Thorax.Model
       @set _(data).chain().pick('conditions', 'encounters', 'medications', 'procedures').defaults(defaults).value(), silent: true
       @trigger 'materialize' # We use a new event rather than relying on 'change' because we don't want to automatically re-render everything
 
+  getExpectedValue: (population) ->
+    measure = population.collection.parent
+    expectedValue = @get('expected_values').findWhere(measure_id: measure.get('hqmf_set_id'), population_index: population.get('index'))
```
carpeliam: Index is available from the population's collection, and shouldn't need to be set on the population itself. I'm also unclear about how/why we're storing the population index, when it's not a natural property of the population... even its sub_id is generated. Could we use sub_id instead? It's more of a natural identifier.

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -84,6 +86,20 @@ class Thorax.Models.Patient extends Thorax.Model
...
+    expectedValue ?= new Thorax.Models.ExpectedValue measure_id: measure.get('hqmf_set_id'), population_index: population.get('index')
+    @get('expected_values').add expectedValue # Note: adding back an existing element is a no-op
```
carpeliam: Would the code be cleaner if you checked to see if expectedValue existed before running lines 92/93? Then, this comment wouldn't be necessary, as you'd never be adding an expected value that already exists.

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -223,70 +221,63 @@ class Thorax.Views.ExpectedValuesView extends Thorax.View
...
-            if checkedValue is true then pevHash[key] = 1 else pevHash[key] = 0
-        parsedValues[m.get('hqmf_set_id')][p.get('sub_id')] = pevHash
-      attr.expected_values = _.extend({}, parsedValues)
+      # We set the active tabs when rendered rather than trying to do it in the template
+      # because changes to the ExpectdValueView that happen when serializing (for materialization)
+      # drive the individual tabs to be re-rendered
+      @$('a[data-toggle="tab"]:first').tab('show')
+      @$('.tab-pane:first').addClass('active') # This seems to be necessary because we're not in the DOM yet?
+      # When the tabs are toggled, we want to send a message over to another view, use an event
+      @$el.on 'shown.bs.tab', 'a[data-toggle="tab"]', (e) =>
```
carpeliam, line 255: I don't think we should ever be using `@$el.on`. I'm pretty sure the only reason you'd ever need to do this is when you're listening to a DOM event that belongs to a child view's DOM. I think the DOM of different views should be siloed (they're not designed to bleed up to parent views); listen to the DOM event in its own view (should that be the ExpectedValueView in this case?), and then, if you need to handle that in the parent view, fire an event between child view and parent view, instead of between child DOM and parent view.

app/assets/javascripts/views/patient_builder.js.coffee
jtferns, line 275: I believe all measures will have a <code>continuous_variable</code> and <code>episode_of_care</code> attribute, so we use the <code>get</code> to evaluate if it is marked as <code>true</code>.  
Alternatively we could strip measures of these attributes if they are <code>false</code>, in which case <code>has</code> should work?

carpeliam, line 275: Ah, I see. I don't think stripping the measures of attributes would be the way to go, but I'm also not crazy about the way `if @model.get(...)` reads either... I'm kind of thinking that defining `isContinuousVariable()` and `isEpisodeOfCare()` methods on measure would read better, but I wouldn't make this a priority.

app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -223,70 +221,63 @@ class Thorax.Views.ExpectedValuesView extends Thorax.View
...
+    serialize: (attr) ->
+      for pc in @model.populationCriteria()
+        attr[pc] = 1 if attr[pc] == true
+        attr[pc] = 0 unless attr[pc]
+
+  context: ->
+    context = super
+    for pc in @model.populationCriteria()
+      unless @measure.get('episode_of_care') || (@measure.get('continuous_variable') && pc == 'OBSERV')
+        context[pc] = true if context[pc] == 1
```
carpeliam: I think we could eliminate the next line and just say `context[pc] = (context[pc] is 1)`. In general, if you're assigning something to true/false based on something else evaluating to true/false, I'd recommend evaluating the thing in a single line, instead of enumerating the truth case and the false case separately.

app/assets/javascripts/models/patient.js.coffee
pkmitre: The sub_id is intended to be used for display purposes, the index is a more natural data representation to indicate which population of a measure is being referred to; given that, it's useful for a population to know which index it is.

app/assets/javascripts/views/patient_builder.js.coffee
pkmitre, line 255: Hmm, not sure why it got rewritten that way, I think @$('a[data-toggle="tab"]').on 'shown.bs.tab' will work fine.

app/assets/javascripts/views/patient_builder.js.coffee
pkmitre: I'm afraid that won't work; the field can contain a boolean from a checkbox (true and false) for regular measures, but it can also contain a number (1, 2, 3, 4, 5) for Episode of Care measures which are truthy but should not be converted to 1.

I'll add a comment to that effect.

app/assets/javascripts/models/patient.js.coffee
pkmitre: Good call.

If Collection#add returned the value being added it would be simpler. Looks like Backbone.js makes that change for version 1.1, but I think we're stuck on 1.0 until Thorax updates.

app/assets/javascripts/models/patient.js.coffee
carpeliam: If the sub_id is only intended for display purposes, then we should keep it in the view. And, the population already has a mechanism for knowing which index it is. With this code, there are 3 ways at the model level to know; the sub_id, the index in the collection, and the population_index. The third can easily be replaced by the second. If you had an `index` method on population which called `@collection.indexOf(this)` I'd like that better than what we have here.

pkmitre:
I just tried using the events object again; like with the draggable, it doesn't look like thorax hooks the event properly

app/assets/javascripts/views/patient_builder.js.coffee
pkmitre, line 255: Turns out @$('a[data-toggle="tab"]').on 'shown.bs.tab' does not work, and the reason is because the sub-view renders again when the model changes, but the top-level does not, so the event doesn't get re-bound (as you originally surmised).

However, your suggestion of "listen to the DOM event in its own view" does _not_ work here, because that sub-view is an automatically generated Thorax collection itemView. We could instantiate that sub-view as a non-anonymous itemView, but that has two potential downsides

1) It's a fair amount of code for a small improvement

2) I've noticed that moving from an anonymous view to one represented by a class (whether it be a CollectionView or an itemView) can change how Thorax renders the DOM, which sometimes interacts in bad ways with Bootstrap.

given these, I'd like to stick with the current approach for now and possibly reconsider it in the future.

carpeliam, line 255: Ok, sounds good. I think this is something we should definitely consider in the future (after this file is refactored into separate files) - Backbone/Thorax goes to decent lengths to isolate views from each other, and for good reason. I'm kind of ok with spirit fouls (look! I just made an Ultimate frisbee analogy) if it's too expensive upfront to do otherwise, but usually I find that this makes the code harder to understand and harder to maintain long term.

I would like to see comments like #L46 removed as they're incorrect, but I'm ok with that being in a follow-up commit.

carpeliam:
Rebased, slightly squashed, and merged.

PR #26, opened 2013-12-23
replace oid with code_list_id

carpeliam:
This updates the patient data criteria to reference the `code_list_id` instead of `oid`.


pkmitre:
Is this the client side change referred to in #22? If so, the two should probably be merged at the same time, is that right?

PR #27, opened 2013-12-24
Deferred calculation

pkmitre:
This branch primarily focuses on deferred calculation, but also includes some major refactoring of existing code to make it easier to support the deferred calculation approach as well as some bugfixes that made more sense to apply on the branch code.


jtferns:
There wouldn't be anything in this pull request that would affect reloading patients from the database, right?

I had some issues with updating the source data criteria when running <code>rake bonnie:db:reset</code>, but I couldn't pinpoint the source of the issue, yet.

pkmitre:
This pull request shouldn't affect measure loading, but when switching to this branch it would make sense to clear out all caches... ie

```
rm -r tmp/cache/assets
```

and then reload the page using a cache-clearing reload.

PR #30, opened 2013-12-27
Total Coverage on Measure View

jtferns:
Story: https://www.pivotaltracker.com/story/show/61762256

Added simple total coverage percentage in aggregate results section of the individual measure view.

Current styling is a simple header-5 with a percent, dropped below the aggregate results views.

Current computation:
- iterate over the data criteria associated with the given measure
- store all of its keys in the <code>rationaleCoverage</code> hash
- iterate over each results' <code>rationale</code> hash
  - for each entry in the <code>rationale</code> hash, if it is truthy, then record a true value in the <code>rationaleCoverage</code> hash, otherwise if it is falsey and has not been marked true yet, then mark it as false
- the percentage is a simple computation of # of true values in the <code>rationaleCoverage</code> hash over the total number of values, rounded to a whole percentage

### DIFF
--- a/app/assets/javascripts/models/coverage.js.coffee
+++ b/app/assets/javascripts/models/coverage.js.coffee
@@ -1,0 +1,28 @@
+class Thorax.Model.Coverage extends Thorax.Model
+  initialize: (attrs, options) ->
+    @results = options.results
+    @measure = options.measure
+    @listenTo @results, 'change add reset destroy remove', @update
+    @update()
+  update: ->
+    completedResults = @results.select (d) -> d.isPopulated()
+    dataCriteria = @measure.get 'data_criteria'
+    rationaleCoverage = {}
+    for result in completedResults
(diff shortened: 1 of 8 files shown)


pkmitre:
There's a lot of figuring going on on in a view method. Isn't coverage a property of a population and a batch of results? Why not do this work in model space? I'd suggest adding a method on the Population model called "coverage" which returns a coverage model. It doesn't even need to take results as an argument, it can just call @calculationResults() internally.

You can then take out the connection between the coverage object and differences (which is somewhat artificial anyway, since coverage doesn't depend on the differences, it depends on the calculations) and just make the coverage object update when the results objects change.

If that's not efficient enough (ie too many updates), we can add a "complete" event on Thorax.Collections.Results (which will update based on it's results updating) and, since @calculationResults() returns one of those, just listen to that.

jtferns:
Ok, I will move this into the Population model following your suggestions

PR #40, opened 2014-01-02
finalize should clear the cache since we are setting the episode ids

andrequina:
episode ids are needed to generate the javascript

the finalize method sets the episode ids for the measure if it is episode of care.  We have already generated the javascript once on load, but we did not have the episode ids at that point.  We need to re-generate with the episode id specified by the user

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -44,7 +44,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/bonnie_bundler.git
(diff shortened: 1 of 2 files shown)


jtferns:
I'm going to merge this into develop since otherwise develop's reset-db task fails on measure js generation.

The application is functional and the jasmine tests are passing.

pkmitre:
Do we want to clear the cache on create in addition to on finalize?

andrequina:
Create deletes an existing measure if there is one... so there won't be cached javascript.

PR #41, opened 2014-01-02
fix coverage measure data critieria calculation

andrequina:
the measure data criteria calculation for coverage was using all data criteria from the measure.  This is not the correct approch.  Instead we want to only use the referenced data criteria for a population

### DIFF
--- a/app/assets/javascripts/models/population.js.coffee
+++ b/app/assets/javascripts/models/population.js.coffee
@@ -28,10 +28,9 @@ class Thorax.Models.Population extends Thorax.Model
 
   dataCriteriaKeys: ->
     return @populationDataCriteriaKeys if @populationDataCriteriaKeys
-    populations = @measure().get('population_criteria')
     criteriaKeys = []
-    for code in Thorax.Models.Measure.allPopulationCodes
-      criteriaKeys = criteriaKeys.concat @getDataCriteriaKeys(populations[code], false)
+    for code in @populationCriteria()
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/models/population.js.coffee
```diff
@@ -24,7 +24,39 @@ class Thorax.Models.Population extends Thorax.Model
...
+    new Thorax.Model.Coverage({}, population: this)
+
+  dataCriteriaKeys: ->
+    return @populationDataCriteriaKeys if @populationDataCriteriaKeys
+    populations = @measure().get('population_criteria')
+    criteriaKeys = []
+    for code in Thorax.Models.Measure.allPopulationCodes
+      criteriaKeys = criteriaKeys.concat @getDataCriteriaKeys(populations[code], false)
+    @populationDataCriteriaKeys = _.uniq(criteriaKeys)
+    @populationDataCriteriaKeys
```
pkmitre, line 35: This seems like it should be a method on a measure rather than a population?

andrequina, line 35: this should be population based, but I think the logic is wrong... it should be using the keys for the population rather than Thorax.Models.Measure.allPopulationCodes

I will make that fix.

jtferns:
I think the population-based approach makes sense for the desired coverage computation.

The branch has no functional issues and passes tests so I will go ahead and merge it.

PR #42, opened 2014-01-02
added icons to rationale view

andrequina:


app/assets/stylesheets/rationale.less
```diff
@@ -9,10 +9,26 @@
     background-color: @state-success-bg;
     color: @state-success-text;
   }
-  .false_eval, .bad_specifics {
+  .true_eval:before {
```
carpeliam: A better way to organize this would be:

``` less
.true_eval {
  background-color: @state-success-bg;
  color: @state-success-text;
  &:before {
    content: "\f00c";
    font-family: FontAwesome;
  }
}
```

This keeps all of the `:before` blocks with the class they're modifying.

PR #43, opened 2014-01-02
Field Values: Filtering and sorting

jtferns:
**Stories**
- https://www.pivotaltracker.com/story/show/62925160
- https://www.pivotaltracker.com/story/show/63158864

Current filtering strategy involves a hash keyed off of PatientDataCriteria <code>type</code>, added to the EditCriteriaValueView.
For now, all PatientDataCriteria will have field value options defined in the <code>globalInclusions</code> attribute, and all ecounter-based PatientDataCriteria will have ecounter-specific field value options also available.

Going forward, we will determine which field values should be long in the global list, and which ones belong in the specific PatientDataCriteria <code>type</code> list.

Additionally, if we can introduce <code>patient_api_function</code> values into the PatientDataCriteria model, we can key the inclusion hash off of those instead of <code>type</code>.

### DIFF
--- a/app/assets/javascripts/models/measure.js.coffee
+++ b/app/assets/javascripts/models/measure.js.coffee
@@ -26,6 +26,57 @@ class Thorax.Models.Measure extends Thorax.Model
   isPopulated: -> @has('data_criteria')
   populationCriteria: -> _.intersection(Thorax.Models.Measure.allPopulationCodes, _(@get('population_criteria')).keys())
 
+  @logicFieldsFor: (criteriaType) ->
+
+    # Define field values for all criteria types
+    globalInclusions = ['anatomical_structure', 'cumulative_medication_duration', 'dose', 
+    'frequency', 'incision_time', 'length_of_stay', 'ordinality', 'reason', 'removal_time', 
+    'route', 'severity', 'source', 'start_date', 'end_date']
+
+    # Define criteria type-specific field values
+    typeInclusions = 
+      care_goals: []
+      characteristics: []
+      communications: []
+      conditions: []
+      devices: []
+      diagnostic_studies: []
+      encounters: ['admit_time', 'discharge_time', 'discharge_disposition', 'facility', 
+        'facility_arrival', 'facility_departure', 'transfer_to', 'transfer_from']
+      functional_statuses: []
+      interventions: []
+      laboratory_tests: []
+      medications: []
+      patient_care_experiences: []
+      physical_exams: []
+      preferences: []
+      procedures: []
+      provider_care_experiences: []
+      provider_characteristics: []
+      risk_category_assessments: []
+      substances: []
+      symptoms: []
+      system_characteristics: []
+      transfers: []
+
+    # start with all field values
+    fields = Thorax.Models.Measure.logicFields
+
+    # grab all defined field values, merging global and type-specific specs
+    allInclusions = _(globalInclusions).union(typeInclusions[criteriaType])
+    allInclusions ?= []
+
+    # check any defined inclusions against the coded_entry_value of each field value
+    matchedInclusions = []
+    for field, attrs of fields
+      for inclusion in allInclusions when attrs['coded_entry_method'] is inclusion
+        matchedInclusions.push field
+    fields = _(fields).pick(matchedInclusions) if matchedInclusions.length > 0
+
+    # sort field values by title
+    fields = _.sortBy(fields, (f) -> f.title)
+    fields
+
 class Thorax.Collections.Measures extends Thorax.Collection
   url: '/measures'
   model: Thorax.Models.Measure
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/views/patient_builder.js.coffee
```diff
@@ -161,7 +161,12 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.Materializer
 
   initialize: ->
     @editValueView = new Thorax.Views.EditCriteriaValueView(model: new Thorax.Model, measure: @measure, fieldValue: false, values: @model.get('value'))
-    @editFieldValueView = new Thorax.Views.EditCriteriaValueView(model: new Thorax.Model, measure: @measure, fieldValue: true, values: @model.get('field_values'))
+    @editFieldValueView = new Thorax.Views.EditCriteriaValueView
+      model: new Thorax.Model
+      measure: @measure
+      fieldValue: true
+      values: @model.get('field_values')
+      modelType: @model.get('type')
```
pkmitre: Consider using "criteriaType" instead of "modelType" as the variable name.

pkmitre:
I think this would be cleaner if a lot of this code was moved into a model. Say, adding a function on the measure class (ie a class method, not an instance method) like Thorax.Models.Measure.logicFieldsFor(criteriaType). That way you could just go back to doing something like

```
context: ->
  _(super).extend
    ...
    fields: Thorax.Models.Measure.logicFieldsFor(@criteriaType)
```

app/assets/javascripts/models/measure.js.coffee
```diff
@@ -26,6 +26,57 @@ class Thorax.Models.Measure extends Thorax.Model
...
 
+  @logicFieldsFor: (criteriaType) ->
+
+    # Define field values for all criteria types
+    globalInclusions = ['anatomical_structure', 'cumulative_medication_duration', 'dose', 
+    'frequency', 'incision_time', 'length_of_stay', 'ordinality', 'reason', 'removal_time', 
+    'route', 'severity', 'source', 'start_date', 'end_date']
+
+    # Define criteria type-specific field values
+    typeInclusions =
```
jtferns, line 37: Is it sufficient to define these inclusions within the class method?

jtferns:
I can also see this logic belonging in the PatientDataCriteria model, as well the logic for figuring out default dates and/or durations?

app/assets/javascripts/models/measure.js.coffee
pkmitre, line 37: Makes sense to me, we can later expand the accessibility of this data if necessary.

app/assets/javascripts/models/measure.js.coffee
```diff
@@ -26,6 +26,57 @@ class Thorax.Models.Measure extends Thorax.Model
...
+      symptoms: []
+      system_characteristics: []
+      transfers: []
+
+    # start with all field values
+    fields = Thorax.Models.Measure.logicFields
+
+    # grab all defined field values, merging global and type-specific specs
+    allInclusions = _(globalInclusions).union(typeInclusions[criteriaType])
+    allInclusions ?= []
```
pkmitre, line 67: Is the "allInclusions ?= []" line necessary?

jtferns, line 67: Hmm, not really at this point.

If <code>globalInclusions</code> or <code>typeInclusions</code> is ever mis-populated, it would prevent that from breaking the view/initial functionality, but it might not be the best practice to keep it around.

app/assets/javascripts/models/measure.js.coffee
```diff
@@ -26,6 +26,57 @@ class Thorax.Models.Measure extends Thorax.Model
...
+
+    # check any defined inclusions against the coded_entry_value of each field value
+    matchedInclusions = []
+    for field, attrs of fields
+      for inclusion in allInclusions when attrs['coded_entry_method'] is inclusion
+        matchedInclusions.push field
+    fields = _(fields).pick(matchedInclusions) if matchedInclusions.length > 0
```
pkmitre, line 74: You can get the coded_entry_methods by doing

```
_(Thorax.Models.Measure.logicFields).chain().values().pluck('coded_entry_method').value()
```

Using something like that is there a simpler way of calculating the matching ones?

jtferns, line 74: Ah I see, that is convenient but we would still need to associate the <code>coded_entry_method</code> with the keys of <code>logicFields</code> somehow.

Some happen to be matching, but not all (e.g., 'admit_time' is keyed as 'ADMISSION_TIME').

jtferns, line 74: How about the following? It doesn't use the values via chain but it's a bit more compact.

```
# start with all field values
fields = Thorax.Models.Measure.logicFields

# grab all defined field values, merging global and type-specific specs
desiredInclusions = _(globalInclusions).union(typeInclusions[criteriaType])

# check any defined inclusions against the coded_entry_value of each field value
filteredKeys = (criteria for criteria, value of fields when value['coded_entry_method'] in desiredInclusions)
fields = _(fields).pick(filteredKeys)    

# sort field values by title
fields = _.sortBy(fields, (f) -> f.title)
fields
```

pkmitre, line 74: Looks good! Compact is good, but more importantly I think it's a bit easier to follow. Will merge shortly.

PR #49, opened 2014-01-08
Fix expand arrows on data criteria selector toggle bug

pkmitre:


carpeliam:
I'll be modifying the HTML here in https://github.com/carpeliam/bonnie-yeoman/ as there are invalid elements in the surrounding markup (e.g. `<div>` can't have an href attribute), I'll incorporate this change into that as well.

PR #50, opened 2014-01-08
Two-step delete (part 1) for Measure view

jtferns:
**Story: (ongoing)** https://www.pivotaltracker.com/story/show/62487888
- Added popover closing functionality on click anywhere but specific popover content or gear icon **(needs review)**
- Updated styling to match v5 mockup

If this is too early to review, please feel free to close this request until I can get to other two-step delete for the story (editable data criteria in the Patient Builder view).

popover_closed
popover_opened
popover_delete

### DIFF
--- a/app/assets/javascripts/views/measure_view.js.coffee
+++ b/app/assets/javascripts/views/measure_view.js.coffee
@@ -4,9 +4,8 @@ class Thorax.Views.Measure extends Thorax.View
     rendered: -> 
       @$("[rel='popover']").popover(content: @$('.popover-tmpl').text())
       $('html').click( (e) -> 
-        unless $(e.target).hasClass('popover-content')
+        unless $(e.target).hasClass('popover-content') or $(e.target).hasClass('show-delete')
           if $('#settings').has(e.target).length == 0 or $(e.target).is('.close') then $('#settings').popover('hide'))
-    'click .delete-measure': 'deleteMeasure'
 
   initialize: ->
     populations = @model.get 'populations'
@@ -32,6 +31,15 @@ class Thorax.Views.Measure extends Thorax.View
     importMeasureView.appendTo(@$el)
     importMeasureView.display()
 
+  showDelete: (e) ->
+    deleteButton = @$('.delete-measure')
+    deleteIcon = @$(e.target)
+    # if we clicked on the icon, grab the icon button instead
+    if deleteIcon[0].tagName is 'I' then deleteIcon = @$(deleteIcon[0].parentElement)
+    deleteIcon.toggle()
+    deleteButton.toggle()
+
   deleteMeasure: (e) ->
     @model = $(e.target).model()
     @model.destroy()
+    bonnie.renderMeasures()
--- a/app/assets/javascripts/views/measure_view.js.coffee
+++ b/app/assets/javascripts/views/measure_view.js.coffee
@@ -1,7 +1,11 @@
 class Thorax.Views.Measure extends Thorax.View
   template: JST['measure']
   events:
-    rendered: -> @$("[rel='popover']").popover(content: @$('.popover-tmpl').text())
+    rendered: -> 
+      @$("[rel='popover']").popover(content: @$('.popover-tmpl').text())
+      $('html').click( (e) -> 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -1,8 +1,12 @@
 class Thorax.Views.Measure extends Thorax.View
   template: JST['measure']
   events:
-    rendered: -> @$("[rel='popover']").popover(content: @$('.popover-tmpl').text())
-    'click .delete-measure': 'deleteMeasure'
+    rendered: -> 
+      @$("[rel='popover']").popover(content: @$('.popover-tmpl').text())
+      # if we click anywhere but the gear icon or the delete icon, close the popover
+      $('html').click( (e) =>
```
carpeliam, line 7: Instead of handling this only with the click event, I'd rather handle this with a blur event, for two reasons: making the site accessible; and keeping the code localized.

Also, for handling of events, is it possible to move this into the `events` hash, instead of having event code within the `rendered` event?

app/assets/stylesheets/measure.less
```diff
@@ -41,12 +41,44 @@
       font-weight: 700;
       font-size: 1.2em;
       .popover {
-        min-width: 187px;
+        min-width: 219px;
```
carpeliam, line 44: I see a lot of random numbers like 219, -87, 76, and 70%; are these numbers mathematically related to each other?

app/assets/javascripts/views/measure_view.js.coffee
jtferns, line 7: Hmm, I can't seem to get the blur event to work correctly within the hash; which element should I be listening to for a blur event?

I had not placed it in the events hash since the <code>$('html')</code> selector allows for clicking on any page element to close the popover. When moving into the events hash, the scope of elements is limited to those present within the template or the associated <code>$el</code>, which limits where the user can click to dismiss the popover (which for the measure view is just the title and description).

So far I have the following, with the limited clicking area:

```
events:
    'click': 'closePopover'
...
closePopover: (e) ->
    # if we click anywhere but the gear icon or the delete icon, close the popover
    unless @$(e.target).hasClass('popover-content') or @$(e.target).hasClass('show-delete')
      if @$('#settings').has(e.target).length == 0 and @$('.popover').is(':visible') then @$('#settings').popover('toggle')
```

app/assets/stylesheets/measure.less
jtferns, line 44: Not necessarily, I had tweaked the values to match the v5 look of the popover with the fewest possible changes. Is there a more appropriate way of determining margin or width values for elements on the page?

app/assets/javascripts/views/measure_view.js.coffee
carpeliam, line 7: What I was going for was to draw focus to the popover when you click on it, and then when that focus is blurred, to trigger the event, with the purpose of keeping all of the code related to this view within the view, and not touching outside elements. I'll try this out.

One of the issues with this as it stands now is that, when you browse to another view (or even re-render this one), that doesn't remove the click handler you've registered here, so this will end up trying to close a popover that doesn't exist anymore. It also might be hard to remove the click handler, seeing as how it's on the HTML element. So I'd really like to find a different approach here.

app/assets/stylesheets/measure.less
carpeliam, line 44: One way is to try to reuse something that's within bootstrap itself (referenced by an existing bootstrap variable, or by using the same class as a bootstrap class). Another way is to establish mathematical relationships where appropriate (e.g. "the margin is twice the padding" or whatever), so that, if you end up changing one thing, you don't need to change all of the others to accomodate. Yet another way, if neither of those other options are appropriate, is to use round numbers, which, in being less exact, can be less brittle.

app/assets/javascripts/views/measure_view.js.coffee
jtferns, line 7: Ah I see, that makes sense. I'll update this if I find anything more reasonable that works.

jtferns:
Will be redesigned and resubmitted, moving away from popover approach.

PR #51, opened 2014-01-08
Expand coverage of patient builder tests

pkmitre:
This adds considerable additional testing to patient builder; part of that includes updating the patient fixtures


andrequina:
Looks good.  We should see if we can get code coverage so that we can tell how much of a particular module is covered.  I see additional tests, but it's hard to tell how well tested things are.  I'll add an item to pivotal.

PR #54, opened 2014-01-09
fixes rationale classes to align with CSS guidance.  The css classes wer...

andrequina:
...e changed to use - rather than _.  However, the actual names of the classes were left the same since a better alternative was not obvious to align with bootstrap class names


carpeliam:
Rebased and merged.

PR #56, opened 2014-01-14
Load measure error handling

andrequina:
Adds error handling around measure loading and prevents loading a measure without selecting a zip file.


app/controllers/measures_controller.rb
```diff
@@ -52,7 +52,25 @@ def create
       existing.delete
     end
 
-    measure = Measures::MATLoader.load(params[:measure_file], current_user, measure_details)
+    begin
+      measure = Measures::MATLoader.load(params[:measure_file], current_user, measure_details)
+    rescue Exception => e
+      errors_dir = File.join('tmp','load_errors')
+      FileUtils.mkdir_p(errors_dir)
```
carpeliam, line 59: can we do this with chef?

PR #59, opened 2014-01-15
Fixed comparisons to show actual value for OBSERV, if it exists

jtferns:
**Story:** https://www.pivotaltracker.com/story/show/63578694
- Initialize an OBSERV expectation of undefined if an actual OBSERV value exists in the corresponding <code>result</code>
- Addresses bug where a patient is passing even when the OBSERV value is cleared and the non-OBSERV populations are matching

Tested with NQF0496, via <code>rake bonnie:db:reset DB=bonnie-production-eh</code>

actual_observ_values


pkmitre:
Due to an issue when initially merging this work was actually merged in pull request #64

PR #63, opened 2014-01-16
Replace patient builder criteria type select box

pkmitre:


pkmitre:
Due to an issue when initially merging this work was actually merged in pull request #65

PR #68, opened 2014-01-21
adding account.js top the precompilation list

rdingwell:


carpeliam:
I'm actually pretty sure we're not using account.js anymore, as `<ENTER>` should submit the form now that we're using submit buttons. I'll verify this.

carpeliam:
I confirmed we're not using this, and issued #69 to remove it. If that PR is accepted, we can delete this one.

pkmitre:
Closing this request and going with request #69 instead

PR #69, opened 2014-01-21
remove account.js, as it's no longer in use

carpeliam:
This fixes any issues people are running into with `rake assets:precompile`.


carpeliam:
This should supplant #68.

PR #72, opened 2014-01-22
fixes issue with results (values) not getting materialized properly

andrequina:


lib/measures/patient_builder.rb
```diff
@@ -93,7 +93,7 @@ def self.derive_entry(data_criteria,value, value_sets)
 
     # derive the values for the source data criteira
     def self.derive_values(entry, values, value_sets)
-      return if values.nil?
+      return if values.nil? || values.empty?
```
pkmitre, line 96: Just FYI values.nil? || values.empty? could be replaced with values.blank? (though I'm not going to hold up this pull request to wait for such a minor edit)

PR #75, opened 2014-01-26
Measure two-step delete, with sliding buttons

jtferns:
#### Story
- https://www.pivotaltracker.com/story/show/62487888
#### Notes
- Replaced measure's delete button with consistent two-step delete button used within patient results and patient data criteria, moved measure settings gear icon into title bar
- Updated styling to match v5 mockups
- Added simple CSS3 sliding animation for buttons
#### Images

0105closed
0105settings
0105expanded
0387closed
0387settings
0387expanded


app/assets/stylesheets/measure.less
```diff
@@ -31,6 +31,56 @@
...
+      padding-right: 6px;
+    }
+    .update-measure {
+      margin-top: 7px;
+      padding-left: 6px;
+      padding-right: 6px;
+      .fa {
+        color: @brand-primary;
+      }
+    }
```
pkmitre, line 65: At a quick glance the arbitrary padding sizes and the explicit browser support for transitions (rather than using a mixin) could be cleaned up... can you work with Liam to get this CSS cleaned up, he'll likely have good approaches.

pkmitre:
Noted one issue: if you click the gear, and then click the minus, clicking the gear doesn't hide the resulting buttons; it should hide the buttons and reset them.

app/assets/stylesheets/measure.less
```diff
@@ -24,12 +24,51 @@
...
+      }
+      .update-measure {
+        .measure-settings-padding;
+        .fa {
+          color: @brand-primary;
+        }
+      }
+    }
+    .measure-settings-expanded {
+      width: 16%;
```
carpeliam: Setting a percentage-based width makes me pretty nervous... this might be worth talking through in person. PS, http://demosthenes.info/blog/775/Which-CSS-Measurements-To-Use-When is a good resource (even if [Bootstrap breaks some of these standards when it comes to ems vs pixels](https://github.com/twbs/bootstrap/issues/1943)).

app/assets/stylesheets/measure.less
```diff
@@ -24,12 +24,51 @@
...
-      font-size: 1.8em;
+      .vertical-spacing;
       padding-right: 15px;
-      line-height: 1.75;
     }
     .full-title {
       line-height: 1.6;
+      padding-right: 2%;
+    }
+    .settings-btn {
```
carpeliam: `.btn-settings` might be more inline with the other btn customizations we do elsewhere.

jtferns:
Updated to use better styles and template, closing and will re-open with rebased branch

PR #78, opened 2014-01-27
added coded entry id to the source data criteria which is the id of the ...

andrequina:
...coded entry generated on the patient for that source data criteria

### DIFF
--- a/app/assets/javascripts/models/patient.js.coffee
+++ b/app/assets/javascripts/models/patient.js.coffee
@@ -110,13 +110,12 @@ class Thorax.Models.Patient extends Thorax.Model
       expectedValues.add @getExpectedValue(population)
     expectedValues
 
-  highlightCodedEntry: (codedEntryId) ->
+  highlightCodedEntry: (codedEntryId, type) ->
     # picked up by EditCriteriaView
-    sourceDataCriterium.trigger 'highlight' for sourceDataCriterium in @get('source_data_criteria').models when sourceDataCriterium.get('coded_entry_id') == codedEntryId
+    sourceDataCriterium.trigger 'highlight', type for sourceDataCriterium in @get('source_data_criteria').models when sourceDataCriterium.get('coded_entry_id') == codedEntryId
 
 class Thorax.Collections.Patients extends Thorax.Collection
-  model: Thorax.Models.Patient  
-  dedupName: (patient) ->
+  model: Thorax.Models.Patient  dedupName: (patient) ->
     return patient.first if !(patient.first && patient.last)
     #matcher to find all of the records that have the same last name and the first name starts with the first name of the 
     #patient data being duplicated 
--- a/app/assets/javascripts/models/patient.js.coffee
+++ b/app/assets/javascripts/models/patient.js.coffee
@@ -89,7 +89,7 @@ class Thorax.Models.Patient extends Thorax.Model
       defaults = {}
       defaults[section] = [] for section in Thorax.Models.Patient.sections
       @set _(data).chain().pick(_(defaults).keys()).defaults(defaults).value(), silent: true
-      for criterium, i in _this.get('source_data_criteria').models
+      for criterium, i in @get('source_data_criteria').models
         criterium.set 'coded_entry_id', data['source_data_criteria'][i]['coded_entry_id'], silent: true
       @trigger 'materialize' # We use a new event rather than relying on 'change' because we don't want to automatically re-render everything
 
@@ -111,7 +111,8 @@ class Thorax.Models.Patient extends Thorax.Model
     expectedValues
 
 class Thorax.Collections.Patients extends Thorax.Collection
-  model: Thorax.Models.Patient  dedupName: (patient) ->
+  model: Thorax.Models.Patient  
+  dedupName: (patient) ->
     return patient.first if !(patient.first && patient.last)
     #matcher to find all of the records that have the same last name and the first name starts with the first name of the 
     #patient data being duplicated 
--- a/app/assets/javascripts/models/patient.js.coffee
+++ b/app/assets/javascripts/models/patient.js.coffee
@@ -110,9 +110,12 @@ class Thorax.Models.Patient extends Thorax.Model
       expectedValues.add @getExpectedValue(population)
     expectedValues
 
+  highlightCodedEntry: (codedEntryId) ->
+    # picked up by EditCriteriaView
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/models/patient.js.coffee
```diff
@@ -103,5 +105,9 @@ class Thorax.Models.Patient extends Thorax.Model
       expectedValues.add @getExpectedValue(population)
     expectedValues
 
+  highlightCodedEntry: (codedEntryId, type) ->
+    # picked up by EditCriteriaView
+    sourceDataCriterium.trigger 'highlight', type for sourceDataCriterium in @get('source_data_criteria').models when sourceDataCriterium.get('coded_entry_id') == codedEntryId
```
carpeliam: It sounds like highlighting is a view activity, not a model activity- if possible, it'd be great to trigger this event from a view, instead of a model. (Also, all things considered, I guess I'd rather see this take a codedEntry model, instead of a codedEntryId, if that's just as easy.)

app/assets/javascripts/models/result.js.coffee
```diff
@@ -154,6 +154,16 @@ class Thorax.Models.Result extends Thorax.Model
       _.extend(orCounts, @calculateOrCountsRecursive(rationale, precondition.preconditions))
     return orCounts
 
+  highlightPatientData: (dataCriteriaKey, populationCriteriaKey) ->
```
carpeliam: Same deal here as mentioned on the patient - this method seems relevant to the view, not to the model.

app/assets/javascripts/models/result.js.coffee
```diff
@@ -154,6 +154,16 @@ class Thorax.Models.Result extends Thorax.Model
...
+    matchingCodedEntries = @get('rationale')[dataCriteriaKey]?['results']
+    if matchingCodedEntries
+      index = hqmf.SpecificsManager.indexLookup[@measure.get('data_criteria')[dataCriteriaKey].source_data_criteria]
+      goodElements = (row[index] for row in @get('finalSpecifics')[populationCriteriaKey]) if index?
+      for codedEntry in matchingCodedEntries
+        type = (if goodElements && goodElements.indexOf(codedEntry.id) < 0 then "partial" else "valid")
```
carpeliam: Instead of passing around strings for the type, I think it's a good idea if we can define all of the different types in one place, and reference them from there.

app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -38,6 +39,21 @@ class Thorax.Views.PopulationLogic extends Thorax.View
             target = @$(".#{code}_children .#{key}")
             if (target.length > 0)
               target.addClass(if updatedRationale[code]?[key] is false then 'eval-bad-specifics' else "eval-#{!!value}")
+  events:
+    rendered: ->
+      @$('.highlight-target').mouseover(_.bind(@highlightEntry, this))
```
carpeliam: This is a bit of a smell - it'd be nice to listen to the mouseover event, but because the highlight-target is actually a sub-view, it doesn't get fired here. This also has the side-effect of firing on all child elements of `.highlight-target`. If at all possible, we should be listening to events within the scope that they're fired, and then sending messages between the views.

app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -38,6 +39,21 @@ class Thorax.Views.PopulationLogic extends Thorax.View
...
+      @$('.highlight-target').mouseout(_.bind(@clearHighlightEntry, this))
+
+  highlightEntry: (e) ->
+    return unless $(e.target).hasClass('highlight-target')
+    view = $(e.target).view()
+    dataCriteriaKey = view?.dataCriteria?.key
```
carpeliam: If it's not possible to move the mouseover/out events to their relevant views, we should remove the question marks here; we shouldn't need to be defensive in this line or the next.

app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -18,6 +18,7 @@ class Thorax.Views.PopulationLogic extends Thorax.View
     for code in Thorax.Models.Measure.allPopulationCodes
       match = @model.get code
       @submeasurePopulations.push(match) if match
+    @on 'foo', -> alert('here')
```
carpeliam: What happens when I foo?

app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -39,5 +41,19 @@ class Thorax.Views.PopulationLogic extends Thorax.View
             if (target.length > 0)
               target.addClass(if updatedRationale[code]?[key] is false then 'eval-bad-specifics' else "eval-#{!!value}")
 
+  highlightPatientData: (dataCriteriaKey, populationCriteriaKey) ->
+    if @latestResult && @latestResult.get('finalSpecifics')?[populationCriteriaKey]
```
carpeliam: how about `if @latestResult?.get('finalSpecifics')?[populationCriteriaKey]`

app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -39,5 +41,19 @@ class Thorax.Views.PopulationLogic extends Thorax.View
...
+      matchingCodedEntries = @latestResult.codedEntriesForDataCriteria(dataCriteriaKey)
+      if matchingCodedEntries
+        goodElements = @latestResult.codedEntriesPassingSpecifics(dataCriteriaKey, populationCriteriaKey)
+        for codedEntry in matchingCodedEntries
+          type = (if goodElements && goodElements.indexOf(codedEntry.id) < 0 then Thorax.Views.EditCriteriaView.highlight.partial else Thorax.Views.EditCriteriaView.highlight.valid)
```
carpeliam: This is a pretty long line of code (same with L52), maybe the ternary operator is inappropriate. You can also collapse `if goodElements && goodElements.indexOf(codedEntry.id)` into `if goodElements?.indexOf(codedEntry.id)`.

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -85,6 +85,8 @@ class Thorax.Models.Patient extends Thorax.Model
       defaults = {}
       defaults[section] = [] for section in Thorax.Models.Patient.sections
       @set _(data).chain().pick(_(defaults).keys()).defaults(defaults).value(), silent: true
+      for criterium, i in _this.get('source_data_criteria').models
```
carpeliam: `_this.` should be `@` ... right?

PR #85, opened 2014-01-31
Add password strength requirements and account lockout for login failure...

pkmitre:
...s

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -1,9 +1,21 @@
 class User
   include Mongoid::Document
   # Include default devise modules. Others available are:
-  # :confirmable, :lockable, :timeoutable and :omniauthable
-  devise :database_authenticatable, :registerable,
+  # :confirmable, :timeoutable and :omniauthable
+  devise :database_authenticatable, :registerable, :lockable,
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -1,10 +1,22 @@
...
+  # :confirmable, :timeoutable and :omniauthable
+  devise :database_authenticatable, :registerable, :lockable,
          :recoverable, :rememberable, :trackable, :validatable
 
+  # Validate password complexity
+  validate :password_complexity
+  def password_complexity
+    if password.present?
+      # Passwords must have characters from at least two groups, identified by these regexes (last one is punctuation)
+      matches = [/[a-z]/, /[A-Z]/, /[0-9]/, /[^\w\s]/].select { |rx| rx.match(password) }.size
```
carpeliam, line 13: Apparently `[[:punct:]]` or `\p{P}` also works for punctuation, but doesn't cover non-punctuation characters like `$` or `~` so this looks good.

carpeliam, line 13: And if somebody figures out how to get a tab character into their password, then God bless them

PR #86, opened 2014-02-03
adding ability for portfolio users to enable/disable crosswalk checking

rdingwell:


app/controllers/registrations_controller.rb
```diff
@@ -0,0 +1,18 @@
+class RegistrationsController < Devise::RegistrationsController
+
+
+  # updated from original to look for changes to crosswalking changes
+  def update_resource(resource, params)
+    params[:crosswalk_enabled] = params[:crosswalk_enabled] != "1" ? false : true
```
pkmitre: When dealing with booleans you typically don't need to use `true` or `false` directly; something like

```
params[:crosswalk_enabled] = params[:crosswalk_enabled] == "1"
```

would be fine...but, further than that, Rails is pretty good about dealing with params, and handles '0' and '1' correctly when saving a boolean, so this line can actually be entirely removed.

app/controllers/registrations_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    crosswalk_enabled = resource.crosswalk_enabled
+    saved = super(resource, params)
+    if saved && resource.is_portfolio? && (resource.crosswalk_enabled != crosswalk_enabled)
+      Measure.by_user(current_user).each do |m|
+        m.map_fns.clear
+        m.save
+     end
```
pkmitre: These lines could be

```
Measure.by_user(current_user).each { |m| m.generate_js clear_db_cache: true, check_crosswalk: resource.crosswalk_enabled }
```

The JS needs to be generated anyway, so adding that work here shouldn't be an issue, plus this is the way the cache is cleared elsewhere.

pkmitre:
It looks like this depends on a change to bonnie_bundler (the ones that add support for check_crosswalk), but that changed dependency isn't captured in the pull request (no update to Gemfile.lock), and so merging this will break production as well as everyone's dev environment.

pkmitre:
Thinking about this a bit more, there are a number of places where JavaScript gets regenerated (measure re-import, debug-mode measure JS reset), and this pull request only inserts the check_crosswalk option in one of those places. Maybe a more holistic approach would be useful, consolidating all these into one place where check_crosswalk can be taken into account. Maybe a method on the User model to regenerate JS, that way instead of

```
Measure.by_user(current_user).each { |m| m.generate_js clear_db_cache: true, check_crosswalk: resource.crosswalk_enabled }
```

We'd just do something like

```
current_user.regenerate_measure_js
```

and optionally pass in the measure(s) we want to reset?

andrequina:
It seems like the crosswalk check is being over thought.  Cross walk checking is not part of the normal process.  You would not be loading new measures, re-importing, etc.  Cross walk checking is something that's done when we want to check if there are any partial matches on codes.

You would have all the measures loaded, and all the expectations would be passing.  Think about an annual update where you have already aligned the patients with the expectations.  The final step is to check if any of the codes are "cross walking".  You would re-generate the javascript, then you would check to see if any of the patients are failing.  If they are you have to identify what codes are crossing over.  I think that re-architecting anything to handle cross walking is not the way to go.  This is not part of the typical use case and would never be done by anyone other than a MITRE person trying to finalize a bundle.

pkmitre:
If it's "not part of the normal process", then pick an approach that isn't so integral to existing code, touching lots of pieces and having the possibility of leaving things in an inconsistent and surprising state. Instead of making this an account setting, just do something like
1. Add an action to the measures controller to regenerate all JS that takes an argument on whether to generate with crosstalk or not (maybe just extend the existing clear_cached_javascript action to be able to clear all measures if no ID is provided).
2. Add buttons somewhere on the site, that only appear to portfolio users, labeled "Regenerate JS With Crosswalk" and "Regenerate JS without Crosswalk", or something similar if that doesn't quite make sense.

andrequina:
If there's a cleaner approach that's fine, I just want to try and avoid re-architecting the calculation infrastructure to support something that's used extremely rarely and for a very specific use case.

app/controllers/registrations_controller.rb
rdingwell: I don'tI agree with this.  JS Generation takes a good amount of time and it needs to be recached on the browser as well so utilizing the mechanism that is already in place to do that in the controller makes more sense.  That way this call doesn't just sit there and hang only to have the dashboard take the time to recache all of the measures on the browser.

rdingwell:
I don't think this was over thought or touches that much code , it was only ~40 lines across all files touched with the majority of those being in the registration controller that was added.  It's pretty simple, if the use is a portfolio user and updates their cross walking preference then all of the is cleared off the measures and regenerated when the user hits the dashboard again.  Any measure that gets loaded during that point in time will also adhere to that setting to avoid inconsistency.

Just having a button that says regenerate with/or without crosswalking enabled is something that can leave the measures in an inconsistent  state.   If another measure is added after that button is pushed what js is going to be generated?    If you leave for the night and come back how do you know what js is currently generated if you dont remember what state you left it in the night before?   Or if someone else comes in to work on the test deck how do they know how the js was generated?   More than one person will be editing the test deck but will use the same account so this needs to be kept track of.  You also don't want it to step on anybody else that may be working on the system (EP and EH test deck users) so it's something that you want to keep track of on a user basis.  

You can make the argument as to whether or not it should be on the profile page but if it's not that same sort of functionality would need to exist somewhere else.   Say you do it with a button on the dashboard. You still need to keep track of this so the flag on the user stays.  You need some way to indicate what mode the measures are currently generated in (this is what the check box on the profile page does now).  When pressed the button will have to go to some controller that can update the users parameter and regenerate the measures. ( This is what the registration controller is doing now).    I can see doing this in the measures controller in the clear cached method eliminating the need for the registrations controller.   I'm fine with this but all it really does is move things around.

pkmitre:
>  Any measure that gets loaded during that point in time will also adhere to that setting to avoid inconsistency.

For that to happen the `create`, `finalize`, and `clear_cached_javascript` all need to be aware of the setting, because they all call generate_js... that's really my main concern, which I probably didn't explain clearly: the JS gets generated in a handful of places, and these places need to be aware of the setting. Ideally I don't want every place that calls generate_js to have to look at the users setting, was just hoping there was a more centralized way to do it. Hence my comment above where I refer to an option of using something like `current_user.regenerate_measure_js`

app/controllers/registrations_controller.rb
pkmitre: Yeah, i guess the work needs to happen some time, so it doesn't really matter when.

pkmitre:
I've moved the check for the user crosswalk setting down into bonnie_bundler right at the level where the JS is generated, so now it's transparent in the rest of the bonnie codebase.

PR #87, opened 2014-02-03
Add support for user-defined codes

carpeliam:


app/assets/javascripts/models/patient.js.coffee
```diff
@@ -54,10 +54,8 @@ class Thorax.Models.Patient extends Thorax.Model
     else unless @get('ethnicity').name? then "CDC-RE: #{@get('ethnicity').code}"
     else @get('ethnicity').name
   getInsurance: ->
-    insurances = @get('insurance_providers')?map (ip) ->
-      ip.name
-    insurances?join(", ")
-    unless insurances? then ''
+    insurances = @get('insurance_providers')?.map (ip) -> ip.name
```
carpeliam, line 57: This is unrelated to the commit, but I wasn't sure where else to put this fix.

app/assets/javascripts/models/measure.js.coffee
```diff
@@ -101,3 +101,6 @@ class Thorax.Collections.Measures extends Thorax.Collection
     @each (m) -> m.get('populations').each (p) -> populations.add(p)
     populations
 
+  valueSets: ->
+    _(@map (m) -> m.get('value_sets')?.models or []).chain().flatten().uniq((vs) -> vs.get('oid')).value()
```
pkmitre: Backbone includes the underscore chain as a collection method; does

```
@chain().map((m) -> m.get('value_sets')?.models or []).flatten().uniq((vs) -> vs.get('oid')).value()
```

clean this up at all?

carpeliam: Works for me. (I don't think I knew that `chain` was added via Backbone.) Getting the value sets from the measures is most likely a temporary approach, but I'm cool with making the change in the meantime.

carpeliam:
I'm running into some problems running the code at the moment (codes don't seem to be persisting across saves)... I think this might be due to some awkward rebase merge conflicts that may have found their way in, I'm going to do a bit more debugging. This makes me want to fix our value set strategy sooner rather than later so that there's some test coverage on this...

pkmitre:
Not sure it's a related issue to your comment above, but I was unable to delete codes once they were added.

carpeliam:
@pkmitre I've fixed the issues I was running up against. There is still a bit of a usability issue; if you delete the last code, it resets the code source to `DEFAULT` and sends back down the default code.. this kind of gives the appearance of not being able to delete a code if it's the last one, not sure if that's what you were running into. I checked this with @andrequina and this is ok for the moment; this will be fixable when we update the data criteria UI so that you don't have to click the [+] button to add codes/values/field values.

PR #88, opened 2014-02-04
Make measure library JS into a standard asset and tweak server config

pkmitre:
This significantly increases performance when loading JS assets

### DIFF
--- a/app/controllers/populations_controller.rb
+++ b/app/controllers/populations_controller.rb
@@ -1,10 +1,8 @@
 class PopulationsController < ApplicationController
 
   def calculate_code
-    # FIXME: Could be faster if only load the required attr
-    measure = Measure.by_user(current_user).find(params[:measure_id])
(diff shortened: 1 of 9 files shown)


app/controllers/populations_controller.rb
```diff
@@ -1,10 +1,8 @@
 class PopulationsController < ApplicationController
 
   def calculate_code
-    # FIXME: Could be faster if only load the required attr
-    measure = Measure.by_user(current_user).find(params[:measure_id])
+    measure = Measure.by_user(current_user).only(:updated_at, :data_criteria, :source_data_criteria, :map_fns).find(params[:measure_id])
     if stale? last_modified: measure.updated_at.try(:utc), etag: measure.cache_key
-      # FIXME: Ensure that this is considered JS in IE
       render text: BonnieMeasureJavascript.generate_for_population(measure, params[:id].to_i)
```
carpeliam: Can we use `render js:` here instead of `render text:`?

pkmitre: After a bit of testing it looks like `render :js` makes explicit what the .js extension implies, so I've made the change.

PR #96, opened 2014-02-06
Save codes with the same format they are on the server

carpeliam:
Prior to this, codes were initialized with a format optimized for client-side use, but then persisted with that same format. This PR isolates the client-side formatting of the codes to the Codes collection itself, and persists codes in the format the server should expect.


app/assets/javascripts/models/patient.js.coffee
```diff
@@ -114,15 +112,15 @@ class Thorax.Models.Patient extends Thorax.Model
     expectedValues
 
 class Thorax.Collections.Patients extends Thorax.Collection
-  model: Thorax.Models.Patient  
+  model: Thorax.Models.Patient
   dedupName: (patient) ->
     return patient.first if !(patient.first && patient.last)
-    #matcher to find all of the records that have the same last name and the first name starts with the first name of the
```
carpeliam, line 120: Just whitespace changes down here, my editor automatically removes EOL whitespace... sorry for any confusion. :/

PR #100, opened 2014-02-13
Show coverage for all patients in Measure view

jtferns:
#### Story
- https://www.pivotaltracker.com/story/show/65372354
#### Notes
- show coverage for all patients in default rendering on Measure view; this gets overidden when a test patient result is expanded
- when a test patient result is expanded, the coverage section shows a button to return to the coverage for all patients; when test patients are collapsed, a disabled button is rendered
- replace total coverage percentage with a jquery spinner with brand-primary color
- highlight logic view with brand-primary and lighter blue
#### Screenshots
- overallcoveragedefault
- overallcoverageexpanded

### DIFF
--- a/app/assets/javascripts/views/measure_view.js.coffee
+++ b/app/assets/javascripts/views/measure_view.js.coffee
@@ -14,6 +14,8 @@ class Thorax.Views.Measure extends Thorax.View
       @logicView = populationLogicView
 
     @populationCalculation = new Thorax.Views.PopulationCalculation(model: population)
+    @logicView.listenTo @populationCalculation, 'logicView:showCoverage', -> @showCoverage()
+    @logicView.listenTo @populationCalculation, 'logicView:clearCoverage', -> @clearCoverage()
 
     @populationCalculation.listenTo @logicView, 'population:update', (population) -> @updatePopulation(population)
     # FIXME: change the name of these events to reflect what the measure calculation view is actually saying
--- a/app/assets/stylesheets/application.css.less
+++ b/app/assets/stylesheets/application.css.less
@@ -19,6 +19,3 @@
 @import "measure_calculation";
 @import "rationale";
 @import "builder";
-
-
-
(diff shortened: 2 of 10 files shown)


app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -16,7 +16,11 @@ class Thorax.Views.Measure extends Thorax.View
     else
       @logicView = populationLogicView
 
-    @populationCalculation = new Thorax.Views.PopulationCalculation(model: population)
+    @populationCalculation = new Thorax.Views.PopulationCalculation
+      model: population
+      coverageView: new Thorax.Views.MeasureCoverageView(model: population.coverage())
```
pkmitre: Can the coverage view be set up within the PopulationCalculation view, rather than being set up outside and passed in?

app/assets/stylesheets/application.css.less
```diff
@@ -20,3 +20,8 @@
 @import "rationale";
 @import "builder";
 
+
+
+
+
+
```
pkmitre: I know that having to tweak the application.css.less file is a pain, but it would be good if we didn't check in these empty changes.

app/assets/javascripts/views/measure_view.js.coffee
jtferns: Ah yes, it should be moved into the PopulationCalculation view; I had <code>logicView</code> listening to <code>coverageView</code> initially, which was why I had it set up externally.

jtferns: Actually, when I move the <code>coverageView</code> into <code>populationCalculation.initialize()</code>, I run into an issue where <code>logicView</code> fails to listen to the <code>show/clearCoverage</code> triggers when viewing measures with multiple populations...

I believe that setting it within <code>populationCalculation</code> causes it to change as a view instance when <code>populationCalculation.updatePopulation()</code> is called and <code>coverageView</code>'s model is updated.

Branch with changes: 64d84a166ff9acec3e319a464378a6db4453ee2f

PR #102, opened 2014-02-17
Basic Patient Builder validation support

pkmitre:


app/assets/javascripts/models/patient.js.coffee
```diff
@@ -111,6 +111,12 @@ class Thorax.Models.Patient extends Thorax.Model
       expectedValues.add @getExpectedValue(population)
     expectedValues
 
+  validate: ->
+    errors = [] # We don't use a hash because we want to control order
```
carpeliam: I think this comment is probably unnecessary - I didn't react to this line thinking that `errors` should be a hash, and I don't know that anybody would be confused/upset or try to refactor this if the comment was removed. That being said, it doesn't take up a lot of space :) but i don't think it does anything for readability.

app/assets/javascripts/templates/patient_builder/patient_builder.hbs
```diff
@@ -3,8 +3,12 @@
     <h1>Test Patient</h1>
   </div>
 
-  <div class="col-xs-10">
-    <span class="timeline-icon"><i class="fa fa-user fa-3x"></i></span>
+  <div class="col-xs-6">
```
carpeliam: Using `col-xs-` here probably doesn't make sense anymore. At some point (doesn't have to be in this PR), these should probably be `col-md`. It's probably just as easy to do that now, provided it doesn't look like crap any more than the rest of everything does when shrunk down to a phone size.

app/assets/javascripts/templates/patient_builder/patient_builder.hbs
```diff
@@ -15,12 +19,12 @@
 <div class="row">
   <form class="col-left" role="form">
     <div class="form-group">
-      <label for="last">Last Name</label>
+      <label for="last" class="control-label">Last Name</label>
```
carpeliam, line 22: Why are we only adding the `.control-label` class to some label elements and not others?

app/assets/javascripts/models/patient.js.coffee
pkmitre: My usual reaction to seeing an associative array instead of a hash is to ask "why isn't this a hash?". But you're right, it's line noise, I'll remove it.

app/assets/javascripts/templates/patient_builder/patient_builder.hbs
pkmitre: I'll switch it in and check.

app/assets/javascripts/templates/patient_builder/patient_builder.hbs
pkmitre, line 22: I only added it to fields that can get validation errors. Do you think using it everywhere would be an improvement? On the one hand, adding further validations would just work, but on the other it's a bit of line noise.

carpeliam, line 22: I think if it's semantically a `control-label`, then anything that's a control label should have that class or not have that class. I think that's probably good principle of least surprise (and with markup, I think it's easy to ignore that kind of line noise). If it's semantically a `validatable-label` or whatever, then I'd use a class that makes sense. (Dunno if "validatable-label" is really what you'd want, that's just a shot in the dark.)

pkmitre, line 22: `control-label` is a bootstrap concept, so I think it makes sense to be consistent.

I'll change the rest of them in the patient builder.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -114,9 +106,24 @@ class Thorax.Views.PatientBuilder extends Thorax.View
 
   save: (e) ->
     e.preventDefault()
+    @$('.has-error').removeClass('has-error')
     $(e.target).button('saving').prop('disabled', true)
     @serializeWithChildren()
-    @originalModel.save @model.toJSON() # FIXME: The sync event on the model, defined above, only fires because there's a Thorax bug!
+    status = @originalModel.save @model.toJSON(),
+      success: (model) =>
```
carpeliam, line 113: this fixes the FIXME, right?

pkmitre, line 113: Yep!

PR #109, opened 2014-02-21
Significantly improve speed of value set loading

pkmitre:
It's not the prettiest code, but with 64 measures loaded it reduces the time to load value sets on my machine from ~10 seconds to ~2 seconds.

I left in the commented out original "clean" code so that the intent is clearer to anyone looking at it.


carpeliam:
Is there ever a time when you'd want to remove the commented out code?

pkmitre:
I'd like to keep them as a way to make the intent clear, given that the performant code is somewhat obscure.

PR #114, opened 2014-02-26
Fix account page forms

jtferns:
#### Story
- https://www.pivotaltracker.com/story/show/65816846
#### Notes
- fixed grid styling on account edit and registration pages
- ~~added labels to form elements in account edit page (for clarity, can be removed if undesired)~~
#### Screenshots
- **previous** account registration page
  - default_accountregister
- **current** account registration page
  - accountregister
- **previous** account edit page
  - default_accountedit
- **current** account edit page
  - screen shot 2014-02-26 at 5 06 32 pm

### DIFF
--- a/app/views/devise/registrations/edit.html.erb
+++ b/app/views/devise/registrations/edit.html.erb
@@ -12,23 +12,31 @@
   <% end %>
 
   <div class="form-group row">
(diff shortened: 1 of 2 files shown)


carpeliam:
I think the account edit page should be consistent with the account registration page - I think we can drop the labels.

PR #122, opened 2014-03-04
Fix several bugs...

jtferns:
#### Stories
- https://www.pivotaltracker.com/story/show/66683682
  - The measure logic should not be collapsed by default
- https://www.pivotaltracker.com/story/show/66683740
  - Expand measure logic on change of selected patient
- https://www.pivotaltracker.com/story/show/66684144
  - re-add failing patients to coverage
- https://www.pivotaltracker.com/story/show/66684206
  - question mark icon should be replaced with N/A in text muted class
- https://www.pivotaltracker.com/story/show/66684130
  - Hide the "show" button when currently disabled
- https://www.pivotaltracker.com/story/show/66684836
  - create a white/black list link for portfolio user
- https://www.pivotaltracker.com/story/show/66684318
  - Align the show and add patient button and the patient names with PASS/FAIL
#### Screenshots
- measure logic should not be collapsed by default
  - one
- question mark icon should be replaced with N/A
  - two
- hide show coverage button when patient result is collapsed
  - three
- create link for portfolio user to white/black-list builder
  - four
- align show coverage button with add patient button and fix vertical alignment in patient result panel headings
  - five_2

### DIFF
--- a/app/assets/javascripts/templates/logic/population_criteria.hbs
+++ b/app/assets/javascripts/templates/logic/population_criteria.hbs
@@ -5,14 +5,14 @@
         <div>
           <span class="{{population.type}} rationale-target"><b>{{translate_population population.type}}: </b></span>
           <div class="pull-right">
-            <i class="fa fa-lg fa-angle-right toggle-icon"></i>
+            <i class="fa fa-lg fa-angle-down toggle-icon"></i>
             <span class="sr-only">toggle {{translate_population population.type}}</span>
           </div>
-        </div>
+        </div> 
       </a>
     </h4>
   </div>
-  <div id="collapse{{cid}}" class="panel-collapse collapse">
+  <div id="collapse{{cid}}" class="panel-collapse collapse in">
     <div class="panel-body">
       {{#if rootPreconditon}}
         {{#if rootPreconditon.preconditions }}
--- a/app/assets/javascripts/models/coverage.js.coffee
+++ b/app/assets/javascripts/models/coverage.js.coffee
@@ -10,7 +10,7 @@ class Thorax.Model.Coverage extends Thorax.Model
 
     # Find all unique criteria that evaluated true in the rationale that are also in the measure
     @rationaleCriteria = []
-    @differences.each (difference) => if difference.get('done') and difference.get('match')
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/templates/logic/population_criteria.hbs
```diff
@@ -4,12 +4,12 @@
...
         <div>
           <span class="{{population.type}} rationale-target"><b>{{translate_population population.type}}: </b></span>
-          <div class="pull-right"><i class="fa fa-lg fa-angle-right toggle-icon"></i></div>
+          <div class="pull-right"><i class="fa fa-lg fa-angle-down toggle-icon"></i></div>
         </div> 
       </a>
     </h4>
   </div>
-  <div id="collapse{{cid}}" class="panel-collapse collapse">
+  <div id="collapse{{cid}}" class="panel-collapse collapse in">
```
carpeliam, line 15: I think you can actually accomplish the same thing here with the following: `<div id="collapse{{cid}}" class="panel-collapse">`

jtferns, line 15: `<div id="collapse{{cid}} class="panel-collapse">` expands out the population logic from the panel heading, even though it is already expanded...

`<div id="collapse{{cid}} class="panel-collapse in">` might be the desired initial class for the population panels?

carpeliam, line 15: Yep! you're right.

carpeliam:
Everything looks ok at a glance, but it's hard to evaluate several issues in a single pull request (hard to look through and see what's related to what). Does anybody else want to take a peek? Or maybe it'd be better to go through them in person...?

andrequina:
Lets get this done today if possible.

carpeliam:
During standup, we talked about this; I've looked through the code and things look ok, but haven't tested each fix individually; @pkmitre mentioned he'd run through the fixes in a browser. We also resolved to keep PRs limited to individual features as they're easy to inspect and should get accepted more quickly.

A bit less important, but not totally unimportant, is that this style of PR also makes it almost impossible to adhere to our git workflow, as there are a number of branches merged into one, and they weren't committed sequentially.

PR #150, opened 2014-03-16
Support downloading a user bundle from the users admin page

andrequina:
...page

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -53,20 +53,12 @@ class Admin::UsersController < ApplicationController
   def bundle
     user = User.find(params[:id])
     measures = user.measures
-
-    config = {}
-    config["hqmf_path"] = 'none'
-    config['version'] = '1.0'
-    config["use_nqf"] = false
-    stringio = Zip::ZipOutputStream::write_buffer do |zip|
-      exporter = Measures::Exporter::BundleExporter.new(measures,config,zip)
-      exporter.export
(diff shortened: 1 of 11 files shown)


andrequina:
this depends on changes in bonnie_bundler in a branch of the same name

app/controllers/admin/users_controller.rb
```diff
@@ -50,6 +50,25 @@ def measures
...
+  def bundle
+    user = User.find(params[:id])
+    measures = user.measures
+
+    config = {}
+    config["hqmf_path"] = 'none'
+    config['version'] = '1.0'
+    config["use_nqf"] = false
+    stringio = Zip::ZipOutputStream::write_buffer do |zip|
+      exporter = Measures::Exporter::BundleExporter.new(measures,config,zip)
```
pkmitre: This approach of passing a zip construct into the exporter feels quite awkward; the new code in the exporter (in the bonnie_bundler pull request) feels even more awkward (setting a zip instance variable, which is referenced in the "write_file" method, which then doesn't actually write a file...).

I think refactoring this into something like

```
config = ...
exporter = Measures::Exporter::BundleExporter.new(measures, config)
zip_data = exporter.export_zip

cookies[:fileDownload] = "true" # We need to set this cookie for jquery.fileDownload

send_data zip_data, :type => 'application/zip', :disposition => 'attachment', :filename => "bundle_#{user.email}_export.zip"
```

would clean things up quite a bit.

PR #151, opened 2014-03-16
add screen reader status indication for highlighted sections of the logi...

andrequina:
...c.


app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -46,8 +46,16 @@ class Thorax.Views.PopulationLogic extends Thorax.View
             if (target.length > 0)
               target.addClass(if updatedRationale[code]?[key] is false then 'eval-bad-specifics' else "eval-#{!!value}")
               target.closest('.panel-heading').addClass(if updatedRationale[code]?[key] is false then 'eval-panel-bad-specifics' else "eval-panel-#{!!value}")
+      @rationaleScreenReaderStatus()
```
carpeliam, line 49: It seems like it might be easier and more direct to use the existing reference to `target`, instead of separating the screen reader code into a separate method.

app/assets/stylesheets/rationale.less
```diff
@@ -33,6 +33,10 @@
     background-color: @blue-lightest;
     color: @blue-darker;
   }
+  .eval-uncovered{
+    background-color: #FFF;
```
carpeliam: I'd prefer `white` and `black` instead of `#FFF` and `#000`.

PR #152, opened 2014-03-17
Clone patients from one measure into another

jtferns:
#### Story
- https://www.pivotaltracker.com/story/show/65921508
#### Notes
- updated measure view to include measure listing
- users can select the patient cloning toggle button from the patient settings gear, which modifies the patient results to checkbox selectables, and expands a measure listing into the view with selectable measures by <code>cms_id</code>
- selecting a measure in the measure listing reveals a clone button, which clones the selected patients into the measure, and then re-routes the application to the selected measure's measure view
#### Screenshots
- NQF0105 before cloned patients
  - screen shot 2014-03-17 at 11 34 03 am
- NQF0104 with new patient cloning settings button
  - screen shot 2014-03-17 at 11 34 48 am
- NQF0104 with patient cloning toggled, patients selected, and NQF0105 selected
  - screen shot 2014-03-17 at 11 35 32 am
- NQF0105 after cloned patients from NQF0104
  - screen shot 2014-03-17 at 11 35 40 am

### DIFF
--- a/app/assets/javascripts/views/measure_view.js.coffee
+++ b/app/assets/javascripts/views/measure_view.js.coffee
@@ -55,25 +55,32 @@ class Thorax.Views.Measure extends Thorax.View
     @$('.measure-listing').removeClass('active')
     @$('.btn-clone-patients').hide()
     m = @$(e.target).model()
-    @$(".measure-#{m.get('hqmf_set_id')}").addClass('active')
-    @$(".btn-clone-#{m.get('hqmf_set_id')}").show()
+    if @$('.select-patient:checked').length
+      @$(".measure-#{m.get('hqmf_set_id')}").addClass('active')
+      @$(".btn-clone-#{m.get('hqmf_set_id')}").show()
 
   cloneIntoMeasure: (e) ->
     $d = @$('.select-patient:checked')
-    measure = @measures.findWhere({hqmf_set_id: @$('.measure-listing.active').model().get('hqmf_set_id')})
+    measure = @measures.findWhere({hqmf_set_id: @$(e.target).model().get('hqmf_set_id')})
     count = 0
+    # @$("#clonePatientsDialog").modal backdrop: 'static'
+    @$(".rebuild-patients-progress-bar").css('width', '0%')
     for diff in $d
       difference = @$(diff).model()
       patient = @patients.findWhere({medical_record_number: difference.result.get('medical_record_id')})
-      clonedPatient = patient.deepClone(omit_id: true, dedupName: true)
+      clonedPatient = patient.deepClone(omit_id: true)
       clonedPatient.set('measure_ids', [measure.get('hqmf_set_id')])
       clonedPatient.save clonedPatient.toJSON(),
         success: (model) =>
           @patients.add model # make sure that the patient exist in the global patient collection
           measure.get('patients').add model # and the measure's patient collection
           if bonnie.isPortfolio then @measures.each (m) -> m.get('patients').add model
-          count++
-          bonnie.navigate "measures/#{measure.get('hqmf_set_id')}", trigger: true if count == $d.length
+          count += 1
+          perc = (count / $d.length) * 100
+          @$(".clone-patients-progress-bar").css('width', perc.toFixed() + '%')
+          if count == $d.length
+            # @$("#clonePatientsDialog").modal 'hide'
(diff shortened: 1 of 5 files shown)


andrequina:
The current measure should be removed from the list of measures you can copy to.

andrequina:
Disable the clone button if no patients are selected... currently it simply does nothing when you click it.

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -38,6 +41,37 @@ class Thorax.Views.Measure extends Thorax.View
...
+  selectMeasureListing: (e) ->
+    @$('.measure-listing').removeClass('active')
+    @$('.btn-clone-patients').hide()
+    m = @$(e.target).model()
+    @$(".measure-#{m.get('hqmf_set_id')}").addClass('active')
+    @$(".btn-clone-#{m.get('hqmf_set_id')}").show()
+
+  cloneIntoMeasure: (e) ->
+    $d = @$('.select-patient:checked')
+    measure = @measures.findWhere({hqmf_set_id: @$('.measure-listing.active').model().get('hqmf_set_id')})
```
andrequina: can't you get the measure off the event rather than relying on the class?  The clone button should be associated with a specific measure?

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -38,6 +41,37 @@ class Thorax.Views.Measure extends Thorax.View
...
+    count = 0
+    for diff in $d
+      difference = @$(diff).model()
+      patient = @patients.findWhere({medical_record_number: difference.result.get('medical_record_id')})
+      clonedPatient = patient.deepClone(omit_id: true, dedupName: true)
```
andrequina: I don't think we have to dedup the name... it's going to another measure... we are probably deduping based on this measure

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -38,6 +41,37 @@ class Thorax.Views.Measure extends Thorax.View
...
+      clonedPatient.set('measure_ids', [measure.get('hqmf_set_id')])
+      clonedPatient.save clonedPatient.toJSON(),
+        success: (model) =>
+          @patients.add model # make sure that the patient exist in the global patient collection
+          measure.get('patients').add model # and the measure's patient collection
+          if bonnie.isPortfolio then @measures.each (m) -> m.get('patients').add model
```
andrequina, line 78: good catch on portfolio!

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -38,6 +41,37 @@ class Thorax.Views.Measure extends Thorax.View
...
+          count++
+          bonnie.navigate "measures/#{measure.get('hqmf_set_id')}", trigger: true if count == $d.length
+
```
andrequina: This seems a bit odd using count to navigate like this.  I can't think of a better way off hand... maybe ask Pete or Liam?

app/assets/stylesheets/measure.less
```diff
@@ -12,6 +12,28 @@
...
+    .fa {
+      color: black;
+    }
+  }
+  .list-group {
+    padding-top: 30px;
+  }
+}
+
 .right-sidebar {
```
andrequina, line 37: Can you have Liam or Jeff take a quick look at the CSS?

andrequina:
we need to have a progress indicator that blocks the user from other tasks... we could be cloning lots of patients which could take a while... we don't want the user navigating while the cloning is happening.

jtferns:
Addressed comments except for navigation and CSS. The former would require some discussion for an alternative approach for navigating; I included a 'Cloning patients...' dialog but I had some issues getting it to hide before the navigation trigger, so I left it in but commented out.

jtferns:
Correctly navigates after cloning patients to the selected measure's view.

Ideally it would be nice to capture the following code in the view's event hash, but I am unfamiliar as to how I can do that:

```
@$("#clonePatientsDialog").on('hidden.bs.modal', -> bonnie.navigate "measures/#{measure.get('hqmf_set_id')}", trigger: true)
```

I attempted the following, but it did not function as desired:

```
'hidden.bs.modal #clonePatientsDialog': -> bonnie.navigate "measures/#{measure.get('hqmf_set_id')}", trigger: true)
```

PR #155, opened 2014-03-18
add additional screen-reader labels for increased accessibility

carpeliam:
I've added several screen-reader clarifications to hopefully make the app more accessible. Let me know if I missed any spots.

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/edit_value.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_value.hbs
@@ -1,6 +1,7 @@
 {{#if fieldValue}}
 <p>
-  <select name="key" class="form-control">
+  <label for="key_{{@cid}}" class="sr-only">field value type</label>
+  <select name="key" id="key_{{@cid}}" class="form-control">
     <option value="">--</option>
     {{#each fields}}
       <option value="{{key}}">{{title}}</option>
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/templates/patient_builder/edit_value.hbs
```diff
@@ -42,10 +46,12 @@
 
   {{#ifCond type '==' 'TS'}}
     <div class="col-md-5">
-      <input type="text" name="start_date" class="date-picker form-control" placeholder="mm/dd/yyyy" data-date-format="mm/dd/yyyy" data-date-autoclose="true">
+      <label for="start_date_{{@cid}}" class="sr-only">{{#if fieldValue}}field {{/if}}value date</label>
+      <input type="text" name="start_date" id="start_date" class="date-picker form-control" placeholder="mm/dd/yyyy" data-date-format="mm/dd/yyyy" data-date-autoclose="true">
```
pkmitre: The "start_date" id won't necessarily be unique on the page, will that be an issue? Same with the start_time id just below.

carpeliam: Presumably (and this is my presumption, not someone's who knows all the rules), the user has tabbed to here with proper context, as there are ways to determine which criteria you're editing. If they don't have the context, then it might be easier/better to give them the context up front, rather than reading the entire data criteria name for each field related to it (same deal when we have multiple codes/values/field values). But I have this question as well.

PR #158, opened 2014-03-19
fixes odd highlight issue on materialize

andrequina:
this fixes an odd highlight issue where the patient builder logic view does not display correctly after re-calculation.  I am not clear what's going on, but wrapping the sr-highlighting with a span seems to resolve the issue

### DIFF
--- a/app/assets/javascripts/views/logic/population_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/population_logic_view.js.coffee
@@ -46,17 +46,20 @@ class Thorax.Views.PopulationLogic extends Thorax.View
           for key, value of rationale
             target = @$(".#{code}_children .#{key}")
             if (target.length > 0)
-              target.addClass(if updatedRationale[code]?[key] is false then 'eval-bad-specifics' else "eval-#{!!value}")
-              target.closest('.panel-heading').addClass(if updatedRationale[code]?[key] is false then 'eval-panel-bad-specifics' else "eval-panel-#{!!value}")
-      @rationaleScreenReaderStatus()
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -46,18 +46,21 @@ class Thorax.Views.PopulationLogic extends Thorax.View
...
-      @rationaleScreenReaderStatus()
+              if updatedRationale[code]?[key] is false
+                targetClass = 'eval-bad-specifics'
+                targetPanelClass = 'eval-panel-bad-specifics'
+                srTitle = '(status: bad specifics)'
+              else
+                bool = !!value
+                targetClass = "eval-#{bool}"
+                targetPanelClass = "eval-panel-#{bool}"
+                srTitle = "(status: #{bool})"
```
carpeliam: the above lines _could_ be rewritten as

```
[targetClass, targetPanelClass, srTitle] = if updatedRationale[code]?[key] is false
  ['eval-bad-specifics', 'eval-panel-bad-specifics', '(status: bad specifics)']
else
  bool = !!value
  ["eval-#{bool}", "eval-panel-#{bool}", "(status: #{bool})"]
```

carpeliam:
@andrequina and I were able to produce this bug in Chrome only. I wonder if maybe this is a factor of how complicated the markup is, or if Chrome is applying hidden visibility to a line of text instead of to a span element. (Seems like the latter.) I'd be curious to see if we could learn more about this, but given the timeframe, it's important to get this merged in quickly.

PR #163, opened 2014-03-19
Reflect the measure period on the patient builder

carpeliam:
This highlights data criteria within the measure period by changing their border color.


app/views/layouts/application.html.erb
```diff
@@ -37,6 +37,7 @@
       bonnie.isPortfolio = <%= current_user.is_portfolio? %>;
       bonnie.isAdmin = <%= current_user.is_admin? %>;
       bonnie.currentUserId = '<%= current_user.id %>';
+      bonnie.measurePeriod = <%= Time.zone.at(APP_CONFIG['measure_period_start']).year %>;
```
andrequina, line 40: Currently we are assuming a one year measurement period.  If there are multi year measures then this would need to be addressed.

PR #164, opened 2014-03-20
Indicate change with flashing background on measure population tab switch

jtferns:
#### Story
- https://www.pivotaltracker.com/story/show/62069906
#### Notes
- Added <code>indicator-on</code> and <code>indicator-off</code> classes for rendering flashing background on measure view page elements that change between measure populations on population tab switches
#### Screenshot

screen shot 2014-03-20 at 10 42 40 am


app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -56,3 +60,7 @@ class Thorax.Views.Measure extends Thorax.View
     e.preventDefault()
     $btn = $(e.currentTarget)
     $btn.toggleClass('btn-danger btn-danger-inverse').prev().toggleClass('hide')
+
+  resetIndicator: ->
+    @$('.right-sidebar').removeClass('indicator-on').addClass('indicator-off')
```
carpeliam: You can actually call `toggleClass` with multiple classes (even if some of them are being added and others are being removed), which means you could also change the `@listenTo` lines above to:

```
@listenTo @logicView, 'population:update', (population) ->
  @resetIndicator()
  setTimeout @resetIndicator, 125
```

jtferns: I can encapsulate it in the same function, but I would have to modify <code>@resetIndicator()</code> to function as desired:

```
resetIndicator: ->
    if @$('.indicator-on').size()
      @$('.right-sidebar').toggleClass('indicator-on indicator-off')
      @$('.panel').toggleClass('indicator-on indicator-off')
    else
      @$('.right-sidebar').removeClass('indicator-off').toggleClass('indicator-on')
      @$('.panel').removeClass('indicator-off').toggleClass('indicator-on')
```

Is this reasonable?

carpeliam: Ah, are you sometimes calling `removeClass('indicator-off')` regardless of whether or not the indicator-off class is actually on an element? I'd prefer to not make logic decisions in the view that are based on markup, so I don't think that `if @$('.indicator-on').length` would be a win here.

Also, you can call `@$('.panel, .right-sidebar')` instead of separating them out.

jtferns:
Rebased and set <code>indicator-off</code> on actual template elements, simplified <code>@resetIndicator()</code> based on comments.

PR #171, opened 2014-03-21
Quick fix for rationale highlighting in patient builder logic view

jtferns:
#### Story
- https://www.pivotaltracker.com/story/show/67942242
#### Note
- the main issue is that if you just switch between populations, the patient model does not change, and therefore does not actually materialize, so the event handler doesn't fire to update the rationale highlighting in-between tab switching
- all other patient model interactions (ones that actually modify the patient model's json) behave and fire materialize events accordingly

### DIFF
--- a/app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
+++ b/app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
@@ -28,6 +28,7 @@ class Thorax.Views.PatientBuilder extends Thorax.View
     @expectedValuesView.on 'population:select', (population_index) =>
       @populationLogicView.setPopulation @measure.get('populations').at(population_index)
       @materialize()
+      @populationLogicView.showRationale @model
     @model.on 'materialize', =>
       @populationLogicView.showRationale @model
     @model.on 'clearHighlight', =>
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -28,6 +28,7 @@ class Thorax.Views.PatientBuilder extends Thorax.View
     @expectedValuesView.on 'population:select', (population_index) =>
       @populationLogicView.setPopulation @measure.get('populations').at(population_index)
       @materialize()
+      @populationLogicView.showRationale @model
```
pkmitre, line 30: Can you add a post-line comment explaining why this is here? ie something like

```
@populationLogicView.showRationale @model # Materialize event may not trigger if model hasn't changed
```

Which leads to the question... do we actually need to call @materialize() here? Will the patient ever have changed?

jtferns, line 30: Ah yes, will do.

As far as I know, nothing population-specific modifies the patient json, but rather the patient json affects the population rationale highlighting. So, we should be ok not having <code>@materialize()</code> on population switches.

pkmitre, line 30: If you take out the materialize(), we can probably take out the comment too.

jtferns:
Removed <code>@materialize()</code> and inline comment, I think it makes sense as-is that we would set the population and update the rationale highlighting on population selection.

andrequina:
It looks like this is good... going to merge.

PR #174, opened 2014-03-21
change value sets in jasmine tests to use HDS value set models rather th...

andrequina:
...an JSON


andrequina:
merging as this will fix broken tests

PR #177, opened 2014-03-21
add placeholder support for IE9

carpeliam:
This adds a placeholder polyfill for IE9, using the recommended method on http://html5please.com/#placeholder.

One issue with this PR is that it doesn't style placeholder colors any differently than regular input text. This is because there's a documented incompatibility between the polyfill and Bootstrap [https://github.com/mathiasbynens/jquery-placeholder/pull/39], that has a pull request which looks like it has the thumbs up, but hasn't been merged. I'd suggest we push on getting that PR accepted, and then update our reference to the polyfill library as appropriate, but not worry too much about color support until then.


andrequina:
merged

PR #181, opened 2014-03-23
added tests for users controller

andrequina:


andrequina:
tests will fail until bonnie_bunder pull request 20 is merged in

PR #197, opened 2014-03-26
Add simple transfer patients test in measure view spec

jtferns:


jtferns:
Will merge these changes alongside further measure view testing in c0a17e9e5145680d29b4b775cb793e65f740d5cf

PR #199, opened 2014-03-26
add screen reader labels for account form elements

andrequina:


carpeliam:
If we're using labels, then we should remove the title. And if you're feeling crazy, maybe drop the trailing colons on the labels too, but this isn't a big deal.

PR #202, opened 2014-03-27
adds titles to time elements, removes update popups from HTML after clos...

andrequina:
...e, and adds sr-only save and cancel buttons

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
+++ b/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
@@ -51,7 +51,6 @@
 
         <div class="form-group">
           <label for="birthdate" class="control-label">Date of Birth</label>
-          <label for="birthtime" class="sr-only">Time of Birth</label>
 
           <div class="row">
             <div class="col-md-6">
@@ -59,7 +58,8 @@
             </div>
 
             <div class="col-md-6">
-              <input type="text" id="birthtime" name="birthtime" class="time-picker form-control" placeholder="--:-- --" title="time" data-show-inputs="false" data-default-time="8:00 AM">
+              <label for="birthtime" class="sr-only">Time of Birth</label>
+              <input type="text" id="birthtime" name="birthtime" class="time-picker form-control" placeholder="--:-- --" data-show-inputs="false" data-default-time="8:00 AM">
             </div>
           </div>
         </div>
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/templates/patient_builder/patient_builder.hbs
```diff
@@ -51,6 +51,7 @@
 
         <div class="form-group">
           <label for="birthdate" class="control-label">Date of Birth</label>
+          <label for="birthtime" class="sr-only">Time of Birth</label>
```
carpeliam: Can we move this label to be closer to its associated content, just above the `<input>`?

carpeliam: And if we're adding a label, can we take out the corresponding title attribute - the label should replace the title, not add to it.

app/assets/javascripts/templates/patient_builder/patient_builder.hbs
```diff
@@ -94,6 +95,7 @@
               </div>
 
               <div class="col-md-5">
+                <label for="deathtime" class="sr-only">Time of Death</label>
                 <input type="text" id="deathtime" name="deathtime" class="time-picker form-control" placeholder="--:-- --" title="time" data-show-inputs="false" data-default-time="8:00 AM">
```
carpeliam: Same deal here - we shouldn't have both a label and a title.

app/assets/javascripts/views/import_measure_view.js.coffee
```diff
@@ -22,7 +22,9 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
       redirectRoute: currentRoute
 
   events:
-    rendered: -> @$("option[value=\"#{eoc}\"]").attr('selected','selected') for eoc in @model.get('episode_ids') if @model? && @model.get('episode_of_care') && @model.get('episode_ids')?
+    rendered: -> 
+      @$("option[value=\"#{eoc}\"]").attr('selected','selected') for eoc in @model.get('episode_ids') if @model? && @model.get('episode_of_care') && @model.get('episode_ids')?
+      @$el.on 'hidden.bs.modal', -> @remove()
```
carpeliam, line 27: Cool. At some point, we should probably clean up the `close` method, but we can do that later.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -64,6 +64,7 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
 
       @$('.date-picker').datepicker().on 'changeDate', _.bind(@materialize, this)
       @$('.time-picker').timepicker(template: false).on 'changeTime.timepicker', _.bind(@materialize, this)
+
```
carpeliam: why was this file modified

PR #204, opened 2014-03-27
added justification to logic statements.  It is a bit off sometimes, but...

andrequina:
... Jeff felt it was better than previous.

### DIFF
--- a/app/assets/javascripts/templates/logic/precondition.hbs
+++ b/app/assets/javascripts/templates/logic/precondition.hbs
@@ -1,5 +1,5 @@
 <ul id="precondition_{{precondition.id}}">
-  <li>
+  <li{{#unless precondition.preconditions }} class="criteria-{{parentPrecondition.conjunction_code}}"{{/unless}}>
     <span class="conjunction {{parentPreconditionKey}} rationale-target">
       {{translate_conjunction parentPrecondition.conjunction_code}}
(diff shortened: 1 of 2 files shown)


jtferns:
If some of the 'off' was regarding negated data criteria elements, I pushed up some changes accounting for negation indentation on a separate branch based off of this one here: 7dfed4e34b7cb7f34dbc60497e41ecc318d46c83

carpeliam:
Maybe this could be accomplished with a [horizontal data definition list](http://getbootstrap.com/css/#type-lists)? I can take a pass at this if you'd like.

thecristen:
So what's going on with this? As far as I can tell the result on both the `justify_data_criteri_text` and `tweak_text_indent` branches are identical (and look good!). What are we aiming for/what's stopping this pull request from going through?

carpeliam:
Sorry, @andrequina and I ended up having an offline conversation about this. The issue is that we're trying to line up content with a hanging indent, which would work if all we wanted to do was indent by some number of pixels, but the reason for the different indent amounts is because we want to indent by some number of characters... which is font-dependent, and we may have different fonts displaying in different ways on different browsers/environments. So it might be better to us a column-based approach to laying the text out, a la the DL idea (or something roughly equivalent, or something even better?).

pkmitre:
No outstanding requirement to tweak the justification, so closing without merging.

PR #206, opened 2014-03-27
Jasmine view tests for measure and dashboard view

jtferns:


spec/javascripts/views/measure_view_spec.js.coffee
```diff
@@ -2,8 +2,47 @@ describe 'MeasureView', ->
...
-    expect(@measureView.$el).toContainText @measure.get('description')
+  afterEach ->
+    @measureView.remove()
+
+  it 'renders measure details correctly', ->
+    expect(@measureView.$('.measure-title')).toContainText @measure.get('title')
+    expect(@measureView.$('.measure-title')).toContainText @measure.get('cms_id')
+    expect(@measureView.$('.measure-dsp')).toContainText @measure.get('description')
+    expect(@measureView.$('[data-call-method="measureSettings"]')).toExist()
+    expect(@measureView.$('[data-call-method="patientsSettings"]')).toExist()
```
pkmitre: These tests feel a bit specific and brittle, ie it feels like we're testing "the app works in this specific way that we happened to implement" rather than "the app works correctly". 

```
expect(@measureView.$('.measure-title')).toContainText @measure.get('title')
```

could become

```
expect(@measureView.$el).toContainText @measure.get('title')
```

spec/javascripts/views/measure_view_spec.js.coffee
```diff
@@ -2,8 +2,47 @@ describe 'MeasureView', ->
...
     @measureView = new Thorax.Views.Measure(model: @measure, patients: @measure.get('patients'))
     @measureView.render()
+    @measureView.appendTo 'body'
 
-  it 'renders correctly', ->
-    expect(@measureView.$el).toContainText @measure.get('description')
+  afterEach ->
+    @measureView.remove()
+
+  it 'renders measure details correctly', ->
```
carpeliam: I generally take issue with the word "correctly" here, because it's an implicit contract. The title of a test should explicitly say what's required. In this case, just removing the word "correctly" is fine, because all you should be doing here is verifying that the measure details are rendered.

spec/javascripts/views/measure_view_spec.js.coffee
```diff
@@ -2,8 +2,47 @@ describe 'MeasureView', ->
...
 
-  it 'renders correctly', ->
+  afterEach ->
+    @measureView.remove()
+
+  it 'renders measure details', ->
+    expect(@measureView.$el).toContainText @measure.get('title')
+    expect(@measureView.$el).toContainText @measure.get('cms_id')
     expect(@measureView.$el).toContainText @measure.get('description')
+    expect(@measureView.$('[data-call-method="measureSettings"]')).toExist()
```
carpeliam: If you wanted to change the name of the method, would you want this test to fail?

andrequina:
@jtferns can you rebase this?  I am getting a conflict and I don't want to do a bad merge.

jtferns:
Rebased on e67c755ac0644e5e0d8c1dc284bcb0725bdb6624, tests passed locally, it should be all set after Travis tests are done.

PR #222, opened 2014-04-02
added language to HTML for RI-3

andrequina:


andrequina:
this was already on the devise layout

PR #230, opened 2014-04-03
Sort user listings in admin view by approved status

jtferns:
#### Note
- sorts user listings by <code>approved</code>, such that unapproved users are at the top, and then sorts the respective groups alphabetically based on email address (original sorting criteria)
#### Screenshot

screen shot 2014-04-03 at 4 19 28 pm


andrequina:
It looks like the commit suggests it is sorting by approved and then email?  Where is the email sort happening?

jtferns:
@andrequina I believe it already does this after sorting by <code>approved</code>; I attached a screenshot to the pull request showing this, as well.

jtferns:
It was doing it implicitly before, probably based on default collection sorting; I changed this to explicitly sort by <code>approved</code> and then by <code>email</code>.

PR #234, opened 2014-04-09
add the ability to load HQMF xml files directly in the measure load dial...

andrequina:
...og

### DIFF
--- a/app/assets/javascripts/templates/import/import_measure.hbs
+++ b/app/assets/javascripts/templates/import/import_measure.hbs
@@ -14,10 +14,34 @@
             <input name="_method" type="hidden" value="post">
             <input name="authenticity_token" type="hidden" value="{{token}}">
           </div>
+          {{#if showLoadInformation}}
+            <div class="alert alert-warning">
+              Measure Data file types can include: 
+              <ul>
+                <li><a href="https://www.emeasuretool.cms.gov/" target="_blank">Measure Authoring Tool Zip file export</a></li>
+                <li><a href="http://www.cms.gov/Regulations-and-Guidance/Legislation/EHRIncentivePrograms/eCQM_Library.html" target="_blank">HQMF XML File (eCQM specifications)</a></li>
+              </ul>
+            </div>
+          {{/if}}
           <div class="form-group">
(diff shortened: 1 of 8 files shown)


andrequina:
this requires an updated bonnie bundler

pkmitre:
Tests are failing, but that looks to be because this pull request doesn't point to the updated bonnie bundler yet.

app/assets/javascripts/templates/import/import_measure.hbs
```diff
@@ -14,12 +14,36 @@
...
           </div>
+          {{#if showLoadInformation}}
+            <div class="alert alert-warning">
+              Measure Data file types can include: 
+              <ul>
+                <li><a href="https://www.emeasuretool.cms.gov/" target="_blank">Measure Authoring Tool Zip file export</a></li>
+                <li><a href="http://www.cms.gov/Regulations-and-Guidance/Legislation/EHRIncentivePrograms/eCQM_Library.html" target="_blank">HQMF XML File (eCQM specifications)</a></li>
+              </ul>
+            </div>
+          {{/if}}
```
pkmitre, line 25: Seems like it would be useful to always have this information present.

app/assets/javascripts/views/import_measure_view.js.coffee
```diff
@@ -24,12 +25,29 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
...
+      username.closest('.form-group').removeClass('has-error')
+      hasUser = true
+    if (password.val().length > 0) 
+      password.closest('.form-group').removeClass('has-error')
+      hasPassword = true
+    @$('#loadButton').prop('disabled', !(hasUser && hasPassword)) 
 
   enableLoad: ->
-    @$('#loadButton').prop('disabled', !@$('input:file').val().length > 0)
+    if /^.*\.xml$/.test(this.$('input:file').val().toLowerCase())
```
pkmitre: This seems a bit awkward, could be

```
if @$('input:file').val().match /xml$/i
```

pkmitre: Also, these XML files are distributed in zip files; wouldn't it make sense for these zip files to be uploadable? I know that makes it harder to test if a VSAC password is needed...

PR #236, opened 2014-04-10
removes x from finalize dialog and fixes typo in bad file error message

andrequina:


pkmitre:
Included these same changes in #234

PR #242, opened 2014-04-11
change bundle measure ids from nqf to cms and default to using hqmf_id (...

andrequina:
...cms_id == false)


andrequina:
requires corresponding bonnie_bunder update

andrequina:
#247 pulls in the required updates to bonnie_bundler

PR #253, opened 2014-04-23
Indicate that IE8 is not supported

jtferns:
_Note: This branch is pending IE8 tests._
#### Story
- https://www.pivotaltracker.com/story/show/69123508
#### Note
- added conditional IE8 tags to main application header and to devise pages

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -1,5 +1,3 @@
-<%= render :partial => "devise/shared/ie8_alert" %>
-
 <h2>Resend confirmation instructions</h2>
 
 <%= form_for(resource, :as => resource_name, :url => confirmation_path(resource_name), :html => { :method => :post }) do |f| %>
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -1,3 +1,5 @@
+<%= render :partial => "devise/shared/ie8_alert" %>
+
 <h2>Resend confirmation instructions</h2>
 
 <%= form_for(resource, :as => resource_name, :url => confirmation_path(resource_name), :html => { :method => :post }) do |f| %>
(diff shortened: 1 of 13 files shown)


app/views/devise/confirmations/new.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= render :partial => "devise/shared/ie8_alert" %>
```
carpeliam: Why not put this in the `devise` layout?

app/views/devise/shared/_ie8_alert.html.erb
```diff
@@ -0,0 +1,7 @@
+<!--[if IE 8]>
+<div class="alert alert-danger">
+  <center>
```
carpeliam: The `center` tag is deprecated. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center

app/views/layouts/_header.html.erb
```diff
@@ -1,3 +1,10 @@
+<!--[if IE 8]>
```
carpeliam: This looks identical to the warning I see in the devise layout. Can we have one partial that we include from both layouts?

app/views/layouts/_ie8_alert.html.erb
```diff
@@ -0,0 +1,5 @@
+<!--[if IE 8]>
+<div class="alert alert-danger" style="text-align: center;">
```
carpeliam: Inline styles are also not cool ;) but we can let it go...

carpeliam:
With this implementation, IE7 users wouldn't receive any notice. Do we want to let IE users know if they fail to meet the minimum standards, regardless of which version?

app/views/layouts/_ie8_alert.html.erb
```diff
@@ -0,0 +1,5 @@
+<!--[if lte IE 8]>
+<div class="alert alert-danger ie8-alert">
+  Your web browser (IE8 or older) is not supported. Please switch to another browser to continue using Bonnie.
```
carpeliam: I'd also recommend giving them a useful call to action. Maybe:

> Please [update your browser](http://www.updateyourbrowser.net/) to continue using Bonnie.

with a link to http://www.updateyourbrowser.net/.

PR #264, opened 2014-05-08
Remove redundant measure title, centered previously off-centered measure title

thecristen:
### Stories
- https://www.pivotaltracker.com/story/show/68458794
- https://www.pivotaltracker.com/story/show/68458728
### Notes
- Removed where it had a redundant "Measure" title. 
- Placed submeasure to go in the measure header
- Centered the measure title on the results page vertically relative to the grey bar
### Screens!

centered

headerwithsub

My first pull request here! Please forgive the back-and-forth-reverting silliness, I'll keep to my own branch now on. :angel:

### DIFF
--- a/app/assets/javascripts/templates/measure.hbs
+++ b/app/assets/javascripts/templates/measure.hbs
@@ -1,7 +1,7 @@
  <div class="main col-sm-8">
   <div class="measure-title">
-    <span class="short-title pull-left"><i class="fa fa-tasks"></i> {{cms_id}}:</span>
-    <span class="full-title">{{title}}</span>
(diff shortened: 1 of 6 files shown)


jtferns:
Some measure population titles can be rather long (e.g., CMS137v2), and the <code>float: right;</code> causes the IPP logic block to wrap awkwardly:
screen shot 2014-05-09 at 10 31 26 am

Can we try reducing the font-size or letting the population title wrap by removing the <code>float: right;</code>?

thecristen:
Thanks @jtferns, I hadn't realized the names could get so long (after following the demo script I only have short names!). So that should be good now.

Also added a fix for https://www.pivotaltracker.com/story/show/68458940 which also manages to get rid of the extra offending line break.

PR #268, opened 2014-05-13
Added script allowing the required fields icons to show error

thecristen:
### Story
- https://www.pivotaltracker.com/story/show/68443700
### Notes
- It only acts when there's a `field_with_errors` on the page. 
- I love jQuery!
- One thing I noticed: The last field (confirm password) never shows as having an error, which strikes me as strange, but I guess that's an artifact of devise?
### Screenshot

iconvalidation

### DIFF
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -99,8 +99,6 @@
       $submit.prop('disabled', !$checkbox.is(':checked'));
     });
 
-    $('.field_with_errors').each(function() {
-      $(this).parent().removeClass('required').addClass('has-error');
-    }); //allows icon to change color appropriately for validation
+    $('.field_with_errors').parent().removeClass('required').addClass('has-error'); //allows icon to change color appropriately for validation
   });
 </script>
--- a/app/views/devise/registrations/new.html.erb
+++ b/app/views/devise/registrations/new.html.erb
@@ -98,5 +98,9 @@
           $submit = $form.find(':submit');
       $submit.prop('disabled', !$checkbox.is(':checked'));
     });
+
+    $('.field_with_errors').each(function() {
+      $(this).parent().removeClass('required').addClass('has-error');
+    }); //allows icon to change color appropriately for validation
   });
 </script>
--- a/app/assets/stylesheets/bootstrap3.less
+++ b/app/assets/stylesheets/bootstrap3.less
@@ -139,3 +139,10 @@ h2 {
 .popover {
   box-shadow: none;
 }
+
+//hack that Twitter uses to fix input field widths in IE9
+.input-group {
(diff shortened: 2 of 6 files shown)


thecristen:
After talking about it a bit with @carpeliam I think the ideal would be to rejigger devise to handle the validation a bit differently in its treatment of the markup, whereas this commit would be more of a quick fix.

jtferns:
If you're refactoring your approach, I would also like to add that this would be applicable in two other locations as well:
- account edit page (https://github.com/projecttacoma/bonnie/blob/master/app/views/devise/registrations/edit.html.erb)
- forgot password page (https://github.com/projecttacoma/bonnie/blob/master/app/views/devise/passwords/new.html.erb)

app/views/devise/registrations/new.html.erb
```diff
@@ -98,5 +98,9 @@
           $submit = $form.find(':submit');
       $submit.prop('disabled', !$checkbox.is(':checked'));
     });
+
+    $('.field_with_errors').each(function() {
```
carpeliam: Just so you know, jquery will perform operations on an entire set, so you don't need to do an operation on each individual item. `$('.field_with_errors').parent().removeClass('required').addClass('has-error')` would probably have the same result.

thecristen:
That's good to know.

For refactoring, from what I can tell [this blog post](http://blog.michaelgreenly.com/2011/12/formatting-rails-errors-for-twitter.html) probably explains clearest what is going on and a decent approach to fix it. Also see [this gist](https://gist.github.com/jamiepenney/1829710) and [this gem which might be useful for this](https://github.com/dabit/bootstrap-form). That probably deserves its own story and someone better at ruby than I am. :ok_hand: 

I'm going to try @carpeliam's approach and then apply it on the pages @jtferns mentioned so watch out :tada:

jtferns:
On the forgot password page, the previously offset <code>div</code> (almost centered) is now appearing left-aligned. I think the <code>div</code> should be updated with a <code>center-block</code> class, too, as is done for the login page ( https://github.com/projecttacoma/bonnie/blob/registration_pg_icons/app/views/devise/sessions/new.html.erb#L1 ).

carpeliam:
I'd vote for putting `.center-block` in the stylesheet as part of the login class, rather than copying `.center-block` to all of the places that reference the login class. @thecristen's [change](https://github.com/projecttacoma/bonnie/commit/28c61fd667d9306829c1c5153247c9e44325cdee#diff-2980e1772cbf71db8ba951c38e580058L5) was to the login class itself, so we should keep the intent in the same place.

thecristen:
Oh right, I'm using LESS so I can do things like that! 
Added a commit incorporating @carpeliam's suggestion, so the three pages using the login class are centered.

PR #271, opened 2014-05-14
Port LESS to SASS

dmpotter44:
Port from LESS to SASS


app/assets/stylesheets/_bootstrap3.scss
```diff
@@ -1,25 +1,22 @@
...
+$gray-lighter: #f6f6f6;
+$gray-light: #e2e2e2;
+$gray-darker: #888;
 
-@brand-primary: #0075C4;
-@gray-lighter: #f6f6f6;
-@gray-light: #e2e2e2;
-@gray-darker: #888;
+$border-radius-base: 0px;
+$border-radius-large: 0px;
```
carpeliam, line 9: It looks like this isn't actually taking effect when I run this in a live server; other things like `$brand-primary` seem to work, but buttons and other elements seem to have the default Bootstrap border radius.

There are a few other things that look different than expected too, like the accordion on a measure page, or the text-decoration on the logo when hovering. I'm guessing it's probably something small and central that's causing each of these issues, it might be helpful to run the LESS and SASS codebases in parallel and compare.

dmpotter44, line 9: I doubt it's these lines specifically that there are having issues, but rather one major issue when moving between LESS and SASS: LESS basically allows any named class to be used as a mixin, SASS does not. However, SASS allows you to "extend" an existing class. This is semantically very similar (a new class takes on the characteristics of an earlier defined class) but it's implemented in a different and quite possibly non-compatible way.

This isn't a problem for anything in our own LESS - we can just rewrite things we want to use as mixins as proper SASS mixins. It becomes a problem when using the SASS version of Bootstrap - several Bootstrap classes were used as mixins in the original LESS version, and I can't duplicate that in the new version without copying over the Bootstrap classes, which seems to me to defeat the purpose of using Bootstrap as an external library.

Hopefully the issues can be resolved without directly copying in Bootstrap CSS, but it's entirely possible that what you're seeing is a slight change in rule priority order caused by the change from LESS mixins to SASS extends.

carpeliam, line 9: That might also be an issue, but it isn't the issue I'm calling out here; the variable declarations here shouldn't interfere with any mixins etc. The [bootstrap variables are declared with `!default`](https://github.com/twbs/bootstrap-sass/blob/master/vendor/assets/stylesheets/bootstrap/_variables.scss#L106), so if you define the variable before importing bootstrap, our values should stick. If you do this:
1. define variables to be used with bootstrap
2. import bootstrap
3. do any other theming you couldn't do by changing variable values

you should be in good shape. Moving the bootstrap import from `application.css.scss` into the middle of `_bootstrap3.scss` (below variable declarations, above everything else) should work.

dmpotter44, line 9: The latest version tries that, I'm not entirely sure if it made a difference because I'm not entirely clear on where, exactly, the variables are used. Plus if I understand LESS's stance on not including a "!default" feature, this shouldn't impact the way variables are declared?

I'm actually a bit unclear on when a variable's value is used. I always kind of assumed LESS and SASS compiled "as they go" so if you changed a variable's value midway through the script the new value would only take effect on further styles and its original value would be used above, but I actually have no idea if that's the case.

carpeliam, line 9: The `_bootstrap3` file is our bootstrap theme, so any of the variables declared in this file should be [bootstrap-specific](http://getbootstrap.com/customize/#less-variables). (Basically, we've got this pattern: https://github.com/twbs/bootstrap-sass#sass, third paragraph down, "You can override these by simply redefining the variable before the `@import` directive, e.g....") Anything below the variables should be an override of a bootstrap-specific style, so those would have to go below the bootstrap import. So the _bootstrap3.scss file should look like this:

```
$brand-primary: #0075C4;
// ...
$font-family-sans-serif: "Helvetica Neue", Roboto, Helvetica, Arial, sans-serif;

@import "bootstrap-sass/lib/bootstrap";

// Fonts
@font-face { /* ... */ }
// ...
```

As far as how LESS/SASS work internally, I hear you, I haven't seen too much in their documentation about this and haven't really dug in to see. All I'm going by so far is just experience and the GitHub page for bootstrap-sass.

bower.json
```diff
@@ -4,7 +4,7 @@
   "dependencies": {
     "thorax": "~2.0.1",
     "jquery": "~2.0.3",
-    "bootstrap": "~3.0.0",
+    "bootstrap-sass": "~3.0.0",
```
carpeliam, line 7: I just noticed that `bootstrap-sass` does not point to the official one, as mentioned on https://github.com/twbs/bootstrap-sass#d-nodejs--bower; this should be changed to `bootstrap-sass-official`.

pkmitre:
A recent less dependency update has addressed the issues that were prompting this conversion.

PR #273, opened 2014-05-16
Add support for Ends Concurrent with Start / Starts Concurrent with End

pkmitre:
When rendering a measure, correctly expand ECWS and SCWE into the appropriate readable text

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,9 +1,9 @@
 GIT
   remote: https://github.com/projectcypress/health-data-standards.git
-  revision: 1992aa22c47b8e9a8a5e53150de70dfd9322a1e9
+  revision: 3dccdd2d93da67bfecd0fd93caaca57635f4d02c
   branch: master
   specs:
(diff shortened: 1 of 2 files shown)


jtferns:
Do you have a measure that can be used to test this?

pkmitre:
The IPP of CMS56v3 contains a concurrent with measure period clause

jtferns:
Testing this with the final [CMS56v3 bundle](http://jira.oncprojectracking.org/secure/attachment/16660/CMS56v3_Final_5.2.14.zip) produced the following from the IPP logic:
- `ends after start of "Measurement Period"` from the human readable now maps correctly
- `ends concurrent with "Measurement Start Date"` from the human readable maps to `ends concurrent with "Measurement Period"`, which reads incorrectly in the population logic view

Is the Measurement Start Date mapping issue addressable within this story?

pkmitre:
Looks like I neglected to push a commit that updates the version of health-data-standards, which contains the bulk of the fix; I've done that now.

PR #277, opened 2014-05-19
Start stop

jmccloud:


app/assets/javascripts/templates/error_dialog.hbs
```diff
@@ -1,5 +1,5 @@
-<div class="modal fade" id="errorDialog" tabindex="-1" role="dialog" aria-labelledby="modalTitle" aria-hidden="true">
-  <div class="modal-dialog">
+<div class="modal fade" id="errorDialog" tabindex="-1" aria-hidden="true">
+  <div class="modal-dialog" role="dialog aria-labelledby="modalTitle">
```
jtferns: I think the <code>role</code> attribute value should be wrapped with an end-quote for this <code>div</code> element.

PR #279, opened 2014-05-21
Enable date shifting a user's patient records via rake task

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/63411726
### Notes
- added `rake bonnie:patients:date_shift` rake task with `EMAIL, DIR, SECONDS, MINUTES, HOURS, DAYS, WEEKS, MONTHS, YEARS` arguments
  - Use `EMAIL` to denote the user to scope the date_shift for all associated patients' source data criteria; first user by default.
  - Use `DIR` to denote direction [forward, backward]; direction is forward by default.
  - Use `SECONDS, MINUTES, HOURS, DAYS, WEEKS, MONTHS, YEARS` to denote time offset [###]; offsets are 0 by default.
- updated Patient Data Criteria model to use `bonnie.measurePeriod` when specifying the default dates for a dropped data criteria in the patient builder
- passed in bonnie.yml's `measure_start_period` into export action for the users_controller when exporting a user bundle within the admin users view
- specified an `effective_date` variable in the measure javascript containing bonnie.yml's `measure_start_period` (with computation for the end date)

### DIFF
--- a/app/assets/javascripts/models/data_criteria.js.coffee
+++ b/app/assets/javascripts/models/data_criteria.js.coffee
@@ -12,7 +12,7 @@ class Thorax.Models.MeasureDataCriteria extends Thorax.Model
     new Thorax.Models.PatientDataCriteria attr
 
   getDefaultTime: ->
-    parseInt(moment.utc().set('year', 2012).set('hour',8).set('minute',0).set('second',0).format('X'))*1000
+    parseInt(moment.utc().set('year', bonnie.measurePeriod).set('hour',8).set('minute',0).set('second',0).format('X'))*1000
 
 class Thorax.Collections.MeasureDataCriteria extends Thorax.Collection
   model: Thorax.Models.MeasureDataCriteria
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -52,7 +52,7 @@ class Admin::UsersController < ApplicationController
 
   def bundle
     user = User.find(params[:id])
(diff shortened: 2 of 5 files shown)


pkmitre:
Shouldn't we shift the patient date of birth and date of death as well? Those generally interact with the measurement period.

jtferns:
Ah right, they can and probably should be shifted accordingly, as well.

andrequina:
We also need to shift any fields that are date related... i.e., admission date/time, departure date/time, incision date/time, etc.  The fields should have a field_type of timestamp or nested_timestamp.

see:

https://github.com/projectcypress/health-data-standards/blob/master/lib/hqmf-model/data_criteria.rb

jtferns:
Patient data criteria field values with timestamp values are now properly date-shifted via 911ef7dad743e4e5542bfc6022e8ae1737d7a836

pkmitre:
It looks like adding 1.year actually adds 365.25 days, which isn't exactly what we want because it changes the time; we want the time to stay the same. We'll have to figure out another approach that accounts for leap year.

jtferns:
a50263fcebba35a60870ab336738913cf7b60197 uses <code>Time.advance</code> instead of the <code>+</code>/<code>-</code> operators for shifting the dates, which are more [precise](http://api.rubyonrails.org/classes/Time.html#method-i-advance) and should not affect the other time sections (the only exception is a leap year, where 02/29/2012 shifts to 02/28/2013 for <code>YEARS=1</code>, etc.).

PR #282, opened 2014-05-23
Validate dates in patient and data criteria

pkmitre:
Compare dates between patient and patient data criteria to make sure encounters happen while patient is alive, patient does not die before being born, etc.

### DIFF
--- a/app/assets/javascripts/models/patient.js.coffee
+++ b/app/assets/javascripts/models/patient.js.coffee
@@ -130,12 +130,43 @@ class Thorax.Models.Patient extends Thorax.Model
 
   validate: ->
     errors = []
-    errors.push ['first', 'Name fields cannot be blank'] unless @get('first').length > 0
-    errors.push ['last', 'Name fields cannot be blank'] unless @get('last').length > 0
-    errors.push ['birthdate', 'Date of birth cannot be blank'] unless @get('birthdate')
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/models/patient.js.coffee
```diff
@@ -130,12 +130,43 @@ class Thorax.Models.Patient extends Thorax.Model
...
+    if @get('deathdate') && moment(@get('deathdate'), 'X').year() < 100
+      errors.push [@cid, 'deathdate', 'Date of death must have four digit year']
+    if @get('deathdate') && @get('birthdate') && @get('deathdate') < @get('birthdate')
+      errors.push [@cid, 'deathdate', 'Date of death cannot be before date of birth']
+
+    @get('source_data_criteria').each (sdc) =>
+      # Note that birth and death dates are stored in seconds, data criteria dates in milliseconds
+      unless sdc.get('start_date')
+        errors.push [sdc.cid, 'start_date', "#{sdc.get('title')} must have start date"]
+      unless sdc.get('end_date')
```
jtferns: This will trigger errors for patient data criteria referencing diagnoses, etc., that have an undefined end date.

pkmitre: Removed the offending test; also fixed an incorrect test where it was looking at the wrong date.

PR #284, opened 2014-05-23
Email all users admin feature

dmpotter44:
This add the "email all users" admin feature to Bonnie. It provides a simple UI for sending an email to all users.

Right now it uses the BCC field to include all user addresses. If Bonnie starts to get a lot of users there may need to be considerations on sending emails in batches or handling this in some other way, but for a small number of users, this (should) work.


pkmitre:
We already have hundreds of users on our production system, so we should probably move to sending multiple emails.

dmpotter44:
Two questions on sending multiple emails: First, would it make sense to send one email per user, so that each email is unique? (We could even allow simple username substitutions in the email subject/body in that case.)

The second question is what the maximum rate of sending should be to prevent being flagged as spam. I'm assuming most email servers have a flood control, and we'd want to stay below that. Right now I plan on making that value configured in the Ruby config, but I suppose it may make sense to expose it in the admin interface?

pkmitre:
One email per user is fine and probably the easiest, and we don’t need to rate limit the emails.

pkmitre:
Merged via 7aac643

PR #292, opened 2014-06-02
Hide patient builder datepickers on input navigation

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/71504922
### Notes
- using a <code>focus input</code> event on the patient builder view, hide any open <code>.date-picker</code>'s if the <code>.datepicker</code> class is visible and the current focused input does not have a <code>date-picker</code> class
- this is a prominent issue when attempting to navigate the patient builder form(s) with arrow keys (e.g., with the JAWS screen reader application)


andrequina:
currently closing on blur makes it impossible to use the date picker.

jtferns:
8c1306506ff474638271b0c77660ef38c220a698 addresses this issue by using a <code>focus</code> event and checking for the appropriate class before hiding the calendar

andrequina:
this is not working for data criteria start and end dates and it's not working for date based fields

jtferns:
cb57b62 should account for patient data criteria <code>.datepicker</code>'s, including field values, and it should hide on inputs or selects (using <code>.form-control</code> rather than input as the event target)

PR #296, opened 2014-06-04
patient builder let SR know about changes

thecristen:
### Story
- https://www.pivotaltracker.com/story/show/71501772
### Notes
- Works in JAWS well
- This is a start! 
  - Will this cover all the cases where we need to tell the screen reader about updates?
  - Will this give too many false positives? Does it matter?


jtferns:
Verified functionality on Win7 IE9 with JAWS; notifications trigger whenever a patient field or expected value input is blurred and when patient data criteria is added.

I think all relevant cases are covered (the only current outlier is the population logic tab switch), and I think there aren't any false positives since the point of action is on the <code>patient.materialize()</code> success callback.

PR #297, opened 2014-06-04
Fix that allows additional criteria to specify results

pkmitre:


jtferns:
<code>laboratory test, result</code> data criteria should have values, as well.
The title includes result, but the [HDS data_criteria.json definition](https://github.com/projectcypress/health-data-standards/blob/master/lib/hqmf-model/data_criteria.json#L576) doesn't, so maybe it should be included as an exception in the array of definitions?

PR #298, opened 2014-06-04
User can delete own account

dmpotter44:
This is a very simplistic implementation of allowing a user to delete their own account. It does ask for confirmation before deleting, but it uses the built-in "data-confirm" system to display a JavaScript "confirm" dialog asking for confirmation. I'm not sure I really like that method since it seems like it would make deleting your account too easy.

I'm also not entirely sure that this properly propagates deleting all associated data - it just uses Devise's built-in delete support.

However, I figured I might as well open the pull request so there's a starting point for getting everything ready to go.


dmpotter44:
The pull request has been updated to add a page to the process, so instead of instantly deleting your account if you OK through the `window.confirm`, it instead brings you to a new page that forces you to hit a checkbox to confirm that you really do, in fact, want to destroy your user account.

pkmitre:
Implemented differently via #335

PR #304, opened 2014-06-06
Have SR read that this is a drop down menu

thecristen:
Story: https://www.pivotaltracker.com/story/show/71505088

The story cites that JAWS is not reading the value type scalar elements. It's actually not reading either of the SelectBoxIt-styled dropdowns at all. (It _seems_ to read "Code Set" but is actually reading a hidden label rather than the visible `<option>` value.)

One way to get around this, which I've done here, is give both dropdowns a label that'll tell the SR that this is a dropdown. It's an improvement over not indicating the presence of a dropdown, but more might be needed to make it accessible..


jtferns:
Rebased on 9d6aa459641fe70f8e813ce2613ce538e49005f0, changed SelectBoxIt behavior on patient data criteria to utilize native selects (via <code>native: true</code>) to have screen readers read out the select box options on navigation.

PR #305, opened 2014-06-06
update dependencies and add measures yml for annual update

andrequina:


config/measures/measures_2_5_0.yml
```diff
@@ -0,0 +1,720 @@
+measures: &measures
+  title: Meaningful Use Stage 2 Quality Bundle
+  version: 2.4.0
```
jtferns: Is the version intentionally specified as 2.4.0?

PR #309, opened 2014-06-10
ruby debugger is no longer supported for Ruby 2.0, switching to byebug i...

andrequina:
...nstead is recommended

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,7 +46,7 @@ group :test, :development, :ci do
 end
 
 group :test, :development do
-  gem 'pry-debugger'
+  gem 'pry-byebug'
   gem 'thin'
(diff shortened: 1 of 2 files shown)


jtferns:
Checking out this branch and attempting a <code>bundle install</code> results in 

```
Gem::InstallError: byebug requires Ruby version >= 2.0.0.
An error occurred while installing byebug (2.7.0), and Bundler cannot continue.
Make sure that `gem install byebug -v '2.7.0'` succeeds before bundling.
```

My dev environment is currently running <code>ruby 1.9.3p448</code>.
Should we move to supporting just Ruby 2.0.0 (via bundler scoping?)?

PR #314, opened 2014-06-16
Use reasonable default date for medication fulfillments

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/72793588
### Notes
- fixed issue with non-UTC moment usage in <code>serialize()</code> view method
- added default date method for populating fulfillment <code>dispense_date</code> and <code>dispense_time</code> based on the following logic
  - if the patient data criteria has a <code>start_date</code>
    - if the patient data criteria has at least one fulfillment
      - if the patient data criteria has both <code>dose_value</code> and <code>frequency_value</code>
        - then compute the number of days given the <code>dose_value</code>, <code>frequency_value</code>, <code>quantity_dispensed_value</code>, and fulfillment <code>dispense_datetime</code> and use that as the offset
      - else set the offset to 15 minutes after the latest fulfillment <code>dispense_date</code>
    - else set the offset to 15 minutes after the <code>start_date</code>

### DIFF
--- a/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
+++ b/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
@@ -220,7 +220,8 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
     serialize: (attr) ->
       if dispenseDate = attr.dispense_date
         dispenseDate += " #{attr.dispense_time}" if attr.dispense_time
-        attr.dispense_datetime = moment(dispenseDate, 'L LT').format('X')
+        attr.dispense_datetime = moment.utc(dispenseDate, 'L LT').format('X')
+    rendered: -> @setDefaultDate()
 
   initialize: ->
     @model = new Thorax.Model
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -230,13 +231,32 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
+    # use the latest fulfillment as a template
+    latest_fulfillment = @fulfillments.max((f) -> f.get('dispense_datetime')) if @fulfillments.length
+    # if dosage and frequency are specified then compute a CMD offset (derived from HQMF2JS Patient API Extension)
+    if @criteria.get('dose_value') and @criteria.get('frequency_value')
+      switch @criteria.get('frequency_unit')
+        when 'h'
+          days = 24
+        when 'd'
+          days = 1
+      dosesPerDay = days / @criteria.get('frequency_value')
```
pkmitre: The logic here seems correct, but the "days" variable name makes it confusing (is it really 24 days?); can't think of a better name offhand, but maybe removing the variable name entirely would help, ie

```
switch @criteria.get('frequency_unit')
  when 'h'
    dosesPerDay = 24 / @criteria.get('frequency_value')
  when 'd'
    dosesPerDay = days / @criteria.get('frequency_value')
```

jtferns: It might just be <code>hours</code>, but only within a CMD context. I opted for the condensed <code>switch</code> block in the latest commit.

PR #315, opened 2014-06-17
Set the height for native select clickable area with SelectBoxIt

jtferns:
### Note
- fix the SelectBoxIt height when using <code>"native": true</code> to fix clickable area triggering select dropdown
  - bootstrap <code>form-control</code> uses <code>30px</code> for select dropdown height

### DIFF
--- a/app/assets/stylesheets/builder.less
+++ b/app/assets/stylesheets/builder.less
@@ -306,5 +306,5 @@
   }
 }
 .selectBoxIt-native {
-  height: 30px !important;
+  height: @input-height-base !important;
 }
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/builder.less
```diff
@@ -306,6 +306,10 @@
   }
 }
 
+.selectBoxIt-native {
+  height: 30px !important;
```
pkmitre: Would it make sense to use the variable from the bootstrap less file (@input-height-base I think)?

thecristen: That variable is for the text input box rather than a select drop down, so, not especially.

pkmitre: That variable is used to set the height for the .form-control class, which I see being used on select boxes?

PR #317, opened 2014-06-20
Qdm v4 updates

andrequina:


app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -51,6 +51,7 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
         categories[type] ||= new Thorax.Collection
         categories[type].add criteria unless categories[type].any (c) -> c.get('description') == criteria.get('description')
     _(categories).omit('transfers')
+    _(categories).omit('derived')
```
jtferns, line 54: `_().omit()` should take multiple [keys](http://underscorejs.org/#omit), so you should be able to simplify this with the above line: `_(categories).omit('transfers','derived')`

PR #272, opened 2014-05-15
Import measure ui

thecristen:
### Story
- https://www.pivotaltracker.com/story/show/67995626
### Notes
- I used a mockup from December as guidance
- Updated file picker
- Switches for measure type and calculation radio buttons
- Changes the rebuild patients toggle to a checkbox, and gives it a nice style
- Shows rebuild patients only when updating
- Decently responsive across all browsers
### Screenshots

The old UI:
old import measure UI

The new:
new import/edit measure UI 1
new import/edit measure UI 2
new import/edit measure UI 3
new import/edit measure UI 4
new import/edit measure UI 5

PR #320, opened 2014-06-20
Import measure ui

thecristen:
Next iteration of https://github.com/projecttacoma/bonnie/pull/272
### Story
- https://www.pivotaltracker.com/story/show/72788640
### Notes
- Updated style to match the design throughout bonnie
- Accessible, changes values with keyboard actions, speaks OK in JAWS
### To-do
- Need to verify that rebuilding patients actually works (replaced a checkbox with a pair of radio buttons, want to make sure the value goes through OK)
### Screenshots

screen shot 2014-06-20 at 18 11 37

### DIFF
--- a/app/assets/javascripts/templates/import/import_measure.hbs
+++ b/app/assets/javascripts/templates/import/import_measure.hbs
@@ -25,7 +25,7 @@
           {{/if}}
 
           <div class="form-group">
-            <label for="measureFileInput" class="col-sm-{{titleSize}} control-label">Bundle</label>
+            <label for="measureFileInput" class="col-sm-{{titleSize}} control-label">Measure Data</label>
              <div class="col-sm-{{dataSize}}">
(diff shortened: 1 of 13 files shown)


app/assets/javascripts/templates/import/import_measure.hbs
```diff
@@ -25,91 +25,87 @@
           {{/if}}
 
           <div class="form-group">
-
-            <label for="measureFileInput" class="col-lg-{{titleSize}} control-label">Measure Data:</label>
-             <div class="col-lg-{{dataSize}}">
-               <input class="file" type="file" id="measureFileInput" name="measure_file">
+            <label for="measureFileInput" class="col-sm-{{titleSize}} control-label">Bundle</label>
```
andrequina: "Bundle" is not really meaningful to anyone... Can we change this to Measure File?

thecristen: Sounds good.

PR #321, opened 2014-06-26
Added support for VSAC effectiveDate and includeDraft on .XML upload

jtferns:
### Stories
- https://www.pivotaltracker.com/story/show/72482370
- https://www.pivotaltracker.com/story/show/72478444
### Notes
- updated Gemfile with latest bonnie_bundler and health-data-standards
- added datepicker to measure upload dialog for specifying VSAC <code>effectiveDate</code> when uploading an XML file
- added <code>version</code> to expanded value sets in white/black list builder
- added VSAC <code>includeDraft</code> parameter when loading sources from an XML file, defaulted to <code>true</code> when loading within Bonnie
### Screenshots
- screen shot 2014-06-26 at 10 53 51 am
- screen shot 2014-06-26 at 10 23 41 am


thecristen:
I feel silly for pointing it out but "VSAC effective date" has a : after it whereas the rest of the headings there don't. :)

PR #330, opened 2014-07-09
fixes empty birthdate calculating to true

andrequina:


pkmitre:
So, this is fixing the issue where an empty birthdate calculates as essentially the unix epoch?

PR #331, opened 2014-07-11
Update bonnie_bundler and health-data-standards for XML errors and patie...

jtferns:
...nt characteristics


jtferns:
Closed since #333 is more current.

PR #337, opened 2014-07-15
Add new Bonnie landing page

jtferns:
### Notes
- added <code>home/show.html.erb</code> and <code>landing.less</code> based on <code>landing.html</code> and <code>landing.less</code> from https://github.com/carpeliam/bonnie-yeoman/commit/873576fde18a5400b5fd603334b2b22bd78cf957
- added <code>after_sign_out_path_for</code> to application controller for redirect to landing page after sign out
- added <code>show</code> route to home controller for landing page behavior
- added authenticated and unauthenticated routes for desired landing page / sign in behavior

### DIFF
--- a/app/assets/stylesheets/application.css.less
+++ b/app/assets/stylesheets/application.css.less
@@ -14,6 +14,7 @@
 @import "mixins";
 
 // Styles by page
+@import "landing";
 @import "login";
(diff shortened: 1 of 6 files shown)


thecristen:
I made some small changes but I think it's good

PR #348, opened 2014-07-22
Revise HTML patient export for portfolio users

imjonsnooow:
Change the "export patients" workflow for users with the `portfolio` role. Users with this role will now get patient information including details across measures, rather than simply details for the measure from which the export was triggered.

**Note:** my knowledge in this domain is extremely limited: would a `portfolio` user always want this type of export information? Or would it be pertinent to add some sort of facility for `portfolio` users to specify whether or not they want the patient data within the scope of the current measure?

**Note 2:** I'm still looking into adding test support for this in `test/functional/patients_controller_test.rb`

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -49,7 +49,7 @@ class PatientsController < ApplicationController
         zip.put_next_entry(File.join("qrda","#{index+1}_#{patient.last}_#{patient.first}.xml"))
         zip.puts qrda_exporter.export(patient, measure, start_time, end_time)
         zip.put_next_entry(File.join("html","#{index+1}_#{patient.last}_#{patient.first}.html"))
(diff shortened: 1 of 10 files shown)


imjonsnooow:
I added test support for this change to `test/functional/patients_controller_test.rb`.

I needed to check the HTML files for data from a measure other than the one implicated as an argument to the export action. I added a second sample measure fixture, and added an encounter from this measure to each of the `record` fixtures.

The test uses Nokogiri to search each HTML file exported for `code` information from the new encounter data.

I added a test (`export patients portfolio`) to ensure that the change causes the second measure's information to be exported. I also updated the `export patients` test to ensure that this information is not included for non-portfolio users.

pkmitre:
Merged via de05d9a16f5966822218b583113b4a2819ab29b3

PR #356, opened 2014-07-24
Added initial testplan, including example measures for testing

pkmitre:


jtferns:
I've always taken a test plan and created gists of them in markdown to run them through as I test: [v1.1](https://gist.github.com/jtferns/06eb1aefb272526fc135), [v1.0](https://gist.github.com/jtferns/5ef7f3afc3f929d5e6fd)

I created a simple one [here](https://gist.github.com/jtferns/d1e278d54b07df7cad20) that can be forked whenever someone is running a test.

Would we want to include that in the testplan file?

pkmitre:
That's a great thought, yes.

Also, it looks like there's some good material already in your manual testing steps; we should sync that up as well.

PR #357, opened 2014-07-25
Rework user admin summary to use dedicated model and views

pkmitre:
This reworked approach, using minimalist sub-views and sub-models to handle the updating and display of user summary info, avoids the need to re-render the whole users view, which in turn prevents the aw-snap issue encountered when manipulating multiple user accounts.

### DIFF
--- a/app/assets/javascripts/views/users_view.js.coffee
+++ b/app/assets/javascripts/views/users_view.js.coffee
@@ -6,20 +6,8 @@ class Thorax.Views.Users extends Thorax.Views.BonnieView
     'change .users-sort-list': 'sortUsers'
 
   initialize: ->
-    @totalMeasures = 0
-    @totalPatients = 0
-    @collection.on 'change add reset destroy remove', @updateSummary, this
-
-  updateSummary: ->
-    @totalMeasures = @collection.reduce(((sum, user) -> sum + user.get('measure_count')), 0)
-    @totalPatients = @collection.reduce(((sum, user) -> sum + user.get('patient_count')), 0)
-    @activeUsers = new Thorax.Collection(@collection.filter((u) -> u.get('measure_count')))
-    @activeMeasuresCount = @activeUsers.reduce(((sum, user) -> sum + user.get('measure_count')), 0)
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/views/users_view.js.coffee
```diff
@@ -6,20 +6,8 @@ class Thorax.Views.Users extends Thorax.Views.BonnieView
...
-    @totalPatients = @collection.reduce(((sum, user) -> sum + user.get('patient_count')), 0)
-    @activeUsers = new Thorax.Collection(@collection.filter((u) -> u.get('measure_count')))
-    @activeMeasuresCount = @activeUsers.reduce(((sum, user) -> sum + user.get('measure_count')), 0)
-    @activeMeasuresMax = _.max(@activeUsers.pluck('measure_count'))
-    @activePatientsCount = @activeUsers.reduce(((sum, user) -> sum + user.get('patient_count')), 0)
-    @activePatientsMax = _.max(@activeUsers.pluck('patient_count'))
-    @topTenPatientCounts = _((new Thorax.Collection(@collection.models, comparator: (u) -> parseInt(u.get('patient_count')) * -1 )).sort().pluck('patient_count')).first(10)
-    @render()
+    @userSummaryView = new Thorax.View model: @collection.summary, template: JST['users/user_summary']
+    @tableHeaderView = new Thorax.View model: @collection.summary, template: JST['users/table_header'], tagName: 'thead'
```
jtferns: If Stats are shown and then a user is modified, the Stats view will re-render as hidden and the button class will be inverted.

A quick solution is to keep the Stats window visible if the button style has been toggled via the following in Thorax.Views.Users.initialize()

```
@collection.summary.on 'change', => @$('.stats-panel').toggleClass('hidden') if @$('.btn-toggle-stats').hasClass('btn-primary')
```

pkmitre: Good catch, I took a slightly different approach (moved the button into the appropriate view so that it gets re-rendered correctly)

PR #278, opened 2014-05-20
Replace Expected Values form with static display and popover form

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/63991566
- https://www.pivotaltracker.com/story/show/65923976 (included fix for MSRPOPL as checkbox for patient-based CV measures)
### Notes
- replaced expected values section in patient builder with static display with the following features:
  - for checkboxes, only the checked populations appear, otherwise <code>None</code> is displayed
  - for EoC measures, all populations are displayed with numerical expected values displayed
  - for CV measures, the IPP behaves like checkboxes, the MSRPOPL behaves like EoC populations, and OBSERV has special behavior where updating MSRPOPL re-renders the popover with the changed value of OBSERV fields
- removed empty population title header from patient builder population logic view template
### Screenshots
- checkboxes, not filled
  - checkboxempty
- checkboxes, filled
  - checkboxfilled
- EoC, filled
  - eocfilled
- CV, not filled
  - screen shot 2014-05-20 at 11 04 18 am
- CV, filled
  - cvfilled
- CV, filled & units toggled
  - cvfilledtoggled

PR #360, opened 2014-08-01
Move expected values form into a popover

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/63991566
### Notes
- reworking of https://github.com/projecttacoma/bonnie/pull/278
- created new Thorax view for popover (instead of handlebars script approach)
- displays set expected values, with pencil button for editing (can be set on popover close or by saving the patient)
### Screenshots
- checkboxes
  - screen shot 2014-08-01 at 9 10 33 am
- numbers (EoC)
  - screen shot 2014-08-01 at 9 11 22 am
- patient-based CV
  - screen shot 2014-08-01 at 9 10 03 am
- EoC CV
  - screen shot 2014-08-01 at 9 27 31 am

### DIFF
--- a/app/assets/javascripts/helpers.js.coffee
+++ b/app/assets/javascripts/helpers.js.coffee
@@ -31,3 +31,16 @@ Handlebars.registerHelper 'ifCond', (v1, operator, v2, options) ->
     when '||'
       if (v1 || v2) then options.fn(this) else options.inverse(this)
     else return options.inverse(this)
+
+# A simple math helper for basic arithmetic operations within templates
+# Useful for rendering indexes starting from 1
+Handlebars.registerHelper 'math', (lvalue, operator, rvalue, options) ->
+  lvalue = parseFloat(lvalue)
+  rvalue = parseFloat(rvalue)
+  return {
+    '+' : lvalue + rvalue,
+    '-' : lvalue - rvalue,
+    '*' : lvalue * rvalue,
+    '/' : lvalue / rvalue,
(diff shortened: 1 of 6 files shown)


pkmitre:
When clicking on one of the check boxes, the screen scroll position jumps (possibly related to a re-render of the view?)

jtferns:
Hmm, I'm unable to reproduce this (tested 179v2 [CV], 161v2 [EoC], and 128v2) on a rebased branch; were you testing a specific measure or scenario?

pkmitre:
Closing this pull request unmerged; some analysis from Lizzie that contributed:

Cristen mentioned it’s a more elegant way to handle setting expectations for CV and Episode of Care measures. I guess I don’t really see why from the screen shots.
For more engaging, I actually think it’s easier to accidentally pass over this while you’re editing a patient.
For solving the issue where expectations aren’t being set, that makes sense to me. But what’s the expected behavior when I clone a patient? Do I need to click the box then? If so, this could be annoying if I want to keep the same populations. If not, then it seems like the benefit of this being an added check goes away.

I don’t find the existing check boxes visually distracting, and I think the extra click would be annoying (but like Cristen mentioned, would be a pretty minor inconvenience).

If it does get implemented, then I think the text under “EXPECTED VALUE” needs to change. I read over it several times without noticing it. It might be good to have the populations in a table list (fixed spacing between cells) and to have the populations the patient won’t be in included as well in a separate list.

PR #366, opened 2014-08-13
Basic matrix view for portfolio users

pkmitre:


thecristen:
Very cool!
One thing I noticed is that, when no expectations are set _and_ the patient is in none of the populations, that patient shows as failing in the matrix view. It seems a little inconsistent having those patients show as 'problem' in the matrix view but 'pass' within that measure's view.

pkmitre:
The matrix should show a blank in the case you describe, and that's what I see on my system. When you get a chance, can you show me what it looks like for you?

thecristen:
Actually you’re right.. Ok, must’ve been a mistake on my eyes then.

app/assets/javascripts/views/matrix_view.js.coffee
```diff
@@ -0,0 +1,45 @@
...
+class Thorax.Views.MatrixCell extends Thorax.Views.BonnieView
+
+  tagName: 'td'
+
+  template: -> null # Empty template, everything is in the class name
+
+  events:
+    rendered: ->
+      if @model.differenceFromExpected().has('match')
+        popoverContent = JST['population_calculation_results'](patient: @model.patient.toJSON(), comparisons: @model.differenceFromExpected().get('comparisons'))
```
jtferns, line 27: The styles for the population calculation results are embedded within <code>measure_calculation.less</code>.

If we move the following out (to somewhere like <code>mixins.less</code>; or embed the same styles within <code>matrix.less</code>) we can get styles on the popovers, too!

```
.pass {
  .text-success;
}
.fail {
  .text-danger;
}
.default {
  color: @brand-primary;
}
```

screen shot 2014-08-15 at 10 14 54 am

app/assets/javascripts/views/matrix_view.js.coffee
```diff
@@ -0,0 +1,45 @@
...
+        @$el.popover trigger: 'hover', placement: 'bottom', container: 'body', title: @model.population.displayName(), html: true, content: popoverContent
```
jtferns: Optionally, having the patient name in the popover title might be helpful when looking at results on the far right and the patients column is not within view.

screen shot 2014-08-15 at 10 24 26 am

pkmitre:
Good suggestions, I'll get them both in.

PR #370, opened 2014-08-15
Plain user bundle export

cgaray:
### Story

https://www.pivotaltracker.com/story/show/76397158
### Notes
- Created new new non-admin user controller
- Added button to measures view / dashboard for bundle export (I'm sure there is a better location for this)
- Added test for user bundle export
### Screenshots

Current export bundle button
export bundle button


andrequina:
Lets put the download button as only the icon (drop the text), and have it to the right of the "Measures" heading.  Right next to the measures heading to the right rather than next to the upload button.

cgaray:
### Story
- [Users can export their own bundles](https://www.pivotaltracker.com/story/show/76397158)
- [Disable downloading a bundle for a user that has no data](https://www.pivotaltracker.com/story/show/76357856)

### Notes
- Removed text from export button and moved closer to row title
- Check for measures and patients length and disable bundle export button if collections are empty

### Screenshots

untitled picture

app/assets/stylesheets/measures.less
```diff
@@ -56,6 +56,19 @@
...
+       display: inline-block;
+   }
+   .btn-wrapper-bundle {
+    border-left: 5px white solid;
+    border-right: 5px white solid;
+    display: inline-block;
+    line-height: 1;
+    position: absolute;
+    top: 0;
+   }
```
thecristen, line 67: I noticed you're using bootstrap's `.btn-wrapper` code with some adjustments. With LESS you can make use of bootstrap's code in the class:

```
.btn-wrapper-bundle {
   .btn-wrapper;
    display: inline-block;
    right: inherit; // overrides the value from .btn-wrapper
}
```

cgaray, line 67: Thanks, that worked great. I'm still finding my way around CSS and LESS.

pkmitre:
Merged via 10878c804519c1fa6cd0663f4fbcef77d2b32d23

PR #373, opened 2014-08-15
fix negation bubbling for specific occurrences for AND NOT only paths

andrequina:


app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -43,10 +43,10 @@ class Thorax.Views.PopulationLogic extends Thorax.Views.BonnieView
       updatedRationale = result.specificsRationale()
       for code in Thorax.Models.Measure.allPopulationCodes
         if rationale[code]?
-          @showRationaleForPopulation(code, rationale, updatedRationale)
-          # FIXME: Update patient population panel heading styling based on calculation results until rationale is fixed to bubble-up negations!
-          @$(".#{code}_children .#{code}").closest('.panel-heading').removeClass('eval-panel-true eval-panel-false').addClass("eval-panel-#{!!result.get(code)}")
-          @$(".#{code}").removeClass('eval-true eval-false').addClass("eval-#{!!result.get(code)}")
+          # if we are highlighting exceptions, have a numerator but don't qualify for the exceptions, then do not hightlight
+          # even if we are in the numerator we have to highlight if we are in the exceptions because we could be in both for EoC
```
jtferns, line 47: I'm not quite sure I understand what's going on with these two comment lines, the first mentions not highlighting, while the second mentions highlighting.

app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -43,10 +43,10 @@ class Thorax.Views.PopulationLogic extends Thorax.Views.BonnieView
...
+          if !(code == 'DENEXCEP' && result.get('NUMER') && !result.get(code))
```
jtferns, line 48: Is this equivalent to `code != 'DENEXCEP'` OR `!result.get('NUMER')` OR `result.get(code)` (by simplifying the negation)?

Is the intent to avoid showing the rationale if the `code` is DENEXCEP and there is a `result` value for NUMER and no `result` value for DENEXCEP?

PR #374, opened 2014-08-18
Update measure chords view button w/ text view toggle

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/77034560
### Notes
- added two buttons to measure settings template for chords and text views
- toggle the appropriate button via <code>.btn-viz-chords</code> and <code>.btn-viz-text</code>
- reset the logic view if measure settings is closed
### Screenshots
- chords
  - screen shot 2014-08-18 at 9 39 36 am
- text
  - screen shot 2014-08-18 at 9 39 41 am

### DIFF
--- a/app/assets/javascripts/templates/measure.hbs
+++ b/app/assets/javascripts/templates/measure.hbs
@@ -4,7 +4,8 @@
     <div class="full-title">{{title}}</div>
     <span class="settings-container">
       <div class="measure-settings">
-        {{#button "toggleVisualization" class="btn btn-default btn-measure-viz"}}<i class="fa fa-sitemap"></i> Chords View{{/button}}
+        {{#button "toggleVisualization" class="btn btn-default btn-measure-viz btn-viz-chords"}}<i class="fa fa-sitemap" aria-hidden="true"></i> Chords View{{/button}}
+        {{#button "toggleVisualization" class="btn btn-default btn-measure-viz btn-viz-text"}}<i class="fa fa-list" aria-hidden="true"></i> Text View{{/button}}
         {{#button "updateMeasure" class="btn btn-default update-measure"}}<i class="fa fa-upload" aria-hidden="true"></i> Update{{/button}}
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/templates/measure.hbs
```diff
@@ -4,7 +4,8 @@
     <div class="full-title">{{title}}</div>
     <span class="settings-container">
       <div class="measure-settings">
-        {{#button "toggleVisualization" class="btn btn-default btn-measure-viz"}}<i class="fa fa-sitemap"></i> Chords View{{/button}}
+        {{#button "toggleVisualization" class="btn btn-default btn-measure-viz btn-viz-chords"}}<i class="fa fa-sitemap" aria-hidden="true"></i> Chords View{{/button}}
+        {{#button "toggleVisualization" class="btn btn-default btn-measure-viz btn-viz-text"}}<i class="fa fa-list" aria-hidden="true"></i> Text View{{/button}}
```
pkmitre, line 8: Why not just change the button text? You can look at the "save" button in patient builder for an example. If I remember correctly, you set alternate texts by specifying button attributes, ie

```
data-foo-text="Foo"
```

and then in JS you change the button text by calling

```
@$('button').button('foo')
```

pkmitre, line 8: But you're doing something else using classes; can the button type class be toggled on click too? If you prefer the current implementation that's fine, just offering another option...

jtferns, line 8: My initial approach utilized jQuery.html(), since I wanted to use an icon when converting 'Chords View' to 'Text View'. I figured calling .html() within the view was not a great idea, since I had to include the `<i></i>` class within the view and we would have elements defined in the template crossing into the view. 

I use the specific class to determine if the visualization is toggled, but I can switch it if it's possible to include an icon and/or detect the changed text for appropriate toggling.

pkmitre, line 8: Makes sense, lets stick with your approach. Will merge as soon as tests pass again after my rebase on master.

PR #383, opened 2014-08-20
display logging information on debug page

ssayer:
needing for cypress crosswalk debugging


ssayer:
looks like this outputs to the page log even if logging is set to false

(this doesn't matter to me, but may affect other users)

PR #386, opened 2014-08-25
Updated hqmf2js to master as of August 25 in Gemfile.lock

okeefm:
This pulls ssayer's fix in hqmf2js into Bonnie.


pkmitre:
HQMF2JS was updated elsewhere

PR #389, opened 2014-08-27
Patient builder scrolling

thecristen:
(replacing pull request #369 with this one for cleaner diffs and a less out-of-date base)
## Stories
- https://www.pivotaltracker.com/story/show/62488424
- https://www.pivotaltracker.com/story/show/77141332 (replace scrolling in logic with pagination)
## Notes
- Uses the bootstrap affix plugin to pin the selected sections to the top when you scroll a certain distance down the page
- Introduces pagination for the logic view, while the other two sections scroll freely
- Includes an update of bootstrap
- Includes an update of jquery UI
- Minor adjustment to the dragged element to keep it working and looking proper
- Added some visual feedback - when you start to drag, the droppable area shows as shaded
## Screenshot

In IE9
screen shot 2014-08-27 at 10 30 05


pkmitre:
I've used this to build a few patients, and it works well, with one issue that sometimes makes things difficult: scrolling the measure logic can only be done in full pages, so if I'm looking at something like 

```
Occurrence A: Encounter, Performed:
Emergency Department Visit <= 1 hour
ends before start of Occurrence A:
Encounter, Performed: Non-Elective
Inpatient Encounter
```

and it's cut off at the bottom of the page and I only see half of it, I have to jump back and forth and look at the top, then the bottom, and that makes it hard to follow the logic. I'd like to be able to scroll more granularly, like in a regular scrolling section, and ideally with my scroll wheel or two finger scrolling.

thecristen:
Ok @pkmitre, feel free to try this updated version.

pkmitre:
One more quick question: has this been tested in different browsers? particularly IE?

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -177,6 +179,53 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+      @$('.logic-pager').show()
+      if $logic.scrollTop() <= buffer
+        @$('.logic-pager.up').hide()
+      else if $logic.scrollTop() >= $logic.prop('scrollHeight') - $logic.height() - buffer
+        @$('.logic-pager.down').hide()     
+
+    moveLogic = (dir) ->
+      page = $logic.height() - $logic.css('line-height').replace('px', '') # scroll down 1 line less than whole screen length
+      if dir is 'up' then $logic.animate scrollTop: $logic.scrollTop() - page, -> logicPagingUpdate()
+      if dir is 'down' then $logic.animate scrollTop: $logic.scrollTop() + page, -> logicPagingUpdate()
```
pkmitre, line 227: Since these utility functions are only used in one place, they're probably not necessary.

app/assets/javascripts/router.js.coffee
```diff
@@ -54,6 +54,7 @@ class BonnieRouter extends Backbone.Router
     document.title += " - #{measure.get('cms_id')}" if measure?
     patientBuilderView = new Thorax.Views.PatientBuilder(model: patient, measure: measure, patients: @patients, measures: @measures)
     @mainView.setView patientBuilderView
+    patientBuilderView.handleAffix()
```
pkmitre, line 57: This would probably be more appropriate in a render event handler in the patient builder itself. ie

  events:
    rendered: -> @handleAffix()

app/assets/stylesheets/builder.less
```diff
@@ -156,26 +156,41 @@
...
+          .pull-right;
+        }
+        &:hover {
+          text-decoration: none;
+          background-color: @blue-darker;
+          &.closed {
+            & .element-title {
+              display: initial;
+            }
+          }
```
pkmitre, line 192: This section changes the behavior of the elements section so that non-selected element types have no visible title

1) This shouldn't be lumped into the same pull request as another feature

2) I think this change makes the patient builder much harder to use

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -177,6 +179,53 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
     @model.set 'expired', false
     @$('#expired').focus()
 
+  handleAffix: ->
+    # ensure patient history is always long enough to not cause weird behavior
+    @$('.criteria-container').css("min-height",$(window).height())
+    $(window).on 'resize', -> @$('.criteria-container').css("min-height",$(window).height())
```
pkmitre, line 185: References to $(window).height() in view code feels like a break of encapsulation. I don't have a better approach to suggest, but just want to note the concern in case we can think of another approach that is straight forward and does't have the same issue.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -177,6 +179,53 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+              width: $(@).find('.scrolling').outerWidth() 
+          $(@).css width: $(@).width() #assign current width explicitly to affixed element  
+        $('.logic-pager.up').hide()
+      .on 'affixed-top.bs.affix', ->
+        $('.logic-pager').hide() # hide the pagination part, removed disabled buttons
+        $(@).each -> 
+          $(@).removeAttr('style') #revert each affixed element to default css styling
+          $(@).find('.scrolling').removeAttr('style').animate scrollTop: 0 #scroll div back to top, remove custom styling  
+    $cols.affix offset: { top: @$('.criteria-container').parent().offset().top } # tell affix to activate after scrolling this many pixels
+    @logicMoving() #handle scrolling and paging the logic pane
```
pkmitre, line 208: After some discussion I understand that this code sets the sizes of the affixed sections appropriately; a comment with a quick explanation would help make the code more readable.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -177,6 +179,53 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
     @$('#expired').focus()
 
+  handleAffix: ->
+    # ensure patient history is always long enough to not cause weird behavior
+    @$('.criteria-container').css("min-height",$(window).height())
+    $(window).on 'resize', -> @$('.criteria-container').css("min-height",$(window).height())
+    # affix side columns to get desired behavior
+    $cols = @$('#criteriaElements, #populationLogic, #history') #these get affixed. add listeners
+      .on 'affix.bs.affix', ->
+        $('.logic-pager').show()
```
pkmitre, line 189: We'd ideally never use $ in a view, sticking to @$ for encapsulation

thecristen:
Replacing this with an updated PR

PR #390, opened 2014-08-27
Expectation naming

thecristen:
- https://www.pivotaltracker.com/story/show/75148928

It was just really bugging me to have different shorthands for the different expectations as I went back and forth between the patient builder view and the measure view. From what I understand, the rationale for the shorter names in the patient builder was because they sometimes overflow from one line of text. From what I've seen, they still overflow depending on the measure. So to me it makes sense to stop using the shorter names in this particular place.


pkmitre:
Merged via e77c100bde068202e3f49afae174f0090c5b7480

PR #395, opened 2014-09-09
Edit population title

thecristen:
## Story
- https://www.pivotaltracker.com/story/show/75134126
## Notes
- Needs more thorax goodness, less DOM hacking
## Screenshot

image

### DIFF
--- a/app/assets/javascripts/views/measures_view.js.coffee
+++ b/app/assets/javascripts/views/measures_view.js.coffee
@@ -46,18 +46,14 @@ class Thorax.Views.MeasureRowView extends Thorax.Views.BonnieView
   editPopulationName: (e) ->
     $(e.currentTarget).find('i').toggleClass 'fa-pencil fa-save'
     title = $(e.currentTarget).prev() 
-
-    if title.is('span') 
+    if title.is('span')
       title.replaceWith -> return '<input id="population_edit" name="population_edit" type="text" value="'+$(@).html()+'">' #make it editable
-    else 
-      newTitle = title.val()
-      title.replaceWith -> 
-        return '<span>'+newTitle+'</span>' #make it uneditable
-
-      pop =  $(e.currentTarget).closest('.row.population') #this population
-      y = pop.siblings().addBack().index(pop) #tells us where this pop is in its group of pops
-
-      @model.save(name: newTitle, index: y) #sets it in model
+    else
+      newTitle = $('#population_edit').val() # grab title
+      thisPop = title.closest('.row.population') # this population
+      i = thisPop.siblings().addBack().index(thisPop) # find index
+      title.replaceWith('<span>'+newTitle+'</span>') # reset display
+      @model.save({name: newTitle, index: i}, {silent: true}) # sets it in model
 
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/views/measures_view.js.coffee
```diff
@@ -34,12 +37,23 @@ class Thorax.Views.MeasureRowView extends Thorax.Views.BonnieView
...
+ 
   updateMeasure: (e) ->
     importMeasureView = new Thorax.Views.ImportMeasure(model: @model)
     importMeasureView.appendTo(@$el)
     importMeasureView.display()
 
+  editPopulationName: (e) ->
+    $(e.currentTarget).find('i').toggleClass 'fa-pencil fa-save'
+    title = $(e.currentTarget).prev() 
+    if title.is('span')
```
pkmitre, line 49: Rather than checking the DOM to see what action is being performed (edit or save), use two different methods (ie editPopulationName and savePopulationName) triggered by two different events.

app/assets/javascripts/views/measures_view.js.coffee
```diff
@@ -34,12 +37,23 @@ class Thorax.Views.MeasureRowView extends Thorax.Views.BonnieView
...
+      title.replaceWith -> return '<input id="population_edit" name="population_edit" type="text" value="'+$(@).html()+'">' #make it editable
```
pkmitre, line 50: We should avoid inline HTML in our JS code, all HTML should be in our templates

app/assets/javascripts/views/measures_view.js.coffee
```diff
@@ -34,12 +37,23 @@ class Thorax.Views.MeasureRowView extends Thorax.Views.BonnieView
...
+    else
+      newTitle = $('#population_edit').val() # grab title
```
pkmitre, line 52: The most "thoraxy" approach would be to create a micro view for the population, where the view model is the population; then we could use @serialize() to get the data and put it on the population object, refer to the views model to get the population index, etc. rather than going directly to the DOM. That's a bit of a refactoring, but still fairly straightforward, and something we should consider.

app/assets/javascripts/views/measures_view.js.coffee
```diff
@@ -34,12 +37,23 @@ class Thorax.Views.MeasureRowView extends Thorax.Views.BonnieView
...
+      thisPop = title.closest('.row.population') # this population
+      i = thisPop.siblings().addBack().index(thisPop) # find index
+      title.replaceWith('<span>'+newTitle+'</span>') # reset display
+      @model.save({name: newTitle, index: i}, {silent: true}) # sets it in model
```
pkmitre, line 56: This is confusing because we're essentially updating the measure model, giving it a name and index, and only incidentally saving it to the population via some backend manipulation.

config/routes.rb
```diff
@@ -15,6 +15,8 @@
 
   root to: 'home#index'
 
+  put 'measures/:id' => 'populations#update_population_titles'
```
pkmitre, line 18: From a backend API perspective this is a bit confusing, we're creating an endpoint where we write to a measure but it's saving to the measure's population.

PR #396, opened 2014-09-10
Prevent false-highlighting-bubbling for specific occurrences within Unions

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/78230864
### Notes
- only consider unique occurrences when building specificsRationale
- extend `orCounts` to consider grouping data criteria representing Unions
### Screenshots
- undesired false-bubbling
  - screen shot 2014-09-10 at 9 03 39 am
- fixed false-bubbling
  - screen shot 2014-09-10 at 9 03 57 am

### DIFF
--- a/app/assets/javascripts/models/result.js.coffee
+++ b/app/assets/javascripts/models/result.js.coffee
@@ -158,11 +158,9 @@ class Thorax.Models.Result extends Thorax.Model
   # walk through data criteria to account for specific occurrences within a UNION
   calculateDataCriteriaOrCounts: (rationale) ->
     orCounts = {}
-    for key, dc of @measure.get('data_criteria') when dc.derivation_operator == 'UNION' && dc.key.indexOf('UNION') != -1
-      trueCount = 0
+    for key, dc of @measure.get('data_criteria') when dc.derivation_operator == 'UNION' && key.indexOf('UNION') != -1
       for child in dc.children_criteria
-        trueCount++ if rationale[dc.key]
-      if orCounts[dc.key]? then orCounts[dc.key] += trueCount else orCounts[dc.key] = trueCount
+        orCounts[key] = (orCounts[key] || 0) + 1 if rationale[key]
     orCounts
 
   codedEntriesForDataCriteria: (dataCriteriaKey) ->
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/models/result.js.coffee
```diff
@@ -156,6 +155,16 @@ class Thorax.Models.Result extends Thorax.Model
       _.extend(orCounts, @calculateOrCountsRecursive(rationale, precondition.preconditions))
     return orCounts
 
+  # walk through data criteria to account for specific occurrences within a UNION
+  calculateDataCriteriaOrCounts: (rationale) ->
+    orCounts = {}
+    for key, dc of @measure.get('data_criteria') when dc.derivation_operator == 'UNION' && dc.key.indexOf('UNION') != -1
```
pkmitre: Why do we need to check the key as well as the derivation operator?

Is dc.key ever going to be different from just key?

jtferns: I had run into issues with a jasmine [spec](https://github.com/projecttacoma/bonnie/blob/master/spec/javascripts/models/result.spec.js.coffee#L13) when I was only checking for the `derivation_operator` (other groupings, typically with temporal references, also have a UNION derivation_operator, so I opted to exclude them from this change).

I think in this context `key` will be the same value as `dc.key`; would it be cleaner to just replace all usage of `dc.key` with key?

app/assets/javascripts/models/result.js.coffee
```diff
@@ -156,6 +155,16 @@ class Thorax.Models.Result extends Thorax.Model
...
+      trueCount = 0
+      for child in dc.children_criteria
+        trueCount++ if rationale[dc.key]
+      if orCounts[dc.key]? then orCounts[dc.key] += trueCount else orCounts[dc.key] = trueCount
+    orCounts
+
   codedEntriesForDataCriteria: (dataCriteriaKey) ->
     @get('rationale')[dataCriteriaKey]?['results']
```
pkmitre, line 168: Code golf... is this any simpler?

```
calculateDataCriteriaOrCounts: (rationale) ->
  orCounts = {}
  for key, dc of @measure.get('data_criteria') when dc.derivation_operator == 'UNION' && key.indexOf('UNION') != -1
    for child in dc.children_criteria
      orCounts[key] = (orCounts[key] || 0) + 1 if rationale[key]
  orCounts
```

jtferns, line 168: I think that's much simpler, I will update with your suggestion.

PR #399, opened 2014-09-17
Edit population titles

thecristen:
Replacing earlier PR #395 
## Story
- https://www.pivotaltracker.com/story/show/75134126
## Notes

Took the inline editing example from the edit user functionality in the admin page, seems to work pretty well here.
## Screenshot

screenshot 2014-09-17 09 41 16

### DIFF
--- a/app/assets/javascripts/templates/measure/population.hbs
+++ b/app/assets/javascripts/templates/measure/population.hbs
@@ -1,0 +1,4 @@
+<span class="sr-only">{{../cms_id}} </span>
+<strong>Population:</strong>
+{{title}}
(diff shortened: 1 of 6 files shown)


pkmitre:
Looks great! Do you think that allowing the title to be saved using the enter key is worth while?

thecristen:
Yeah, I think that would be worthwhile. Good catch!

pkmitre:
Just noticed one more thing: when a browser is in a bit wider configuration, the text box takes up a relatively small amount of the available space:
screen shot 2014-10-08 at 10 05 16 am
It would be nice if the text box took as much space as available.

thecristen:
Here's the latest! 
image

PR #408, opened 2014-10-06
Update patient model parse handling of race and ethnicity

jtferns:
### Story
- https://www.pivotaltracker.com/story/show/79806350
### Notes
- `race` and `ethnicity` are extracted from the backend hash/object representation, which happens twice when loading the `deepClone()`'d model in the patient builder
- Note: the desired race and ethnicity were always being saved correctly, but not rendered when editing the patient within the patient builder

### DIFF
--- a/app/assets/javascripts/models/patient.js.coffee
+++ b/app/assets/javascripts/models/patient.js.coffee
@@ -11,8 +11,8 @@ class Thorax.Models.Patient extends Thorax.Model
     # This section is a bit unusual: we map from server side values to a more straight forward client
     # side representation; the reverse mapping would usually happen in toJSON(), but in this case it
     # happens on the server in the controller
-    attrs.ethnicity = attrs.ethnicity?.code
-    attrs.race = attrs.race?.code
+    attrs.ethnicity = attrs.ethnicity?.code if typeof attrs.ethnicity == 'object'
+    attrs.race = attrs.race?.code if typeof attrs.race == 'object'
     attrs.payer = attrs.insurance_providers?[0]?.type || 'OT'
 
     attrs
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/models/patient.js.coffee
```diff
@@ -11,8 +11,8 @@ class Thorax.Models.Patient extends Thorax.Model
     # This section is a bit unusual: we map from server side values to a more straight forward client
     # side representation; the reverse mapping would usually happen in toJSON(), but in this case it
     # happens on the server in the controller
-    attrs.ethnicity = attrs.ethnicity?.code
-    attrs.race = attrs.race?.code
+    attrs.ethnicity = attrs.ethnicity?.code if typeof attrs.ethnicity == 'object'
+    attrs.race = attrs.race?.code if typeof attrs.race == 'object'
```
pkmitre: This code might be a bit confusing outside the context of the pull request; a comment would help, but maybe a different change would be useful:

```
attrs.ethnicity_code = attrs.ethnicity?.code
attrs.race_code = attrs.race?.code
```

This would require changes in other places too (at least the controller and the form, probably calculation code too), so might not be worth it...

jtferns: I think we already have some decent workarounds in place (mapping from backend to client-side and vice versa), so I am hesitant to update the attribute entirely.

How about this as an alternative:

```
# extract demographics from hash, or use extracted values when cloning
attrs.ethnicity = attrs.ethnicity?.code || attrs.ethnicity
attrs.race = attrs.race?.code || attrs.race
```

pkmitre: Seems like a good approach!

PR #413, opened 2014-10-31
Display measure complexity on measure view page

pkmitre:
## Story

https://www.pivotaltracker.com/story/show/80044824

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -30,7 +30,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/bonnie_bundler.git
-  revision: 4351c648f0afefa6bdafc71f407efabeba840cb9
+  revision: 6d6b2c60a8e2164297cc98462af3eebc6320a477
   branch: master
   specs:
     bonnie_bundler (1.0.0)
--- a/app/assets/javascripts/helpers.js.coffee
+++ b/app/assets/javascripts/helpers.js.coffee
@@ -5,6 +5,15 @@ Handlebars.registerHelper 'debug', -> debugger
 # add a helper for formatting dates
 Handlebars.registerHelper 'moment', (date, format) -> moment.utc(date).format(format)
 
+# helper for displaying complexity score in graphical form
+Handlebars.registerHelper 'complexityIcon', (score) ->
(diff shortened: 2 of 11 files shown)


jtferns:
With my local instance pointed to this branch, I ran into several issues within the measure view where I was looking at a measure that did not have a complexity model on the measure (e.g., the measures that are loaded from a local db reset).

I think it would be useful to handle missing complexity values gracefully, given that a user will have to reload their measures to obtain meaningful complexity results.

The following restored functionality to the measure view, but provided what I think is an incorrect complexity result (see screenshot below).

```
diff --git a/app/assets/javascripts/views/measure_complexity.js.coffee b/app/assets/javascripts/views/measure_complexity.js.coffee
index 854f55f..4a3a7e6 100644
--- a/app/assets/javascripts/views/measure_complexity.js.coffee
+++ b/app/assets/javascripts/views/measure_complexity.js.coffee
@@ -17,15 +17,15 @@ class Thorax.Views.MeasureComplexity extends Thorax.Views.BonnieView
   context: ->
     complexity = @model.get('complexity')
     # Calculate overall score; this is just the worst population or variable, over the entire measure
-    scores = _(complexity.populations).values()
-    scores = scores.concat _(complexity.variables).values()
+    scores = _(complexity?.populations).values()
+    scores = scores.concat _(complexity?.variables).values()
     overall = _(scores).max()
     # Figure out which populations are currently displayed and grab that complexity data
     complexities = []
       if population = @population.get(code)
-        complexities.push label: code, value: complexity.populations[population.code]
+        complexities.push label: code, value: complexity?.populations?[population.code]
     # Add in the variable complexity data, which is always displayed
-    for name, value of complexity.variables
+    for name, value of complexity?.variables
       complexities.push label: "$#{name}", value: value
     _(super).extend complexities: complexities, overall: overall
```

screen shot 2014-11-03 at 2 10 33 pm

pkmitre:
I've added a rake task that re-saves all measures, which we'll run when deploying this code.

I also updated the complexity format to use an associative array rather than a hash so that we can support variable names containing '$' or '.', which mongo does not allow in keys.

PR #414, opened 2014-11-12
adjust interaction with data criteria elements

thecristen:
de-bundling this from my patient builder scrolling work.

The main change is that, when opening a category, all others compress to save space gracefully. Long text is handled by cutting it off with an ellipsis.

image


thecristen:
Replacing this with a new PR

PR #422, opened 2015-01-09
Measures_yml argument in the bundle_user task defined incorrectly 

okeefm:
The `measures_yml` argument in the bundle_user task was being defined with a question mark, which was being interpreted as part of the argument name. Therefore, the passed-in `measures_yml` was never being used; instead, only the `measures_2_5_0.yml` file specified as a default in the task was being read.


pkmitre:
Merged via 02bf81886842b422f46161cae7f332ac0dcb799d

PR #431, opened 2015-01-28
Navigation improvements and UI tweaks

thecristen:
## Stories

None! As my last hurrah before the 1.3 release I wanted to replace the 3 indicator circles with something a little more dynamic and useful that acknowledges the increased complexity of Bonnie. I did like having the measure period visible, so I kept that. Now if you go through a measure or a patient or the patient bank you will see a breadcrumb that shows where you are in that process. It's styled to be unobtrusive, and contains links to the expected places to be more useful. It doesn't take significantly more vertical space than was used before. 
And if you're on a different page (complexity, admin, account) the breadcrumb doesn't show at all.

I also cleaned up some markup in a few places, streamlined some headings, adjusted some widths, added some icons, and overall just tried to smooth over some rough edges. By happenstance [this story](https://www.pivotaltracker.com/story/show/70891450) is rendered obsolete.
### Dashboard

image
### Measure

image
### New patient

image
### Existing patient

image
### Patient bank

image
### Random other page!

image

### DIFF
--- a/app/assets/javascripts/router.js.coffee
+++ b/app/assets/javascripts/router.js.coffee
@@ -12,7 +12,7 @@ class BonnieRouter extends Backbone.Router
     # FIXME deprecated, use measure.get('patients') to get patients for individual measure
     @patients = new Thorax.Collections.Patients()
 
-    @breadcrumb = new Thorax.Views.Breadcrumb(el: '.breadcrumb')
+    @breadcrumb = new Thorax.Views.Breadcrumb()
 
     @on 'route', -> window.scrollTo(0, 0)
 
@@ -43,13 +43,12 @@ class BonnieRouter extends Backbone.Router
     @mainView.setView(complexityView)
 
   renderMeasure: (hqmfSetId) ->
-    @navigationSetup "Measure View", "measure"
+    @navigationSetup "Measure View", "dashboard"
     measure = @measures.findWhere({hqmf_set_id: hqmfSetId})
     document.title += " - #{measure.get('cms_id')}" if measure?
     measureView = new Thorax.Views.Measure(model: measure, patients: @patients)
     @mainView.setView(measureView)
     @breadcrumb.addMeasure(measure)
-    $('.navbar-nav > li').removeClass('active').filter('.nav-dashboard').addClass('active')
 
   renderUsers: ->
     @navigationSetup "Admin", "admin"
--- a/app/assets/javascripts/router.js.coffee
+++ b/app/assets/javascripts/router.js.coffee
@@ -33,29 +33,35 @@ class BonnieRouter extends Backbone.Router
     else
       dashboardView = new Thorax.Views.Measures(collection: @measures.sort(), patients: @patients)
     @mainView.setView(dashboardView)
+    @breadcrumbSetup "dashboard"
 
   renderComplexity: (measureSet1, measureSet2) ->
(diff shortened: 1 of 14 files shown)


jtferns:
Could `Logout` use an [icon](http://fortawesome.github.io/Font-Awesome/icon/sign-out/), too?

thecristen:
@jtferns Yeah, I was thinking that very same icon! Because of the way that particular link is generated in the rails template, I couldn't quite grok how to add it inside the link without significant changes.. if you'd like to add it in go ahead. :)

noranda:
I would suggest grouping together some of the navigation links that are rarely used (i.e. Admin, Account, Help). That would give the navigation more space. Otherwise I think it looks great!

jtferns:
```
<li><%= link_to '<i class="fa fa-sign-out"></i> Logout'.html_safe, destroy_user_session_path, :method => :delete %></li>
```

works as a replacement:
screen shot 2015-01-28 at 2 47 51 pm

I derived it from [this post](http://stackoverflow.com/a/9401988).

pkmitre:
Can the breadcrumb code be refactored to use a Thorax view and template instead of inline HTML and jQuery? One possibility:

1) Create a new Thorax.Views.Breadcrumbs class with related template
2) Initialize it in the router.js initialize method, ie

```
@breadcrumbs = new Thorax.Views.Breadcrumbs
```

3) Instead of calling your breadcrumbSetup() method, call @breadcrumbs.setup()

thecristen:
@pkmitre something like 28e3618?

app/assets/javascripts/views/breadcrumb.js.coffee
```diff
@@ -0,0 +1,35 @@
+class Thorax.Views.Breadcrumb extends Thorax.Views.BonnieView
+
+  template: JST['breadcrumb']
+
+  initialize: ->
+    @setModel new Thorax.Model
+    @model.on 'change', => @render()
```
pkmitre: An explicit render on model change shouldn't be necessary in Thorax, that's built in, so if it doesn't work without this line there's some other issue.

app/assets/javascripts/router.js.coffee
```diff
@@ -70,16 +76,15 @@ class BonnieRouter extends Backbone.Router
...
   navigationSetup: (title, selectedNav) ->
     @calculator.cancelCalculations()
+    @breadcrumb.clear()
     document.title = "Bonnie v#{bonnie.applicationVersion}: #{title}"
     if selectedNav?
-      $('ul.nav > li, .indicator-circle').removeClass('active').filter(".nav-#{selectedNav}, .indicator-#{selectedNav}").addClass('active')
-    switch selectedNav
-      when 'complexity', 'admin' then $('.nav-context').addClass('hidden')
-      else $('.nav-context').removeClass('hidden')
+      $('ul.nav > li').removeClass('active').filter(".nav-#{selectedNav}").addClass('active')
```
pkmitre, line 86: The breadcrumb now appears on the complexity dashboard, where we don't want it, and the spacing there seems off.

pkmitre:
After playing with it a bit more, some thoughts:

1) I think the extra vertical space this design takes up, even though it's not much, makes the pages a bit harder to use

2) I'm not sure about the change from "test patient" to "patient"; I think we need to make it clear that this is not expected to be real patient data

PR #433, opened 2015-01-28
Show shortened measure description by default

thecristen:
- Set to show the first three lines of text.

| Short | Not so short |
| :-: | :-: |
| image | image |


pkmitre:
1) How about "show less" instead of "hide", picky I know, but the link doesn't actually hide the content...

2) The mouse cursor shows "text" instead of "pointer" when hovering over the links, which makes it harder to tell that this is clickable

PR #445, opened 2015-04-03
Add measure logic javascript to HQMF vs SimpleXML comparison

pkmitre:


lib/tasks/parser_comparison.rake
```diff
@@ -16,6 +16,11 @@ namespace :bonnie do
       app_javascript.gsub!('Δ', 'bigDeltaForD3')
       value_sets ||= measure.value_sets
       value_sets_by_oid = value_sets.index_by(&:oid)
+      measure_javascript = ''
+      measure.populations.each_with_index do |population, idx|
+        measure_javascript += "/////////////////////// JAVASCRIPT FOR POPULATION #{idx} ///////////////////////"
+        measure_javascript += measure.as_javascript(idx)
```
jtferns: This might be kind of lame, but for measures that aren't parsing correctly, can we capture any JS generation errors?:

```
measure_javascript += begin
          measure.as_javascript(idx)
        rescue => e
          "Failed to generated measure JS: #{e.message}"
        end
```

pkmitre: Good idea

PR #450, opened 2015-04-25
update bootstrap version to v3.3.4

thecristen:
and be more selective about importing bootstrap's LESS files into our own CSS
## Story

https://www.pivotaltracker.com/story/show/91736682
## Notes

Tooltips are [no longer wedged left](https://github.com/twbs/bootstrap/pull/14015).
Instead of importing the entire bootstrap.less, I changed application.css.less to import only the components utilized by Bonnie. Bootstrap components excluded:
- glyphicons.less
- code.less
- pagination.less
- pager.less
- labels.less
- badges.less
- thumbnails.less
- media.less
- responsive-embed.less
- carousel.less
  This keeps the application.css file small enough to not run into problems in IE9.


jtferns:
The value sets White/Black List builder utilizes the following, so we probably want to keep them around:
- labels.less
- badges.less

thecristen:
@jtferns sounds good!

jtferns:
This looks good throughout the app (nothing is broken or missing, white/black list has badges and labels now).

PR #451, opened 2015-04-27
sort patient results by status and name

thecristen:
https://www.pivotaltracker.com/story/show/91736798

Should list results in the patient bank first by presence of expectations (all patients made for the measure, then all patients not made for the measure), then by match (passing patients then failing patients), then by last name and first name.


jtferns:
The first sorting option checks for the existence of expectations, which leads to the following sorting outcome if there's a patient that is passing but has no expectations:

screen shot 2015-04-27 at 4 15 45 pm

I think patients with expected values are more meaningful, but I wanted to ensure it's the intended behavior before merging.

thecristen:
Adding a `@differences.sort()` seemed to improve things on my end.

pkmitre:
Merged via e2b715a9fbb7ec3fdf85db66f56b98241ab583fe

PR #452, opened 2015-04-27
User sees an indication of those patients already cloned when looking at list of patients for cloning

thecristen:
## Story

https://www.pivotaltracker.com/story/show/81520184
## Notes
- When a patient gets cloned from the patient bank, that patient gets an origin_data attribute containing some data about the patient it was cloned from. Now it actually persists to the back end!
- The template simply swaps in a different icon if an origin_data attribute is present
- Had to update FontAwesome to get the favored icon.


app/controllers/patients_controller.rb
```diff
@@ -123,6 +123,8 @@ def update_patient(patient)
 
     patient.expected_values = params['expected_values']
 
+    patient['origin_data'] = params['origin_data']
```
jtferns: If we create Patient A from Measure A, share it, and then clone it into Measure B from the Patient Bank, we will have a clone (Patient B) that has `origin_data` set to [A].

If we then share Patient B, and someone clones it from the Patient Bank, it will generate a clone (Patient C) with `origin_data` set to [B].

I figured since we have origin data set to an Array, why not persist all the origins (the last item will always be most current origin)?

If so, Patient C would have `origin_data` set to [A, B].

app/controllers/patients_controller.rb
```diff
@@ -123,6 +123,8 @@ def update_patient(patient)
 
     patient.expected_values = params['expected_values']
 
+    patient['origin_data'] << params['origin_data'] if params['origin_data']
```
jtferns, line 128: I had an issue cloning patients from the Patient Bank with this, I think you have to initialize the `origin_data` before you can push elements into it (e.g., `patient['origin_data'] ||= []`).

pkmitre:
I'm not sure this does quite everything we want; it lists a users own patients that happen to be shared in the patient bank, but what I think we want is an indication of other patients that have already been cloned into this measure. For example:

User X shares patient BOB
User Y clones patient BOB into measure ABC
User Y visits the patient bank again for measure ABC
User Y should see some indication that they've already cloned patient BOB

thecristen:
Fair points. I've just added an indicator on the clone 'parents'  based on the patient ID from the first entry in each patient's `origin_data`. I thought about using all entries in origin_data but I don't think it makes much sense to show the clone of a clone of a clone, though I'm open to hearing alternative views.

pkmitre:
I think just showing first-clones is the right call.

PR #453, opened 2015-05-08
show only patients with existing associated measures

thecristen:
https://www.pivotaltracker.com/story/show/91739156

On my old copy of the production db I saw dozens of patients just vanish from the patient bank!

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -6,7 +6,7 @@ class PatientsController < ApplicationController
 
   # Index method used for patient bank, returning all shared patient records
   def index
-    records = Record.where(is_shared: true).to_a
+    records = Record.where(is_shared: true).to_a.select { |r| get_associated_measure(r).exists? } # select shared patients with existing measures
     # Some gymnastics to deal with a 1+N problem, so we don't need additional queries for each
     # record for cms_id and user_email; prepopulate these values using lookup tables, with lookups
     # for email by user_id and cms_id by user_id and measure_id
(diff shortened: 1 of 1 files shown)


app/controllers/patients_controller.rb
```diff
@@ -6,7 +6,7 @@ class PatientsController < ApplicationController
 
   # Index method used for patient bank, returning all shared patient records
   def index
-    records = Record.where(is_shared: true).to_a
+    records = Record.where(is_shared: true).to_a.select { |r| get_associated_measure(r).exists? } # select shared patients with existing measures
```
pkmitre: This approach creates a 1+N problem, meaning that there's an additional query for every record, which doesn't scale very well. If you look at the lines below, you'll see some gymnastics related to avoiding 1+N issues. There's already a query that grabs all the measures for the patient records in order to get the CMS ids; that same resulting set of measures could be used to determine which patient records to prune.

app/controllers/patients_controller.rb
```diff
@@ -16,6 +16,7 @@ def index
     measure_ids = records.map { |r| r.measure_ids.first }.uniq
     measures = Measure.only(:_id, :hqmf_set_id, :user_id, :cms_id).where(:hqmf_set_id.in => measure_ids, :user_id.in => user_ids)
     cms_lookup = measures.each_with_object(Hash.new { |h, k| h[k] = {} }) { |m, h| h[m.user_id][m.hqmf_set_id] = m.cms_id }
+    records = records.select { |r| measures.map(&:hqmf_set_id).include? r.measure_ids.first } # select shared patients with existing measures
```
thecristen, line 19: Thanks @pkmitre, I think this should work better.

PR #456, opened 2015-05-21
Add fixture data into Complexity viz to test measure sets viz

jtferns:
- used fixture data in complexity view for more viz testing
- added basic complexity graph test using svg and circle counts


thecristen:
thanks! I'm taking these changes and putting them on the `complexity_tests` branch.

PR #457, opened 2015-05-22
Allow subsequent colons in patient builder elements descriptions

jtferns:
### Story
- [CMS344 Communication elements does not display specific element info in titles](https://www.pivotaltracker.com/story/show/95241960)
### Notes
- updated data criteria description parsing to only split on the first `:`
### Screenshot

screen shot 2015-05-22 at 7 37 33 pm

### DIFF
--- a/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
+++ b/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
@@ -32,8 +32,8 @@ class Thorax.Views.SelectCriteriaItemView extends Thorax.Views.BuilderChildView
     desc = @model.get('description').split(/, (.*)/)?[1] or @model.get('description')
     _(super).extend
       type: desc.split(": ")[0]
-      # replace first ':' with '#' and split on that to allow subsequent colons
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -32,8 +32,8 @@ class Thorax.Views.SelectCriteriaItemView extends Thorax.Views.BuilderChildView
     desc = @model.get('description').split(/, (.*)/)?[1] or @model.get('description')
     _(super).extend
       type: desc.split(": ")[0]
-      detail: desc.split(": ")[1]
-
+      # replace first ':' with '#' and split on that to allow subsequent colons
+      detail: desc.replace(': ', '#').split('#')[1]
```
thecristen: Interesting approach, but what if there's ever a crazy edge case that has a `#` in it!? Maybe taking a substring might be appropriate here (like `desc.substring(desc.indexOf(':')+2)`)

jtferns: :+1:, I like that better than the `#`!

PR #459, opened 2015-06-15
Add support for Care Goals

jtferns:
### Story
- [Care Goal elements throw an error when building patients in Bonnie; also the category is missing an icon](https://www.pivotaltracker.com/story/show/96197700)
### Notes
- Updated HDS and Patient API to resolve Care Goal entry and calculation issues
- Added icon for Care Goals in Patient Builder ( @thecristen does this look good? )
- Added `related_to` and `target_outcome` field values for Care Goal patient data criteria
- Fixed a few typos
### Screenshot

screen shot 2015-06-15 at 11 19 20 am

### DIFF
--- a/lib/generation/generator.rb
+++ b/lib/generation/generator.rb
@@ -1,5 +1,5 @@
 module HQMF
-	class Generator
+  class Generator
 
     # Create a patient with trivial demographic information and no coded entries.
     #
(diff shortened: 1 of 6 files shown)


lib/generation/generator.rb
```diff
@@ -47,10 +47,12 @@ def self.classify_entry(type)
         "conditions"
       when :allDevices
         "medical_equipment"
+			when :careGoals
```
jtferns: I'm not sure why the indents are off on these lines, but I do not see this in Atom...

pkmitre: Tabs vs spaces...

PR #461, opened 2015-07-09
Allow users to change measure and calculation type

thecristen:
## Story
- https://www.pivotaltracker.com/story/show/91739674
## Notes
- _Major_ refactoring on the measures controller, breaking out the create method

### DIFF
--- a/app/assets/javascripts/views/import_measure_view.js.coffee
+++ b/app/assets/javascripts/views/import_measure_view.js.coffee
@@ -23,11 +23,15 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
       redirectRoute: currentRoute
 
   events:
-    rendered: -> 
+    rendered: ->
       @$("option[value=\"#{eoc}\"]").attr('selected','selected') for eoc in @model.get('episode_ids') if @model? && @model.get('episode_of_care') && @model.get('episode_ids')?
       @$el.on 'hidden.bs.modal', -> @remove() unless $('#pleaseWaitDialog').is(':visible')
       @$('.date-picker').datepicker('setDate', moment().format('L'))
       @$('.effective-date').hide()
+      # set the inputs as checked if update
+      if @model? and @model.get('episode_of_care') is true then @$('input[value="episode"]').attr('checked', true) else @$('input[value="patient"]').attr('checked', true)
+      if @model? and @model.get('type') is 'eh' then @$('input[value="eh"]').attr('checked', true) else @$('input[value="ep"]').attr('checked', true)
+
     'ready': 'setup'
     'change input:file':  'enableLoad'
     'keypress input:text': 'enableLoadVsac'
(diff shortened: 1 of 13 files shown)


thecristen:
Another note... the finalize measure dialog does not appear after updating a measure's calculation type to Episode of Care. I'd like it to...

pkmitre:
This works great! I'm wondering if users will figure out how to use it, ie will they realize they can update a measure's attributes without re-uploading the measure.

pkmitre:
I did just notice one issue: if you change a measure from patient based to episode of care the episode of care selection dialog doesn't appear (though it does the next time you click update on the measure).

thecristen:
@pkmitre I've just rebased and I'm no longer seeing that particular issue (I can now change a measure to an episode of care measure and get the finalize dialog). Feel free to take another look!

thecristen:
Updated with additional tests and everything seems to be in working order.

edeyoung:
If you change a measure to "episode of care" and it has no available episode of care to select from, you are stuck - the done button is disabled and there doesn't seem to be a way to cancel out.

You can see this with cervical cancer screening.

Looks like this was managed previously by checking the criteria. Attached is a screen shot loading cervical cancer from the master branch.

image

app/assets/javascripts/views/import_measure_view.js.coffee
```diff
@@ -23,42 +23,39 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
...
-    rendered: -> 
+    rendered: ->
       @$("option[value=\"#{eoc}\"]").attr('selected','selected') for eoc in @model.get('episode_ids') if @model? && @model.get('episode_of_care') && @model.get('episode_ids')?
       @$el.on 'hidden.bs.modal', -> @remove() unless $('#pleaseWaitDialog').is(':visible')
-      @$("input[type=radio]:checked").next().css("color","white")
       @$('.date-picker').datepicker('setDate', moment().format('L'))
       @$('.effective-date').hide()
+      # set the inputs as checked if update
+      if @model? and @model.get('episode_of_care') is true then @$('input[value="episode"]').attr('checked', true) else @$('input[value="patient"]').attr('checked', true)
+      if @model? and @model.get('type') is 'eh' then @$('input[value="eh"]').attr('checked', true) else @$('input[value="ep"]').attr('checked', true)
```
edeyoung, line 33: Can this be done in import_measure.hbs?

app/assets/javascripts/views/import_measure_view.js.coffee
```diff
@@ -23,42 +23,39 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
...
-    @$('#loadButton').prop('disabled', !(hasUser && hasPassword)) 
+    @$('#loadButton').prop('disabled', !(hasUser && hasPassword))
 
   enableLoad: ->
     if @$('input:file').val().match /xml$/i
       @$('#vsacSignIn').removeClass('hidden')
     else
       @$('#vsacSignIn').addClass('hidden')
-      @$('#loadButton').prop('disabled', !@$('input:file').val().length > 0)
+      # @$('#loadButton').prop('disabled', !@$('input:file').val().length > 0)
```
edeyoung, line 58: consider deleting this

edeyoung:
When I update a measure by loading in a different (or same) version of the measure (not necessarily changing the EP/EH settings, etc.), the measure looks like it duplicates:

image

test/functional/measures_controller_test.rb
```diff
@@ -233,7 +275,7 @@ class << measure_file
     measure = Measure.where({hqmf_id: "40280381-3D27-5493-013D-4DCA4B826AE4"}).first
     assert_equal "42BF391F-38A3-4C0F-9ECE-DCD47E9609D9", measure.hqmf_set_id
 
-    post :create, {measure_file: measure_file, hqmf_set_id: measure.hqmf_set_id}
+    post :create, {measure_file: measure_file, hqmf_set_id: measure.hqmf_set_id, hqmf_id: measure.hqmf_id}
```
edeyoung, line 278: should this also have been changed to `post :update`? If so, it seems like this test should have failed due to the measure already existing.

edeyoung:
When I try to upload (rather than update) a measure, the LOAD button doesn't enable.

image

pkmitre:
There appear to be a few issues with this particular pull request, and it's difficult to diagnose what's happening because of the extensive refactoring combined with the new feature. I'm going to close this pull request and suggest that, if we tackle this again, we separate out the refactoring changes from the feature work.

PR #464, opened 2015-07-24
Population state

elehner:
# Story

https://www.pivotaltracker.com/story/show/79969856
# Notes

Handles the saved state through additions to the "Bonnie" global variable.  This state is maintained on a per measure basis.

### DIFF
--- a/app/assets/javascripts/router.js.coffee
+++ b/app/assets/javascripts/router.js.coffee
@@ -29,6 +29,9 @@ class BonnieRouter extends Backbone.Router
     'value_sets/edit':                                 'renderValueSetsBuilder'
 
   renderMeasures: ->
+    _.each(@measures.models, (measure) =>
+      measure.set('displayedPopulation', measure.get('populations').first())
+    )
     @navigationSetup "Dashboard", "dashboard"
     if @isPortfolio
       dashboardView = new Thorax.Views.Matrix(collection: @measures, patients: @patients)
--- a/app/assets/javascripts/router.js.coffee
+++ b/app/assets/javascripts/router.js.coffee
@@ -29,9 +29,7 @@ class BonnieRouter extends Backbone.Router
     'value_sets/edit':                                 'renderValueSetsBuilder'
 
   renderMeasures: ->
-    _.each(@measures.models, (measure) =>
-      measure.set('displayedPopulation', measure.get('populations').first())
-    )
+    @measures.each (measure) -> measure.set('displayedPopulation', measure.get('populations').first())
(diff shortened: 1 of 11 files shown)


pkmitre:
Interesting approach, some thoughts:

1) I think the idea of storing the state, and then referring to it, is a clever way to handle this issue... but I don't like the idea of creating essentially a new global variable to store it in. Other options I can think of are a) putting it directly on the measure model or b) using the existing router class (router.js.coffee), which already puts a bonnie global variable into our namespace.

2) Function names could be clearer, ie setPopulationForMeasure doesn't say anything about this being for display state, I think something like (combining with item 1 above) measure.storeDisplayedPopulation(population) would be clearer.

3) When getting the displayed population, you have an argument for the base (default) population; this is always the same I think, so we probably don't need to pass it in

4) One issue I see with the state-storing approach is that when a user navigates back to the main page, but then later goes into a measure, the selected population is still preserved; I think that behavior would likely be confusing to a user, it should probably always default to the first population in that case.

elehner:
1a makes the most sense, so I'll try for that, and 2, 3, and 4 are very doable.

The Bonnie variable I use was already created [here](https://github.com/projecttacoma/bonnie/blob/master/app/assets/javascripts/measure_visual.js.coffee), I just use ||= in case it hasn't been called at that point.  Should we try to move what's in that variable to the bonnie (lower case b) variable that seems to be used for the other globals?

pkmitre:
I'd forgotten about the visualization code; I think that's ok to leave where it is.

elehner:
Should now have solved the aforementioned issues:
1. Uses method a, and removed file
2. No longer uses functions (uses model attributes), but renamed the variable to displayPopulation
3. Sets the base inside the measure model and on reset only
4. Partially, assumes user will have to go to a measures route before going to another measure

app/assets/javascripts/router.js.coffee
```diff
@@ -29,6 +29,9 @@ class BonnieRouter extends Backbone.Router
     'value_sets/edit':                                 'renderValueSetsBuilder'
 
   renderMeasures: ->
+    _.each(@measures.models, (measure) =>
```
pkmitre: Thorax has iterators for collections, so just

```
@measures.each (measure) -> measure.set('displayedPopulation', measure.get('populations').first())
```

Also, you only need a thin arrow here (fat arrow is if want 'this' inside the function to be the same as the encapsulating 'this', which isn't needed since you don't use 'this' aka @)

elehner: Makes sense!

PR #467, opened 2015-08-17
check if patients have origin data first

thecristen:
An error was being thrown when a patient had no origin data. This helps prevent that error.

### DIFF
--- a/app/assets/javascripts/views/patient_bank/patient_bank_view.js.coffee
+++ b/app/assets/javascripts/views/patient_bank/patient_bank_view.js.coffee
@@ -109,8 +109,8 @@ class Thorax.Views.PatientBankView extends Thorax.Views.BonnieView
     @bankFilterView.appliedFilters.all (filter) -> filter.apply(patient)
 
   differenceContext: (difference) ->
-    origins = @model.get('patients').models.map (p) -> p.get('origin_data')
-    origin_ids = _.compact(origins).map (o) -> o[0].patient_id
+    origins = @model.get('patients').map (p) -> p.get('origin_data')
+    origin_ids = _.compact(_.flatten(origins)).map (o) -> o.patient_id
     patient = difference.result.patient
     _(difference.toJSON()).extend
       patient: patient.toJSON()


app/assets/javascripts/views/patient_bank/patient_bank_view.js.coffee
```diff
@@ -110,13 +110,15 @@ class Thorax.Views.PatientBankView extends Thorax.Views.BonnieView
 
   differenceContext: (difference) ->
     origins = @model.get('patients').models.map (p) -> p.get('origin_data')
-    origin_ids = _.compact(origins).map (o) -> o[0].patient_id
+    if origins
```
pkmitre: Won't this always be true? Map should always return at least an empty array, which is truthy.

Gemfile.lock
```diff
@@ -323,6 +323,3 @@ DEPENDENCIES
   thin
   turn
   uglifier (>= 1.0.3)
-
-BUNDLED WITH
```
pkmitre: Mind taking this change out of the pull request, since it's unrelated?

Also I think we should probably all try to run the same version of bundler so we don't always get changes to this line...

app/assets/javascripts/views/patient_bank/patient_bank_view.js.coffee
```diff
@@ -110,13 +110,15 @@ class Thorax.Views.PatientBankView extends Thorax.Views.BonnieView
 
   differenceContext: (difference) ->
     origins = @model.get('patients').models.map (p) -> p.get('origin_data')
```
pkmitre: I know this is old code, but thorax collections should support map by default, so I think just

```
@model.get('patients').map
```

ought to work.

app/assets/javascripts/views/patient_bank/patient_bank_view.js.coffee
pkmitre: And, if it is in fact always true, it looks like this code will not have fixed the issue since the rest of the code is essentially the same...

thecristen:
All correct! I've reverted the initial commit and added a change that worked for me - I flatten the arrays because compacting to get rid of zero-length arrays that were contributing to this error.

PR #473, opened 2015-10-19
Bonnie 104932540 update value sets

alexanderelliott121:


app/controllers/measures_controller.rb
```diff
@@ -16,16 +16,16 @@ def show
   end
 
   def value_sets
-    if stale? last_modified: Measure.by_user(current_user).max(:updated_at).try(:utc)
-      value_set_oids = Measure.by_user(current_user).only(:value_set_oids).pluck(:value_set_oids).flatten.uniq
+   if stale? last_modified: Measure.by_user(current_user).max(:updated_at).try(:utc)
+      version_enriched_oids = Measure.by_user(current_user).only(:oid_to_version).pluck(:oid_to_version).uniq.flatten
```
pkmitre, line 20: Similar comment as other pull request, would be good to use common naming for the same concept, ie version_enriched_oids -> versioned_oids

lib/tasks/measure_migration.rake
```diff
@@ -0,0 +1,41 @@
...
+    task :migrate_measures_value_sets => :environment do
+      user_oid_to_version = Hash.new
+      oid_to_content = Hash.new
+      HealthDataStandards::SVS::ValueSet.each do |vs|
+        user_oid_to_version[[vs.oid, vs.user]] = vs.version
+        if (oid_to_content[[vs.oid, vs.version]] != nil)
+          if (vs.concepts == oid_to_content[[vs.oid, vs.version]].concepts)
+            vs.delete
+          else
+            puts "We gots a problem boss!"
```
pkmitre: Could likely use some more information in the error message. Also, generally don't have an issue with casual error messages, but for an open source project would like something a bit more formal.

lib/tasks/measure_migration.rake
```diff
@@ -0,0 +1,41 @@
...
+          end
+        else
+          oid_to_content[[vs.oid, vs.version]] = vs
+          vs.versioned_oid = vs.oid + ":::" + vs.version
```
pkmitre: If this is going to be a standard format, I think it makes sense to make it a method on ValueSet. Call the method versioned_oid, and this line can go away! Plus we might not even need to store it on the value set object in the DB, we could just always derive it.

pkmitre: Oops, except we do need the value in the DB for efficient lookups...

I do still think that having the versioned_oid construction code on the ValueSet class makes sense, so that it's in one place and more tightly coupled with ValueSet. You could have a before_save action on the ValueSet model that automatically populates the versioned_oid field.

pkmitre:
Superseded by another pull request: #505

PR #474, opened 2015-10-20
Added email active users button and statistic

edeyoung:
Added button to admin page to email users who have at least one measure loaded and have logged in within the last 6 months.
Added a statistic to the stat drop down for active users.

https://www.pivotaltracker.com/story/show/105779154


app/assets/javascripts/models/user.js.coffee
```diff
@@ -25,6 +25,11 @@ class Thorax.Collections.Users extends Thorax.Collection
       totalMeasures: @reduce(((sum, user) -> sum + user.get('measure_count')), 0)
       totalPatients: @reduce(((sum, user) -> sum + user.get('patient_count')), 0)
       activeUsers: activeUsers
+      activeRecentUsers: activeUsers.filter((user) -> 
+        date = new Date();
+        date.setMonth(date.getMonth() - 6);
+        user_date = new Date(user.get('last_sign_in_at'));
+        user_date > date)
```
pkmitre: We use the very useful moment.js library; this filter could probably be something like

```
moment().subtract(6, 'months') < moment(user.get('last_sign_in_at'))
```

config/routes.rb
```diff
@@ -38,7 +38,23 @@
...
+        post 'approve'
+        post 'disable'
+        get 'patients'
+        get 'measures'
+        get 'bundle'
+        post 'log_in_as'
+      end
+    end
+  end
+
```
pkmitre: Looks like the admin namespace is being duplicated? Should just need one with two post resources under the collection.

pkmitre:
Looks like there's a template issue. You renamed the mailer from Admin::AllUsersMailer to Admin::UsersMailer, but there's a matching template app/views/admin/all_users_mailer/  that would need to be renamed too.

FYI I tested it by trying this using the base demo dataset, with just one test user. I'm pretty sure emails don't get sent in development, they just get logged, so probably could be tested with a bigger data set too.

PR #475, opened 2015-10-25
Fix materialization

pkmitre:
The is issue is that codes are being reselected on re-materialization when they were originally auto-picked, and this is probably a bad idea; we want to keep codes that were originally picked regardless of   how, unless a white list is used (a cypress specific scenario, which lets a white list change impact all cypress patients with a simple re-materialization).

This addresses the cause of https://jira.oncprojectracking.org/browse/BONNIE-114, though does not actually correct the data issue.


pkmitre:
Good catch, definitely a silly mistake, fixed conditional.

PR #479, opened 2015-11-05
modified coloring logic

edeyoung:
Addresses https://jira.oncprojectracking.org/browse/BONNIE-97

Reworked the logic for coloring. Added specific occurrence information to the rational through change in hqmf2js. Used this to compare both specific occurrence clauses and regular logic clauses dependent on specific occurrences to the final set of specific occurrences.

### DIFF
--- a/app/assets/javascripts/models/result.js.coffee
+++ b/app/assets/javascripts/models/result.js.coffee
@@ -26,24 +26,24 @@ class Thorax.Models.Result extends Thorax.Model
     for code in Thorax.Models.Measure.allPopulationCodes
       if specifics = @get('finalSpecifics')?[code]
         updatedRationale[code] = {} 
+        
         # get the referenced occurrences in the logic tree using original population code
-        occurrences = _.uniq @population.getDataCriteriaKeys(@measure.get('population_criteria')[@population.get(code)?.code])
-        # get the good and bad specifics
-        occurrenceResults = @checkSpecificsForRationale(specifics, occurrences, @measure.get('data_criteria'))
+        criteria = _.uniq @population.getDataCriteriaKeys(@measure.get('population_criteria')[@population.get(code)?.code], false)
+        criteriaResults = @checkCriteriaForRationale(specifics, criteria, rationale, @measure.get('data_criteria'))
         submeasureCode = @population.get(code)?.code || code
         parentMap = @buildParentMap(@measure.get('population_criteria')[submeasureCode])
 
         # check each bad occurrence and remove highlights marking true
-        for badOccurrence in occurrenceResults.bad
-          if (rationale[badOccurrence])
-            updatedRationale[code][badOccurrence] = false
+        for badCriteria in criteriaResults.bad
+          if (rationale[badCriteria])
+            updatedRationale[code][badCriteria] = false
             # move up the logic tree to set AND/ORs to false based on the removal of the bad specific's true eval
-            @updateLogicTree(updatedRationale, rationale, code, badOccurrence, orCounts, parentMap)
+            @updateLogicTree(updatedRationale, rationale, code, badCriteria, orCounts, parentMap)
         # check the good specifics with a negated parent.  If there are multiple candidate specifics
         # and one is good while the other is bad, the child of the negation will evaluate to true, we want it to
         # evaluate to false since if there's a good negation then there's an occurrence for which it evaluated to false
-        for goodOccurrence in occurrenceResults.good
-          @updatedNegatedGood(updatedRationale[code], rationale, goodOccurrence, parentMap)
+        for goodCriteria in criteriaResults.good
+          @updatedNegatedGood(updatedRationale[code], rationale, goodCriteria, parentMap)
     return updatedRationale
 
   updatedRationale: ->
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/models/result.js.coffee
```diff
@@ -63,21 +63,47 @@ class Thorax.Models.Result extends Thorax.Model
...
+        if criteriaSpecificId != "*"
+          return false
+      return true
+      
+    for finalSpecificIds in finalSpecificIdsSet
+      matches = true
+      # should have the same length
+      for i in [0..criteriaSpecificIds.length]
+        criteriaSpecificId = criteriaSpecificIds[i]
+        finalSpecificId = finalSpecificIds[i]
```
pkmitre: A bit code-golfy, but this could be something like

```
for criteriaSpecificId, idx in criteriaSpecificIds
    finalSpecificId = finalSpecificIds[idx]
```

app/assets/javascripts/models/result.js.coffee
```diff
@@ -63,21 +63,47 @@ class Thorax.Models.Result extends Thorax.Model
...
+        
+        if criteriaSpecificId != finalSpecificId && criteriaSpecificId != "*" && finalSpecificId != "*"
+          matches = false
+
+      if matches
+        return true
+
+    return false
```
pkmitre: Thinking aloud a bit, would the preceding code be clearer using JavaScript's "every" and/or "some" methods?

```
return finalSpecificIdsSet.some (finalSpecificIds) ->
    criteriaSpecificIds.every (criteriaSpecificId, idx) ->
        finalSpecificId = finalSpecificIds[idx]
        criteriaSpecificId == finalSpecificId || criteriaSpecificId == "*" || finalSpecificId == "*"
```

(Tried to get all the boolean senses correct in there, but didn't test...)

app/assets/javascripts/models/result.js.coffee
```diff
@@ -63,21 +63,47 @@ class Thorax.Models.Result extends Thorax.Model
...
-      # we're good if the occurrence is referenced by ID, Any indicates that we did not use it in the true path, thus it's bad
-      for row in finalSpecifics
-        good = true if row[index] != hqmf.SpecificsManager.any
-      if good then results.good.push(occurrence) else results.bad.push(occurrence)
+  checkCriteriaForRationale: (finalSpecifics, criteria, rationale, dataCriteriaMap) ->
+    results = {bad: [], good: []}
+
+    for criterion in criteria
+      criterionRationale = rationale[criterion]
+      if criterionRationale == false || typeof criterionRationale.specifics == 'undefined' || criterionRationale.specifics.length == 0
```
pkmitre: Could probably use

```
!criterionRationale.specifics?
```

in there.

app/assets/javascripts/models/result.js.coffee
```diff
@@ -63,21 +63,47 @@ class Thorax.Models.Result extends Thorax.Model
...
+        results.good.push(criterion)
+      else
+        matches = false
+        for criteriaSpecificIds in criterionRationale.specifics
+          if @idsMatch(criteriaSpecificIds, finalSpecifics)
+            matches = true
+            break
+
+        if matches then results.good.push(criterion) else results.bad.push(criterion)
```
pkmitre: Another possible use of some / every?

```
if criterionRationale.specifics.some (criteriaSpecificIds) -> @idsMatch(criteriaSpecificIds, finalSpecifics)
    results.good.push(criterion)
else
    results.bad.push(criterion)
```

Not sure if it's an improvement, what do you think?

edeyoung: wasn't familiar with some/every. This gets the intent across nicely. I'll make the change and test it.

app/assets/javascripts/models/result.js.coffee
edeyoung: combining the two is a bit confusing to me. I like the idea though. I'll play around with it and see if I can make it into something that reads intuitively to me.

app/assets/javascripts/models/result.js.coffee
edeyoung: i think i agree that it's a bit code-golfy. i kind of like having a single intent for `for` loops. I'm leaning towards keeping it as is.

app/assets/javascripts/models/result.js.coffee
pkmitre: Ruby uses any? and all?, which somehow I find less confusing to work with. "some" just seems a bit ambiguous to me, which I think takes away a bit of clarity...

app/assets/javascripts/models/result.js.coffee
```diff
@@ -63,22 +63,42 @@ class Thorax.Models.Result extends Thorax.Model
...
+  idsMatch: (criteriaSpecificIds, finalSpecificIdsSet) ->
+    # if we have no criteria specifics, then there are no specific requirements
+    # for the criteria
+    if criteriaSpecificIds.length == 0
+      return true
+    # if we have no specifics in the final specifics list, all criteria specifics 
+    # should be wild cards.
+    if finalSpecificIdsSet.length == 0
+      return criteriaSpecificIds.every (criteriaSpecificId) ->
+        return criteriaSpecificId == "*"
```
pkmitre: I find this use of return confusing, I have to think about whether it's at the top level function vs a helper function, plus the typical CoffeeScript style is to rely on the implicit return. I think removing the return and making it a single line would help, ie

```
if finalSpecificIdsSet.length == 0
    return criteriaSpecificIds.every (criteriaSpecificId) -> criteriaSpecificId == "*"
```

PR #481, opened 2015-11-09
Uses display names from value sets for criteria descriptions

elehner:
Addresses https://jira.oncprojectracking.org/browse/BONNIE-109

The code was improperly filtering on the base description.  This now applies the value set display name to correct what is shown and prevent unintentional conflicts.

### DIFF
--- a/app/assets/javascripts/models/measure.js.coffee
+++ b/app/assets/javascripts/models/measure.js.coffee
@@ -22,7 +22,7 @@ class Thorax.Models.Measure extends Thorax.Model
     for key, data_criteria of attrs.data_criteria
       data_criteria.key = key
       # Apply value set display name if one exists for this criteria
-      if bonnie.valueSetsByOid[data_criteria.code_list_id]?.display_name?
+      if !data_criteria.variable && bonnie.valueSetsByOid[data_criteria.code_list_id]?.display_name?
         data_criteria.description = "#{data_criteria.description.split(':')[0]}: #{bonnie.valueSetsByOid[data_criteria.code_list_id].display_name}"
       if data_criteria.field_values
         data_criteria.references = {}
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/models/measure.js.coffee
```diff
@@ -21,6 +21,9 @@ class Thorax.Models.Measure extends Thorax.Model
 
     for key, data_criteria of attrs.data_criteria
       data_criteria.key = key
+      # Apply value set display name if one exists for this criteria
+      if bonnie.valueSetsByOid[data_criteria.code_list_id]?.display_name?
+        data_criteria.description = "#{data_criteria.description.split(':')[0]}: #{bonnie.valueSetsByOid[data_criteria.code_list_id].display_name}"
```
pkmitre: I can see why we need to patch the description here rather than get it in the HQMF parser (we don't have the value set name there). How well does this work for SimpleXML measures, for HQMF R1 measures, etc? Should be fine, but probably worth testing...

edeyoung:
Code looks good and things behave as expected.

This seems to address the issue by using the display name for the oid in the display name for the criteria. However, it looks like criteria are still being filtered by display name. Is there a chance that the display name for two different value sets would be the same?  If so, I expect we'd still want to display both criteria (even though it would be confusing). Should wherever the filtering is happening be modified as well to filter on oid code rather than display name? Or, if this could happen, would it be so much of an edge case that it wouldn't be worth it?

If is doesn't make sense to do the changes above, then this looks good to merge in.

elehner:
It's not really so much an "edge case" as a problem this should solve. By modifying the display names to the value set display name (based on the oids) this should solve the problem. Are there any instances where the value set display names are the same but the oids are different?

edeyoung:
Yeah, that's what I was wondering as well.

elehner:
Added a check against the oid to the description check, should prevent that conflict.

edeyoung:
Could the decision be based just on the code_list_id rather than on both the code_list_id and the description?

elehner:
I was thinking about that, but I was wondering if there was a case when the oid field would be blank and we would want to keep two elements with blank oids, or they have the same oid with no map, as long as they had varying descriptions.

edeyoung:
Makes sense. I’ll merge it.

PR #482, opened 2015-11-13
Default patient MRN source

lrasmus:
Provide a default MRN source of "Bonnie" for generated patients.  Otherwise, when exporting they all come out with the same (default) MRN of "12345".


pkmitre:
Merged via e16bdefd2eef16d711ae8469429731dd156611dd

pkmitre:
Thanks for the pull request! If you're using the healthit.gov version of Bonnie, this pull request should be deployed at the end of the day today (11/16/2015).

PR #484, opened 2015-11-16
added list of value sets used by the measure at the end of the measur…

edeyoung:
…e page

### DIFF
--- a/app/assets/javascripts/helpers.js.coffee
+++ b/app/assets/javascripts/helpers.js.coffee
@@ -24,9 +24,9 @@ Handlebars.registerHelper 'ifPortfolio', (options) ->
 
 Handlebars.registerHelper 'ifCond', (v1, operator, v2, options) ->
   switch operator
-    when '==' 
+    when '=='
       if (v1 == v2) then options.fn(this) else options.inverse(this)
-    when '!=' 
+    when '!='
       if (v1 != v2) then options.fn(this) else options.inverse(this)
     when '<'
       if (v1 < v2) then options.fn(this) else options.inverse(this)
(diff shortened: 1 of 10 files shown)


thecristen:
Hey, this is cool!

Visually, it doesn't feel very comfortable to scan - I think we can alleviate that by sorting the list first (sorting by description, so that all the Encounter, Performed and such end up together), and maybe lining up the titles?

Also it doesn't seem to handle negations well - for example I get the following on CMS135:

```
"Medication, Order: ACE Inhibitor or ARB Ingredient" using "ACE Inhibitor or ARB Ingredient (2.16.840.1.113883.3.526.3.1489)"
"Medication, Order not done: ACE Inhibitor or ARB Ingredient" using "Medical Reason (2.16.840.1.113883.3.526.3.1489)"
"Medication, Order not done: ACE Inhibitor or ARB Ingredient" using "Patient Reason (2.16.840.1.113883.3.526.3.1489)"
"Medication, Order not done: ACE Inhibitor or ARB Ingredient" using "System Reason (2.16.840.1.113883.3.526.3.1489)"
```

All the same OID, not really informative or clear.. might be better off to just filter out negations.

edeyoung:
Thanks for the comments! I'll make those adjustments and try a few things with alignments.

thecristen:
I was playing around with it and got something more like the following: 
image

If you like it feel free to cherry-pick it off `value_set_list_sorted` [(relevant commit)](https://github.com/projecttacoma/bonnie/commit/63ef97d7409ef5d3bf336e0c463b7d533fe57a75) :)

pkmitre:
Since this is really tabular data, how about a table?

screen shot 2015-11-30 at 4 08 05 pm

I haven't figured out exactly what would happen when clicking on a row...

pkmitre:
Also, I prefer the OIDs to be left justified, they often have common prefixes and that's easier to parse out if they are left aligned.

edeyoung:
regarding the table view, yeah, that was the issue I was running into. I like having the expand view vs a pop up or something (which maybe we'd need to do if we did the table view). With the expand view, a person could have multiple lists expanded. I feel like this could be useful.

pkmitre:
I like how the table view looks (I find it easier to read), but we could certainly style a non-table approach to look like the table, and still have something reasonable happen with expanding lists so that we can get the view-multiple-at-a-time feature (which I agree is useful).

PR #488, opened 2015-11-20
Ignores specific variables in measure view

elehner:
Only uses the base variable for the measure view. The front-end portion of the solution to the problem described here:
https://jira.oncprojectracking.org/browse/BONNIE-70


pkmitre:
Probably easiest to just cherry pick this commit over into that branch, and I'll just look at both.

elehner:
Was cherry-picked on top of reg_occ_coloring

PR #490, opened 2015-11-25
Any type of union will now be considered like an 'or'

edeyoung:
Before, only criteria with `UNION` as the operator and with `UNION` included in the name would be added to the `orCounts` structure. There are instances where criteria have `UNION` as the operator, but the name includes `union` or `GROUP`. These will now also be included in the `orCounts` structure.


edeyoung:
This pull request is not yet ready. It causes unexpected changes.

PR #492, opened 2015-12-04
handled bug expressed in the demo version of the database with patien…

edeyoung:
…ts in CMS128v2. for some reason, there was a data criteria that was not present in the rationale. added code to handle this gracefully. still need to figure out why this is the case.

### DIFF
--- a/app/assets/javascripts/models/result.js.coffee
+++ b/app/assets/javascripts/models/result.js.coffee
@@ -87,6 +87,13 @@ class Thorax.Models.Result extends Thorax.Model
 
     for criterion in criteria
       criterionRationale = rationale[criterion]
+
+      # handle the case where the rationale does not contain a criteria
+      if !criterionRationale?
+        if console?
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/models/result.js.coffee
```diff
@@ -87,6 +87,13 @@ class Thorax.Models.Result extends Thorax.Model
 
     for criterion in criteria
       criterionRationale = rationale[criterion]
+
+      # handle the case where the rationale does not contain a criteria
+      if !criterionRationale?
+        if console?
+          console.log('WARNING: data criteria ' + criterion + ' is not contained in the rationale')
```
pkmitre: A coffeescripty way to use the existential operator (?), as well as coffeescript string substitution, would be to replace

```
if console?
  console.log('WARNING: data criteria ' + criterion + ' is not contained in the rationale')
```

with

```
console?.log("WARNING: data criteria #{criterion} is not contained in the rationale")
```

PR #493, opened 2015-12-07
in the patient editor, only one result can be added. 

edeyoung:
makes it so the view for adding more results goes away when a result is present. this does not take into account the data criteria type. it is possible we will need to revisit this and make specific to data criteria types.

### DIFF
--- a/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
+++ b/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
@@ -2,6 +2,7 @@
 class Thorax.Views.BuilderChildView extends Thorax.Views.BonnieView
   events:
     ready: -> @patientBuilder().registerChild this
+    
   patientBuilder: ->
     parent = @parent
     until parent instanceof Thorax.Views.PatientBuilder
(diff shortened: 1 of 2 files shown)


pkmitre:
This approach isn't particularly thoraxy. Specifically, directly manipulating the DOM state should be avoided when possible. Another approach would be to provide the display-it-or-not state as part of the context, and just have the view rendering use that to determine whether to show that the control.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -182,6 +183,8 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
     e.preventDefault()
     $(e.target).model().destroy()
     @triggerMaterialize()
+    @editValueView.render()
+    @editValueView.populate()
```
pkmitre: I don't think we need the call to populate() with the approach using hide in the HTML (though I think we do need render() right?)

edeyoung: yes and yes. removed populate(). still need render().

PR #494, opened 2015-12-08
Allows users to add custom codes to a data criteria.

edeyoung:
image
image
image

### DIFF
--- a/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
+++ b/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
@@ -197,13 +197,19 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
 class Thorax.Views.CodeSelectionView extends Thorax.Views.BuilderChildView
   template: JST['patient_builder/edit_codes']
   events:
-    'change select':           'validateForAddition'
-    'change .codeset-control': 'changeConcepts'
+    'change select[name=codeset]': (e) ->
+      @model.set codeset: $(e.target).val()
+      @changeConcepts(e)
+      @validateForAddition()
+    'change select[name=code]' : ->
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -195,12 +195,19 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
...
+      @validateForAddition()
+    'change select[name=code]' : ->
+      @validateForAddition()
+    'keyup input[name=custom_code]': 'validateForAddition'
+    'keyup input[name=custom_codeset]': 'validateForAddition'
     rendered: ->
       @$('select.codeset-control').selectBoxIt('native': true)
 
   initialize: ->
+    @custom_key = 'custom'
```
pkmitre: I'm not sure I see the value in having a variable to hold the string 'custom', rather than just using the string 'custom' in the appropriate places. If the goal is to prevent silent failures if one of the instances is changed without changing the others, the use of the variable still has that issue because of how handlebars templates work (they'll silently handle whatever is there even if it's not defined, i think).

edeyoung: I was trying to stay away from a string version of a magic number. 'custom' would be used 4 times in the coffee script and 3 times in the template. This approach seemed less error prone if the 'custom' string needed to change, etc.

I like keeping it in a variable to stay away from magic number issues. Is there a more coffee-script/handlebars-esque way to do it?

pkmitre:
Another thought: this feature interacts with the warning message for criteria that don't have codes from the measure, ie a warning message might be displayed if this feature is used. I think that's ok, and may actually be desirable, but want to make sure we consider it.

PR #495, opened 2015-12-10
VSAC credentials are now remembered per session for measure importing

holmesie:


pkmitre:
Ah, I assumed it was a new function, thanks!

pkmitre:
I took a look at the VSAC API, and I think the right thing to do is actually use something built into the API: instead of getting a Ticket as the first step (which is what we do now, it only lasts for a few minutes), get a CAS TicketGrantingTicket, which is valid for 8 hours. That's something that can be stored in the user's session (on the server side, not exposed to the user). There will be a few more moving parts (the front end and back end will need to coordinate regarding whether a valid TicketGrantingTicket is present), and we'll want to think about how the API itself will change with that approach. We may want to move to the V2 API as part of this too. More work, but a more secure approach.

holmesie:
I agree, closing this pull request.

PR #497, opened 2015-12-16
Removing populationCodes from rationaleCriteria to fix BONNIE-124

notpace:
BONNIE-124: Some measures show a test coverage of 0% even when test patients exist

### DIFF
--- a/app/assets/javascripts/models/coverage.js.coffee
+++ b/app/assets/javascripts/models/coverage.js.coffee
@@ -14,7 +14,7 @@ class Thorax.Model.Coverage extends Thorax.Model
       result = difference.result
       rationale = result.get('rationale')
       @rationaleCriteria.push(criteria) for criteria, result of rationale when result
-    @rationaleCriteria = _.unique(@rationaleCriteria)
+    @rationaleCriteria = _.difference(_.unique(@rationaleCriteria), Thorax.Models.Measure.allPopulationCodes)
 
     # Set coverage to the fraction of measure criteria that were true in the rationale
     @set coverage: ( @rationaleCriteria.length * 100 / @measureCriteria.length ).toFixed()


app/assets/javascripts/models/coverage.js.coffee
```diff
@@ -14,7 +15,7 @@ class Thorax.Model.Coverage extends Thorax.Model
       result = difference.result
       rationale = result.get('rationale')
       @rationaleCriteria.push(criteria) for criteria, result of rationale when result
-    @rationaleCriteria = _.unique(@rationaleCriteria)
+    @rationaleCriteria = _.difference(_.unique(@rationaleCriteria), @populationCodes)
```
edeyoung: is it sufficient to just remove `populationCodes`? Is it possible for other items to be in `rationaleCriteria` that wouldn't be conatined in `measureCriteria`? Based on the lines below, it looks like `rationaleCriteria` should be a subset of `measureCriteria`. Maybe we should remove any codes in `rationaleCriteria` that aren't in `measureCriteria`.

Also, (and I'm guessing this is a change I made... I can discuss with you where I think this might have happened), why is this breaking now? Is more information going into `rationaleCriteria` than before, and is there something we can change in the creation of `rationaleCriteria` rather than removing things here after the fact?

notpace: I'm not confident that it is sufficient to just remove `populationCodes`, but some poking around in the rendered page made me convinced that at the very least they need to be removed.  I have a rough idea of how all this works functionally, but I'm still getting to know the codebase.

Also yeah I'm guessing this was commit https://github.com/projecttacoma/bonnie/commit/08f73c74b43f77f6865b2177348b5e9c1d1eff0e that sought to add variables to the items included in the `rationaleCriteria`.

edeyoung: Yeah, that was definitely it. If I remember correctly, I made that change to enable the non-occurrenced variables to be colored. The problem was, only occurrenced variables were being colored. The coloring is determined by what's in rationaleCriteria. Even when non-occurrenced variables got added to the rationale (by forcing a calculation), they got removed here because they weren't a part of `measureCriteria`.

A possible solution may be to add back in `_(@rationaleCriteria).intersection(@measureCriteria)` but add it to another variable, and then use that variable to calculate the coverage.

edeyoung: it's probably also worth poking around to see if there are any other side affects from expanding `rationaleCriteria`.

pkmitre: I think I like @edeyoung's solution to "to add back in _(@rationaleCriteria).intersection(@measureCriteria) but add it to another variable, and then use that variable to calculate the coverage", because I think it's most likely to ensure that coverage amounts match what we did before the bug was introduced, and I think consistency has value here.

pkmitre: Just to confirm a verbal conversation: while consistency has value, getting good data to users is also important; we were missing variables in the coverage percentage, and adding them in is good.

pkmitre: One tweak: I'd prefer to skip the use of the @populationCodes variable, and just do

```
@rationaleCriteria = _.difference(_.unique(@rationaleCriteria), Thorax.Models.Measure.allPopulationCodes)
```

I'm a big fan of code locality when possible, it's one less mental jump to figure out what's being subtracted out. (This only works because the original constant has a reasonable name, an intermediate variable can be useful to increase clarity in some cases when there are naming issues).

notpace: Sounds reasonable - I can make that update.

pkmitre:
I think this fix needs more thought, it doesn't actually address issues with already loaded measures, they still display as 0 due to a value over 100.

I think the original approach of doing

```
@rationaleCriteria = _(@rationaleCriteria).intersection(@measureCriteria)
```

is probably the best bet, and instead address the missing variable highlighting by looking at the code that constructs @measureCriteria and rethinking the approach it uses to select criteria (adding variable criteria as needed)

notpace:
In that case, I think we just need to revert commit 08f73c7 for a temporary fix to the coverage% >100.

PR #498, opened 2015-12-22
New feature to export patients to an Excel spreadsheet

jbradl11:
Added Excel export functionality, along with an export button on Measure page. Refactored current QRDA export code.


Gemfile
```diff
@@ -31,7 +31,7 @@ gem 'hquery-patient-api', :git => 'https://github.com/projecttacoma/patientapi.g
 
 # needed for HDS
 gem 'rubyzip', '< 1.0.0'
-
+gem 'axlsx'
```
pkmitre: Taking out the blank line makes it seem like the comment applies to all the gems, not just rubyzip

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -55,8 +55,22 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
...
+    @model.get('populations').whenDifferencesComputed =>
+      differences = []
+      @model.get('populations').each (population) ->
+        differences.push(_(population.differencesFromExpected().toJSON()).extend(population.coverage().toJSON()))
+
+      $.fileDownload "patients/excel_export?hqmf_set_id=#{@model.get('hqmf_set_id')}", 
+        successCallback: => @exportPatientsView.excelSuccess()
         failCallback: => @exportPatientsView.fail()
         httpMethod: "POST"
         data: {authenticity_token: $("meta[name='csrf-token']").attr('content'), results: differences }
```
pkmitre, line 78: It doesn't look like the excel_export action in the patients controller makes use of the results parameter, like the qrda_export action does. In that case, do we need to calculate and send the differences for the excel version?

Actually... I didn't realize this is how the existing export works! Do these differences contain the calculation results, and could we have used these instead of re-calculating on the back-end? Hmmm.

app/controllers/patients_controller.rb
```diff
@@ -97,6 +97,20 @@ def export
     send_data stringio.sysread, :type => 'application/zip', :disposition => 'attachment', :filename => filename
   end
 
+  def excel_export
+    #Grab all records for the given measure
```
pkmitre: Typical ruby style is to include a space between the # and the comment, ie 

```
# Grab all records for the given measure
```

instead of

```
#Grab all records for the given measure
```

app/controllers/patients_controller.rb
```diff
@@ -97,6 +97,20 @@ def export
...
+    measure = Measure.where(hqmf_set_id: params[:hqmf_set_id], user_id: current_user.id).first
+    records = Record.by_user(current_user).where({:measure_ids.in => [params[:hqmf_set_id]]})   
+    package = nil
```
pkmitre: Why are we setting package to nil here? I don't think this is needed.

app/assets/javascripts/views/measure_view.js.coffee
thecristen, line 78: I'm under the impression that `differencesFromExpected()` do contain calculation results! Hmm indeed.

app/controllers/patients_controller.rb
```diff
@@ -97,6 +97,20 @@ def export
...
+    #Only generate excel document if there are patients for the given measure.
+    if records.length > 0
+      cookies[:fileDownload] = "true" # We need to set this cookie for jquery.fileDownload
+      package = PatientExport.export_excel_file(measure, records)
+      send_data package.to_stream.read, type: "application/xlsx", filename: "#{measure.cms_id}.xlsx"
+    else
+      render :json => measure
```
pkmitre: Why are we sending JSON for the measure if we have no patient records? How about

```
render nothing: true
```

config/initializers/numeric.rb
```diff
@@ -0,0 +1,13 @@
...
+    return "" if self < 1
+    s, q = "", self
+    loop do
+      q, r = (q - 1).divmod(26)
+      s.prepend(Alph[r])
+      break if q.zero?
+    end
+    s
+  end
+end
```
pkmitre: Why are we extending the Numeric class in an initializer? Some issues:

1) No comments explaining what this does

2) Looks like it's trying to change a number into a column header string... why not just make a small helper method where this is used?

lib/patient_export.rb
```diff
@@ -0,0 +1,258 @@
+class PatientExport
```
pkmitre, line 2: Maybe some brief comments at the top describing what this is for and where it's used

lib/patient_export.rb
```diff
@@ -0,0 +1,258 @@
...
+
+  def self.export_excel_file(measure, records)
+    #List of attributes we want to print to excel.
+    @@attributes = ['notes', 'first', 'last', 'birthdate', 'expired', 'deathdate',
```
pkmitre: For stuff like this, probably better to use constants, ie

```
class PatientExport

  DISPLAYED_ATTRIBUTES = ['notes', 'first', 'last', 'birthdate', 'expired', 'deathdate', 'ethnicity', 'race', 'gender']
```

lib/patient_export.rb
```diff
@@ -0,0 +1,258 @@
...
+            'ethnicity', 'race', 'gender']
+
+    #Hash of populcation codes, and number of data criteria for each
+    @@population_hash = Hash.new
```
pkmitre: Since we never instantiate this class, I think @population_hash would be fine

lib/patient_export.rb
```diff
@@ -0,0 +1,258 @@
...
+
+    Axlsx::Package.new do |package|
+      package.workbook do |workbook|
+
+        # Styles
+        fg_color = "000033"
+        header_border = { :style => :thick, :color =>"000066", :edges => [:bottom] }
```
pkmitre: Consider making these class constants

lib/patient_export.rb
```diff
@@ -0,0 +1,258 @@
...
+
+        needs_fix = styles.add_style(:sz => 14,
+                                  :bg_color => "FFFFFFF",
+                                  :border => { :style => :thin,
+                                              :color =>"DDDDDD",
+                                              :edges => [:bottom] },
+                                  :fg_color => "FF0000" )
+
+        #Adding a new sheet per population
+        measure.populations.each_with_index do |val, population_index|
```
pkmitre: I like more descriptive variable names like "population" instead of "val"

app/assets/javascripts/views/measure_view.js.coffee
pkmitre, line 78: The key here may be that these calculation results may contain the specifics rationale, or at least provide a way for us to pass them back to backend without having to move the specifics rationale calculation to the back end (though that's probably the right thing to do long term anyway).

PR #499, opened 2015-12-23
Updated to a newer version of bootstrap-datepicker to fix z-indexing

hossenlopp:
Updated to a newer version of bootstrap-datepicker. This fixes the z-indexing issue when selecting the effective date when importing a measure.


pkmitre:
You mentioned that the date picker default location has moved to over vs under, and I didn't think that would be an issue, but I tried it out and it actually feels awkward to me. Is there a way to change the behavior so that it works as it used to?

Also, it looks like you may have tried to rebase this pull request, but it looks like there's a merge on it now, which we should clear up; we just want a single commit on this branch that contains the changes.

hossenlopp:
I redid my changes against the latest revision of master. This single commit updates the library and initializes the datepickers with the 'bottom left' orientation option.

pkmitre:
The only difference I see now is that if the field is scrolled to the bottom of the page, it no longer automatically moves the popup to the top. I think that's ok, and I'll merge this.

PR #500, opened 2015-12-27
Remembering VSAC credentials

holmesie:
After the user enters their VSAC credentials when importing a measure, the resulting VSAC API ticket granting ticket will be stored in the users session storage for future use. For the next 7.5hrs, this ticket granting ticket will be used to retrieve service tickets for VSAC API calls made by this user (rather than retrieving a new ticket granting ticket every single time a VSAC API call is made, which requires the user to renter their VSAC username/password each time). The user can optionally "log out" of an active VSAC API session (the ticket granting ticket is thrown out).

### DIFF
--- a/app/assets/javascripts/views/import_measure_view.js.coffee
+++ b/app/assets/javascripts/views/import_measure_view.js.coffee
@@ -43,6 +43,7 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
           @$(element).next().css("color","")
     'focus input': (e) -> if not @$(e.target).hasClass('date-picker') and $('.datepicker').is(':visible') then @$('.date-picker').datepicker('hide')
     'change input[name="include_draft"]': 'toggleDraft'
+    'click #clearVSACCreds': 'clearCachedVSACTicket'
 
   enableLoadVsac: ->
     username = @$('#vsacUser')
@@ -55,9 +56,33 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
       hasPassword = true
     @$('#loadButton').prop('disabled', !(hasUser && hasPassword)) 
 
+  clearCachedVSACTicket: ->
+    @$('#vsacSignIn').removeClass('hidden')
+    @$('#vsacCachedMsg').addClass('hidden')
+    @$('#loadButton').prop('disabled', true)
+    $.post '/measures/vsac_auth_expire' 
+
+  toggleVSAC: ->
+    $.ajax
+      url: '/measures/vsac_auth_valid'
+      success: (data, textStatus, jqXHR) ->
+        if data? && data.valid
+          $('#vsacSignIn').addClass('hidden')
+          $('#vsacCachedMsg').removeClass('hidden')
+          $('#loadButton').prop('disabled', false) 
+          # If the measure import window is open long enough for the VSAC
+          # credentials to expire, we need to reshow the username and
+          # password dialog.
+          setTimeout ->
+            @clearCachedVSACTicket()
+          , new Date(data.expires) - new Date()
+        else
+          $('#vsacSignIn').removeClass('hidden')
+          $('#vsacCachedMsg').addClass('hidden')   
+
   enableLoad: ->
     if @$('input:file').val().match /xml$/i
-      @$('#vsacSignIn').removeClass('hidden')
+      @toggleVSAC()
     else
       @$('#vsacSignIn').addClass('hidden')
       @$('#loadButton').prop('disabled', !@$('input:file').val().length > 0)
--- a/app/controllers/measures_controller.rb
+++ b/app/controllers/measures_controller.rb
@@ -77,7 +77,7 @@ class MeasuresController < ApplicationController
         unless includeDraft
           effectiveDate = Date.strptime(params[:vsac_date],'%m/%d/%Y').strftime('%Y%m%d')
         end
-        measure = Measures::SourcesLoader.load_measure_xml(params[:measure_file].tempfile.path, current_user, params[:vsac_username], params[:vsac_password], measure_details, true, false, effectiveDate, includeDraft) # overwrite_valuesets=true, cache=false, includeDraft=true
+        measure = Measures::SourcesLoader.load_measure_xml(params[:measure_file].tempfile.path, current_user, params[:vsac_username], params[:vsac_password], measure_details, true, false, effectiveDate, includeDraft, get_ticket_granting_ticket) # overwrite_valuesets=true, cache=false, includeDraft=true
       else
         measure = Measures::MATLoader.load(params[:measure_file], current_user, measure_details)
       end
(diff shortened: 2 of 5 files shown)


thecristen:
Yes!
Is it possible to show the active VSAC username here? If so, I think the logged in message would be clearer if changed to something like "Logged in as **blahblah@site.com**. [Log out?]()" (Note: while Github automagically linked that email address I wouldn't expect it to show as linked here..)

holmesie:
Great idea Cristen, I just pushed a commit with that change - so now it looks something like (using a bogus username):

screen shot 2015-12-28 at 7 24 17 am

app/assets/javascripts/views/import_measure_view.js.coffee
```diff
@@ -53,9 +54,28 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
...
+
+  toggleVSAC: ->
+    if Cookies.get('tgt')?
+      @$('#vsacSignIn').addClass('hidden')
+      @$('#vsacCachedMsg').removeClass('hidden')
+      @$('#loadButton').prop('disabled', false) 
+      # Auto disable cached VSAC credentials when the cookie expires
+      setTimeout -> 
+        $("#clearVSACCreds").click();
+      , Math.abs(new Date(JSON.parse(Cookies.get('tgt')).expires) - new Date())
```
edeyoung: Why is `Math.abs` used here? Seems like it should just be the difference. Seems like `Math.abs` may open the door for weird edge cases. For example, if a cookie expires as 15:00, I'll get the same result from this if `toggleVSAC` gets called at 12:00 or 18:00.

app/assets/javascripts/views/import_measure_view.js.coffee
```diff
@@ -53,9 +54,28 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
       hasPassword = true
     @$('#loadButton').prop('disabled', !(hasUser && hasPassword)) 
 
+  clearCachedVSACTicket: ->
+    Cookies.remove('tgt')
+    @$('#vsacSignIn').removeClass('hidden')
+    @$('#vsacCachedMsg').addClass('hidden')
+    @$('#loadButton').prop('disabled', true)
```
edeyoung: I was working on something similar a few weeks ago where things were hidden/shown depending on state. Pete mentioned that a best practice is to manage this as much as possible in the `.hbs` file and to limit changing the dom as much as possible in the `.js` files. Would it be possible to refactor some of this into the `.hbs` file?

ttaylor249:
Is there a reason this is stored in a cookie returned to the user's 
browser rather than storing it in the user's session on the server side?

On 12/27/2015 04:25 PM, Adam M. Holmes wrote:

> After the user enters their VSAC credentials when importing a
> measure, the resulting VSAC API ticket granting ticket will be
> stored in a cookie. For the next 7.5hrs, this ticket granting ticket
> will be used to retrieve service tickets for VSAC API calls made
> by this user (rather than retrieving a new tgt every single time
> a VSAC API call is made). The user can optionally "log out" of
> an active VSAC API session (the tgt is thrown out).
> 
> screen shot 2015-12-27 at 4 24 45 pm
> https://cloud.githubusercontent.com/assets/14923551/12012149/5ec4b334-acb6-11e5-8432-946cc70f4d1e.png
> 
> ---
> 
> ```
>     You can view, comment on, or merge this pull request online at:
> ```
> 
> https://github.com/projecttacoma/bonnie/pull/500
> 
> ```
>     Commit Summary
> ```
> - Remembering VSAC credentials
> - updated get_ticket_granting_ticket to use correct VSApi method
>   
>   ```
>   File Changes
>   ```
> - _M_ app/assets/javascripts/application.js.coffee
>   https://github.com/projecttacoma/bonnie/pull/500/files#diff-0 (1)
> - _M_ app/assets/javascripts/templates/import/import_measure.hbs
>   https://github.com/projecttacoma/bonnie/pull/500/files#diff-1 (8)
> - _M_ app/assets/javascripts/views/import_measure_view.js.coffee
>   https://github.com/projecttacoma/bonnie/pull/500/files#diff-2 (17)
> - _M_ app/controllers/measures_controller.rb
>   https://github.com/projecttacoma/bonnie/pull/500/files#diff-3 (20)
> - _A_ vendor/assets/components/js-cookie/js.cookie.js
>   https://github.com/projecttacoma/bonnie/pull/500/files#diff-4 (145)
>   
>   ```
>   Patch Links:
>   ```
> - https://github.com/projecttacoma/bonnie/pull/500.patch
> - https://github.com/projecttacoma/bonnie/pull/500.diff
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/projecttacoma/bonnie/pull/500.

## 

Tim Taylor
ttaylor@mitre.org
Office: 781-271-2099
Mobile: 617-893-0107

Principal Software Systems Engineer
The MITRE Corporation
202 Burlington Road
Bedford, MA 01730-1420

pkmitre:
Yes to Tim's comment: the TGT should be stored in the user session, it shouldn't be exposed on the client side.

app/assets/javascripts/views/import_measure_view.js.coffee
```diff
@@ -70,6 +94,8 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
     @importDialog = @$("#importMeasureDialog")
     @importWait = @$("#pleaseWaitDialog")
     @finalizeDialog = @$("#finalizeMeasureDialog")
+    $('#clearVSACCreds').on 'click' , =>
```
pkmitre: Thorax has a streamlined way to access click handlers, look at the events: hash initialized near the top of the Thorax.Views.ImportMeasure class

app/assets/javascripts/views/import_measure_view.js.coffee
```diff
@@ -55,9 +55,33 @@ class Thorax.Views.ImportMeasure extends Thorax.Views.BonnieView
...
+      success: (data, textStatus, jqXHR) ->
+        if data? && data.valid
+          $('#vsacSignIn').addClass('hidden')
+          $('#vsacCachedMsg').removeClass('hidden')
+          $('#loadButton').prop('disabled', false) 
+          # If the measure import window is open long enough for the VSAC
+          # credentials to expire, we need to reshow the username and
+          # password dialog.
+          setTimeout -> 
+            $("#clearVSACCreds").click();
```
pkmitre: Instead of clicking a button and going through the DOM, can we just call the click handler?

app/controllers/measures_controller.rb
```diff
@@ -202,6 +202,50 @@ def create
...
+        params[:vsac_password], 
+        APP_CONFIG['nlm']['ticket_url']
+      )
+      # Create a new ticket granting ticket session variable that expires 
+      # 7.5hrs from now
+      if !tgt.nil? && !tgt.empty?
+        ticket = {'ticket' => tgt, 
+          'expires' => Time.now + 27000, 
+          'username' => params[:vsac_username]}
+        session[:tgt] = ticket.to_json
```
pkmitre: We don't need to JSON-ify stuff to put it in the rails session

app/controllers/measures_controller.rb
```diff
@@ -202,6 +202,50 @@ def create
     redirect_to "#{root_path}##{params[:redirect_route]}"
   end
 
+  def get_ticket_granting_ticket
```
pkmitre: We should make this private to make it clear that it's not intended to be called as an action.

app/controllers/measures_controller.rb
```diff
@@ -202,6 +202,50 @@ def create
...
+      # Retrieve a new ticket granting ticket
+      tgt = HealthDataStandards::Util::VSApi.get_tgt_using_credentials(
+        params[:vsac_username], 
+        params[:vsac_password], 
+        APP_CONFIG['nlm']['ticket_url']
+      )
+      # Create a new ticket granting ticket session variable that expires 
+      # 7.5hrs from now
+      if !tgt.nil? && !tgt.empty?
+        ticket = {'ticket' => tgt,
```
pkmitre: As a style thing we prefer the newer ruby hash syntax when possible, plus also using symbols as keys, ie

```
{ ticket: tgt, expires: Time.now + 27000, username: params[:vsac_username] }
```

PR #501, opened 2015-12-31
Add pagination to the value sets code listing

thecristen:
Includes major refactoring of the value sets listing.

This pagination implementation is handled by a backbone plugin [(backbone.paginator)](https://github.com/backbone-paginator/backbone.paginator) and a jquery plugin [(twbs-pagination)](https://github.com/esimakin/twbs-pagination).

iygw621u9y


edeyoung:
I've seen some pagination things where you can see the end index page. It looks something like this:

(within first few pages)
1, 2, 3, 4, ...., 18

(within middle pages)
1, ...., 8, 9, 10, .... 18

(within last few pages)
1, ...., 15, 16, 17, 18

Would it be possible to do something like that? I find that's easier to track where I am in the pagination scheme.

notpace:
Is there a reason why the pagination is on top of the code list?  I'm used to seeing it on the bottom of most search results.

edeyoung:
In the refactoring, I noticed a lot of the `for...in...` loops got replaced with `_.each()`. Is this because underscore adds additional functionality to its looping?
Was curious about speed differences between the two and found this: http://jsperf.com/coffeescript-for-vs-underscore-each-vs-jquery-each/26
According to this, looks like the coffeescript looping is faster on chrome at least.

Which should be used moving forward?

pkmitre:
When otherwise relatively equivalent I tend to prefer language-supported features rather than library features. In this case, I do tend to prefer for...in... as part of coffeescript vs _.each, which relies on the underscore library.

A somewhat related case is when using backbone collections, which support a built in each method; in that case, use the each built into backbone since the collection is already tied to backbone and it wouldn't mean bringing in another library since backbone is already used.

thecristen:
Good points; I think the adjustments in my last commit probably make the most sense with respect to which type of looping to choose.

To address @notpace's comment: Each page of results does not have the same height, so bottom-pagination jumps up and down from page to page. Placing the pagination at the top was an easy way of making it easier to click "Next" repeatedly without having to adjust mouse position! 

And as to the type of pagination @edeyoung refers to, there seems to be no turnkey way of doing this with the plugin we're using, but I think it's a fine idea and we should feel free to consider it for a future task.

app/assets/javascripts/helpers.js.coffee
```diff
@@ -41,3 +41,20 @@ Handlebars.registerHelper 'ifCond', (v1, operator, v2, options) ->
...
+  if n
+    i = 0
+    while i < n
+      data.index = i + 1
+      out += opts.fn(this, data: data)
+      i += 1
+  else
+    out = opts.inverse(this)
+
+  return out
```
pkmitre, line 55: This feels more like JavaScript than CoffeeScript (ie variables don't need to be pre-declared in CoffeeScript); how about something more like

```
Handlebars.registerHelper 'times', (n, opts) ->
  if n
    out = ''    
    for index in [1..n]
      out += opts.fn(this, data: { index: index })
  else
    out = opts.inverse(this)

  return out
```

(haven't actually run it)

pkmitre:
Really impressed by having an animated image to show it off in the pull request!

app/assets/javascripts/models/value_sets.js.coffee
```diff
@@ -1,5 +1,12 @@
 class Thorax.Collections.ValueSetsCollection extends Thorax.Collection
-  comparator: (vs) -> vs.get('display_name').toLowerCase()
+  initialize: (_, options) ->
+    @sorting = options.sorting || 'default'
```
pkmitre: If one were to ever initialize the collection on instantiation (ie new Thorax.Collections.ValueSetsCollection([vs1, vs2, vs3]), the value sets would just get thrown away, I think we could just do something like

```
class Thorax.Collections.ValueSetsCollection extends Thorax.Collection
  initialize: (valueSets, options) ->
    super(valueSets)
    @sorting = options.sorting || 'default'
```

(not tested)

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,107 @@
...
+            if hasOverlap then matchedCodes.push(code1)
+
+          if matchedCodes.length > 0
+            @overlappingValueSets.add
+              cid: vs1.cid + "_" + vs2.cid
+              codes: new Backbone.PageableCollection @sortAndFilterCodes(matchedCodes), @pagination_options
+              oid1: vs1.oid
+              name1: vs1.name
+              oid2: vs2.oid
+              name2: vs2.name
```
pkmitre: I think I may have liked the previous approach of showing each overlap twice, with each OID listed once on the left and once on the right... lets discuss!

app/assets/javascripts/models/value_sets.js.coffee
thecristen: After playing around with it it doesn't seem like the models get thrown away in this case! I did update the function anyway (mimicking the pattern seen [in the Result collection](https://github.com/projecttacoma/bonnie/blob/master/app/assets/javascripts/models/result.js.coffee#L205))

app/assets/javascripts/helpers.js.coffee
thecristen, line 55: Done.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
         oid = sdc.get('code_list_id')
-        valueSetName = sdc.get('title')
+        cid = sdc.cid
+
         if bonnie.valueSetsByOid[oid]?
           version = bonnie.valueSetsByOid[oid].version
-          code_concepts = @sortAndFilterCodes(bonnie.valueSetsByOid[oid].concepts)
+          # if it's a date (rather than "Draft"), format the version
+          version = moment(version, "YYYYMMDD").format("MM/DD/YYYY") if moment(version, "YYYYMMDD").isValid()
+          code_concepts = bonnie.valueSetsByOid[oid].concepts
```
edeyoung: i think these are hold overs from the previous version, but they should be changed to be consistent with other variable names (`code_concepts` vs `codeConcepts`)

Same for all the other variables.

app/assets/javascripts/templates/measure/codes_items.hbs
```diff
@@ -0,0 +1,13 @@
+<td class="code">{{code_system_name}}</td>
```
edeyoung: does it make sense to put all the `.hbs` files associated with values sets into `/measure/value_sets/` or something similar? I think that would help give the files context for what they represent.

app/assets/javascripts/views/measure_value_sets_codes_list.js.coffee
```diff
@@ -0,0 +1,33 @@
...
+        onPageClick: (event, page) => @collection.getPage(page)
+
+  itemView: (item) => new Thorax.Views.MeasureCodes(model: item.model)
+
+  initialize: ->
+    # when a new page is accessed the collection resets. when this happens,
+    # we re-render the view so the display shows updated selected page etc.
+    @listenTo @collection, 'reset', -> @render()
+
+class Thorax.Views.MeasureCodes extends Thorax.Views.BonnieView
```
edeyoung: What about naming to `MeasureValueSetCodes` or something like that? `MeasureCodes` makes it sound like you would get all the measure codes, but really, you'd just get the ones associated with a particular value list. Or maybe just `Codes` so it doesn't specify it either way? Or `ValueSetCodes`?

app/assets/javascripts/views/measure_value_sets_codes_list.js.coffee
```diff
@@ -0,0 +1,33 @@
+class Thorax.Views.PagedCodesList extends Thorax.CollectionView
```
edeyoung: same comment here as below. Do we want to rename this to something like `MeasureValueSetPagedCodesList`. Ick. That's a mouthful. I think renaming the class below (`MeasureCodeList`) is more important, but it might be worth thinking about renaming this one too.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
 class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
-
   template: JST['measure/value_sets']
 
   initialize: ->
-    @getValueSets()
-    @findOverlappingValueSets()
+    @summaryValueSets = [] # array of {generic value set descriptor, oid, and code}
+    @dataCriteria = new Thorax.Collections.ValueSetsCollection [], sorting: 'complex'  # all criteria that aren't supplemental criteria
```
edeyoung: I know this is a subjective thing, but I really like to have parens with function calls. It makes it easier for me to parse the line. Consider adding.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
+    @supplementalCriteria = new Thorax.Collections.ValueSetsCollection [], sorting: 'complex'  # ethnicity/gender/payer/race criteria
+    @overlappingValueSets = new Thorax.Collections.ValueSetsCollection [] # all value sets that overlap
+    @overlappingValueSets.comparator = (vs) -> [vs.get('name1'), vs.get('oid1')]
+
+    # options passed to a Backbone.PageableCollection instance
+    @pagination_options =
```
edeyoung: would we ever want to reconfigure this? is this the best location for this? i'm going to not very helpfully say it feels like it should be somewhere else, but I don't know where.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
-    supplementalCriteriaProperties = ["ethnicity", "gender", "payer", "race"]
+    supplementalCriteria = []
+    dataCriteria = []
 
-    dataCriteria = [] # all criteria that aren't supplemental criteria
-    supplementalCriteria = [] # ethnicity/gender/payer/race criteria
-    summaryValueSets = [] # array of {generic value set descriptor, oid, and code}
-
-    for sdc in @model.get('source_data_criteria').models
+    @model.get('source_data_criteria').each (sdc) =>
```
edeyoung, line 31: I think @pkmitre mentioned sometime in the distant past about using `for` instead of `.each`. consider changing that here.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
-        valueSetName = sdc.get('title')
+        cid = sdc.cid
+
         if bonnie.valueSetsByOid[oid]?
           version = bonnie.valueSetsByOid[oid].version
-          code_concepts = @sortAndFilterCodes(bonnie.valueSetsByOid[oid].concepts)
+          # if it's a date (rather than "Draft"), format the version
+          version = moment(version, "YYYYMMDD").format("MM/DD/YYYY") if moment(version, "YYYYMMDD").isValid()
+          code_concepts = bonnie.valueSetsByOid[oid].concepts
+          _(code_concepts).each (code) ->
```
edeyoung: same as above: consider changing `.each` to `for`

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
-        # only add value set info summaryValueSets if it isn't there already
-        # includes the common name for the value set, the oid, and the codes.
-        if _.where(summaryValueSets, {oid: oid}).length == 0
-          nameParts = valueSet.name.split(':')
-          if nameParts.length > 1
-            name = nameParts[1]
-          else
-            name = nameParts[0]
-          summaryValueSets.push({oid: oid, cid: cid, name: name, codes:code_concepts})
+        codes = new Backbone.PageableCollection @sortAndFilterCodes(code_concepts), @pagination_options
```
edeyoung: this is another place where lack of parens for a function call is confusing to me.
it doesn't look like coffeescript has a rule other than "readability" which isn't very helpful. Should we have some sort of rule for bonnie though? If so, what should it be? @pkmitre?

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
+        if _.where(@summaryValueSets, { oid: oid }).length == 0
+          nameParts = valueSet.name.split(':')
+          name = if nameParts.length > 1 then nameParts[1] else nameParts[0]
+          @summaryValueSets.push({ oid: oid, cid: cid, name: name, codes: codes })
 
-    @criteriaSets = criteriaSets
-    @summaryValueSets = summaryValueSets
+    # now that we have all the value sets, filter them
+    @supplementalCriteria.add @filterValueSets(supplementalCriteria)
+    @dataCriteria.add @filterValueSets(dataCriteria)
```
edeyoung: another parens or no parens spot (depending on what @pkmitre's decision is)

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
   findOverlappingValueSets: ->
-    overlappingValueSets = []
-    for valueSet1 in @summaryValueSets
-      for valueSet2 in @summaryValueSets
-        if valueSet1.oid == valueSet2.oid
-          continue
+    # determines if one or more codes in a value set are in another value set.
+    summaryValueSets = @filterValueSets(@summaryValueSets)
+
+    for vs1 in summaryValueSets
```
edeyoung: consider changing variable `vs1` to a more descriptive variable name. Sim for `vs2`.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
+      for vs2 in _(summaryValueSets).without(vs1)
         matchedCodes = []
-        for code1 in valueSet1.codes
-          for code2 in valueSet2.codes
-            if code1.code_system_name == code2.code_system_name && code1.code == code2.code
-              matchedCodes.push(code1)
+
+        vs1.codes.each (code1) =>
```
edeyoung: another `for` vs `.each` spot

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
+          hasOverlap = _.some vs2.codes.models, (code2) ->
```
edeyoung: do we need to use underscore for `some`? looks like it's included in javascript 1.6, but it's not totally clear to me what browsers support v1.6.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
+            overlapsCode = _(code2.get('code')).isEqual code1.get('code')
+            overlapsCodeSystem = _(code2.get('code_system_name')).isEqual code1.get('code_system_name')
```
edeyoung: Does it make sense to use the underscore `isEqual` here? Since we're comparing strings, `==` should work fine.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -1,118 +1,103 @@
...
-      @$('#' + description_id).animate
-        'max-height': @$('#' + description_id)[0].scrollHeight # expand
-      @$('#' + expand_id).html 'Show less <i class="fa fa-caret-up"></i>'
-    else
-      @$('#' + description_id).animate
-        'max-height': parseInt(@$('#' + description_id).css('line-height')) # contracts
-      @$('#' + expand_id).html 'Show more <i class="fa fa-caret-down"></i>'
+          @overlappingValueSets.add
+            cid: vs1.cid + "_" + vs2.cid
+            codes: new Backbone.PageableCollection @sortAndFilterCodes(matchedCodes), @pagination_options
```
edeyoung: two more parens vs no parens here.

app/assets/javascripts/views/measure_value_sets_codes_list.js.coffee
```diff
@@ -0,0 +1,33 @@
...
+        startPage: @collection.state.currentPage
+        initiateStartPageClick: false
+        onPageClick: (event, page) => @collection.getPage(page)
+
+  itemView: (item) => new Thorax.Views.MeasureCodes(model: item.model)
+
+  initialize: ->
+    # when a new page is accessed the collection resets. when this happens,
+    # we re-render the view so the display shows updated selected page etc.
+    @listenTo @collection, 'reset', -> @render()
```
edeyoung: this line is written in a confusing way for me (maybe the parens thing again, but this one is legit confusing to me). Consider reworking.

app/assets/javascripts/views/measure_value_sets.js.coffee
pkmitre, line 31: I think we should prefer `for` to the user of underscore each (`_.each`), but prefer the built in `each` interator provided by Thorax to `for`. Is that confusing enough?

app/assets/javascripts/views/measure_value_sets.js.coffee
pkmitre: This one is an underscore usage of `each`

app/assets/javascripts/views/measure_value_sets.js.coffee
pkmitre: This might not be super well thought out, but: use parentheses 1) for constructors, I find the extra space after the `new` makes mental parsing more difficult 2) for any nested function calls 3) if there are multiple arguments and any of the arguments is complex, ie a) a multi-argument function or b) a deeply nested reference, where nesting can involve function calls, array references, or object references. So, OK:

```
funcCall arg1, arg2, arg3

funcCall nestedFunc1(nestedArg1)

funcCall nestedFunc1(nestedArg1), nestedFunc2(nestedArg2)

funcCall arrayRef[index], nestedFunc1(nestefArg1)

new ObjectClass(arg1, arg2)

funcCall nestedFunc1(arrayRef[index])
```

Not OK:

```
new ObjectClass arg1, arg2

funcCall nestedFunc1 nestedArg1

funcCall nestedFunc1(arrayRef[index], nestedArg2), arrayRef[nestedFunc2(nestedArg2)]
```

Thoughts?

app/assets/javascripts/views/measure_value_sets.js.coffee
pkmitre: http://caniuse.com/ is a great reference

app/assets/javascripts/views/measure_value_sets.js.coffee
edeyoung: I personally like parentheses always (I'm stuck in my old-fashioned ways, I guess), but I think what you laid out makes sense.

app/assets/javascripts/views/measure_value_sets.js.coffee
edeyoung: it looks like `some` is well supported: http://kangax.github.io/compat-table/es5/#test-Array.prototype.some

app/assets/javascripts/views/measure_value_sets_codes_list.js.coffee
thecristen: Are you referring to line 22? There are tons of `@listenTo` statements written in the same fashion throughout Bonnie's views.

app/assets/javascripts/views/measure_value_sets.js.coffee
thecristen: Since I reuse `@pagination_options` in two different places, I do think this is the most appropriate location

app/assets/javascripts/views/measure_value_sets.js.coffee
thecristen: This one's not a regular array

app/assets/javascripts/views/measure_value_sets_codes_list.js.coffee
edeyoung: Yeah, I'd leave it then. Pete explained the function passing structure to me for these types of calls (which I didn't understand well previously). This line makes much more sense to me now.

pkmitre: Though... shouldn't that be a fat arrow so that @render resolves correctly?

thecristen: I actually don't need the arrow so I just removed it...

pkmitre: Actually without an arrow @render will just be called during initialize, not when the collection is reset, which is definitely not what we want.

PR #503, opened 2016-01-05
Updated to using the VSAC V2 API and the RetrieveMultipleValueSets call.

hossenlopp:
- This fixes the effective date functionality
- This also updates the gemfile.lock for bonnie_bundler

### DIFF
--- a/lib/tasks/vsac.rake
+++ b/lib/tasks/vsac.rake
@@ -9,15 +9,18 @@ namespace :bonnie do
       effective_date = Date.parse(ENV['DATE']).strftime('%Y%m%d') if ENV['DATE']
       include_draft = false
       config = APP_CONFIG["nlm"]
-      api = HealthDataStandards::Util::VSApi.new(config["ticket_url"], config["api_url"], ENV['USERNAME'], ENV['PASSWORD'])
-      result = api.get_valueset(ENV['OID'], effective_date, include_draft)
-      # This would be for V2 API:
-      # config['api_url'] = 'https://vsac.nlm.nih.gov/vsac/svs/RetrieveValueSet'
-      # api = HealthDataStandards::Util::VSApiV2.new(config["ticket_url"], config["api_url"], ENV['USERNAME'], ENV['PASSWORD'])
-      # result = api.get_valueset(ENV['OID'], effective_date: effective_date)
+      
+      # This would be for V1 API:
+      #config['api_url'] = 'https://vsac.nlm.nih.gov/vsac/ws/RetrieveValueSet'
+      #api = HealthDataStandards::Util::VSApi.new(config["ticket_url"], config["api_url"], ENV['USERNAME'], ENV['PASSWORD'])
+      #result = api.get_valueset(ENV['OID'], effective_date, include_draft)
+      
+      # We are using V2 API:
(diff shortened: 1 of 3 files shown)


lib/tasks/vsac.rake
```diff
@@ -9,15 +9,18 @@ namespace :bonnie do
...
       include_draft = false
       config = APP_CONFIG["nlm"]
-      api = HealthDataStandards::Util::VSApi.new(config["ticket_url"], config["api_url"], ENV['USERNAME'], ENV['PASSWORD'])
-      result = api.get_valueset(ENV['OID'], effective_date, include_draft)
-      # This would be for V2 API:
-      # config['api_url'] = 'https://vsac.nlm.nih.gov/vsac/svs/RetrieveValueSet'
-      # api = HealthDataStandards::Util::VSApiV2.new(config["ticket_url"], config["api_url"], ENV['USERNAME'], ENV['PASSWORD'])
-      # result = api.get_valueset(ENV['OID'], effective_date: effective_date)
+      
+      # This would be for V1 API:
```
edeyoung: I'm confused about this comment. Why did it say V2 earlier and now say V1? Can we just remove this commented code?

NEVERMIND! :) read it all more carefully.
(I guess the question still stands regarding if we can/should just remove this block of code)

hossenlopp: Now that you bring that up, I think it can be removed. This rake task had the V2 operations commented out before. It was helpful to switch to the V2 code when I was first figuring out what was going on and how to solve it. I don't think we'll ever need to go back to V1 so I guess having the block there to switch to V1 isn't helpful. Someone could just look at the git history if they needed too. I'll remove it.

pkmitre:
When loading an XML measure in Bonnie, and using the default setting of "Draft" for value sets, I get the following error:

> VSAC value sets could not be loaded.
> 
> Please verify that you are using the correct VSAC username and password. Error Loading Value Sets from VSAC: Error Loading Value Sets from VSAC: No ValueSet found for oid '2.16.840.1.113762.1.4.1'

This OID is for a value set used in most measures: AdministrativeGender, so this issue will likely impact all measures.

While I can get the measure to load by selecting "Published'' instead of "Draft", this won't work as default behavior. We need to figure out what's going on when "Draft" is selected and figure out what we should do.

hossenlopp:
It turns out when using the 'includeDraft' parameter you need to also use the 'profile' parameter. Theres a utility call in the VSAC V2 API that lists these:

``` XML
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<ProfileList>
    <profile>Most Recent CS Versions</profile>
    <profile>MU2 Update 2015-05-01</profile>
    <profile>MU2 EP Update 2014-05-30</profile>
    <profile>MU2 EH Update 2014-04-01</profile>
    <profile>MU2 EP Update 2013-06-14</profile>
    <profile>MU2 EH Update 2013-04-01</profile>
    <profile>MU2 Update 2012-12-21</profile>
    <profile>MU2 Update 2012-10-25</profile>
</ProfileList>
```

Hard coding the use of 'Most Recent CS Versions' makes 'includeDraft' work. This change would have to be made in https://github.com/projectcypress/health-data-standards.

hossenlopp:
Pull request created at projectcypress/health-data-standards#343 to fix the include_draft option to also set the profile parameter to the default or a provided profile.

PR #504, opened 2016-01-05
Changed VSAC username password fields to use both keyup and change events

hossenlopp:
- Enables the load button for users that copy paste username and password in or use a browser autofill


edeyoung:
Did this get tested on IE?

hossenlopp:
I just tested it in IE. I couldn't get IE's autofill to kick in but copy and paste worked fine.

PR #505, opened 2016-01-06
Bonnie 104932540 update value sets final

alexanderelliott121:


app/controllers/measures_controller.rb
```diff
@@ -16,16 +16,16 @@ def show
...
 
       # Not the cleanest code, but we get a many second performance improvement by going directly to Moped
       # (The two commented lines are functionally equivalent to the following three uncommented lines, if slower)
       # value_sets_by_oid = HealthDataStandards::SVS::ValueSet.in(oid: value_set_oids).index_by(&:oid)
       # @value_sets_by_oid_json = MultiJson.encode(value_sets_by_oid.as_json(except: [:_id, :code_system, :code_system_version]))
-      value_sets = Mongoid::Sessions.default[HealthDataStandards::SVS::ValueSet.collection_name].find(oid: { '$in' => value_set_oids }, user_id: current_user.id)
+      value_sets = Mongoid::Sessions.default[HealthDataStandards::SVS::ValueSet.collection_name].find(bonnie_hash: { '$in' => bonnie_hashes })
       value_sets = value_sets.select('concepts.code_system' => 0, 'concepts.code_system_version' => 0)
-      @value_sets_by_oid_json = MultiJson.encode value_sets.index_by { |vs| vs['oid'] }
+      @value_sets_by_oid_json = MultiJson.encode value_sets.index_by { |vs| vs['bonnie_hash'] }
```
pkmitre: It makes sense to update this variable name to not be @value_sets_by_oid_json since the value sets aren't indexed by oid any more

app/controllers/measures_controller.rb
```diff
@@ -16,16 +16,16 @@ def show
   end
 
   def value_sets
-    if stale? last_modified: Measure.by_user(current_user).max(:updated_at).try(:utc)
-      value_set_oids = Measure.by_user(current_user).only(:value_set_oids).pluck(:value_set_oids).flatten.uniq
+   if stale? last_modified: Measure.by_user(current_user).max(:updated_at).try(:utc)
+      bonnie_hashes = Measure.by_user(current_user).only(:bonnie_hashes).pluck(:bonnie_hashes).uniq.flatten
```
pkmitre: Looks like you swapped the order of uniq and flatten, which I think is going to cause incorrect behavior here

lib/tasks/measure_migration.rake
```diff
@@ -0,0 +1,98 @@
+namespace :bonnie do
+  namespace :measures do
+
+	  desc "Delete Value Set indexes that are not needed in Bonnie"
```
pkmitre: Indenting is off here

test/functional/measures_controller_test.rb
```diff
@@ -73,15 +73,11 @@ class << measure_file
...
       attr_reader :tempfile
     end
-
     post :create, {measure_file: measure_file, measure_type: 'eh', calculation_type: 'episode'}
     assert_response :redirect
     measure = Measure.where({hqmf_id: "40280381-3D27-5493-013D-4DCA4B826AE4"}).first
-
     assert_equal "42BF391F-38A3-4C0F-9ECE-DCD47E9609D9", measure.hqmf_set_id
     assert_equal 29, measure.value_sets.count
-    assert_equal @user.id, measure.user_id
```
pkmitre, line 83: I can see deleting the value set user_id test in the line below, but why the measure user_id test?

test/functional/valuesets_controller_test.rb
```diff
@@ -2,20 +2,21 @@
...
     dump_database
     collection_fixtures("users", "draft_measures")
     @user = User.by_email('bonnie@example.com').first
-    
+
     associate_user_with_measures(@user,Measure.all)
 
     @user.measures.first.value_set_oids.uniq.each_with_index do |oid|
       vs = HealthDataStandards::SVS::ValueSet.new(oid: oid)
+      vs.version = oid
```
pkmitre: Why set the version to the oid? Is it just to have something there? If so, it would be a little less opaque to say something like `vs.version = 'DRAFT'`

lib/tasks/measure_migration.rake
```diff
@@ -0,0 +1,105 @@
...
+  namespace :measures do
+    desc "Delete Value Set indexes that are not needed in Bonnie"
+    task :delete_unnecessary_value_set_indexes => :environment do
+      HealthDataStandards::SVS::ValueSet.collection.indexes.drop({"concepts.code"=>1})
+      HealthDataStandards::SVS::ValueSet.collection.indexes.drop({"concepts.code_system"=>1})
+      HealthDataStandards::SVS::ValueSet.collection.indexes.drop({"concepts.code_system_name"=>1})
+      HealthDataStandards::SVS::ValueSet.collection.indexes.drop({"concepts.display_name"=>1})
+    end
+
+    desc "Delete Value Set indexes that are not needed in Bonnie"
```
alexanderelliott121, line 4: Remove duplicate method.

PR #507, opened 2016-01-12
Undo limiting number of results to 1

edeyoung:
There are several criteria that can have multiple results. This gets rid of the requirement that you can only enter a single result (originally in commit https://github.com/projecttacoma/bonnie/commit/51fa175c2b96f11bd161a1c786c9136535b63c71).


edeyoung:
Adjustment to just limit only having one result to risk category assessment actually trivial so not doing reversion first, just doing limit.

PR #509, opened 2016-01-14
Fixed coffescript code that changes focus between patient-builder elements

hossenlopp:
This change reflects a change to the order in the templates that was not updated in this code.


hossenlopp:
Decided to just remove this code. Does that event that does @materialize() still need to occur?

pkmitre:
Yes, we definitely need to re-materialize the patient if any of the fields change.

pkmitre:
FYI I did a tiny code re-format and squashed the commits, and will merge as soon as the tests pass again

PR #510, opened 2016-01-27
Excel export specific rationale

jbradl11:
Passing specific rationale to the backend when exporting to excel and changing the TRUE, FALSE values based on the specific rationale.

### DIFF
--- a/app/assets/javascripts/models/result.js.coffee
+++ b/app/assets/javascripts/models/result.js.coffee
@@ -200,6 +200,11 @@ class Thorax.Models.Result extends Thorax.Model
     index = hqmf.SpecificsManager.indexLookup[@measure.get('data_criteria')[dataCriteriaKey].source_data_criteria]
     goodElements = (row[index] for row in @get('finalSpecifics')[populationCriteriaKey]) if index? and @get('finalSpecifics')?[populationCriteriaKey]?
 
+
+  toJSON: ->
+    _(super).extend({specificsRationale: @specificsRationale()}) 
+
+
 class Thorax.Collections.Results extends Thorax.Collection
   model: Thorax.Models.Result
   initialize: (models, options) -> @parent = options?.parent
--- a/app/assets/javascripts/models/result.js.coffee
+++ b/app/assets/javascripts/models/result.js.coffee
@@ -200,7 +200,7 @@ class Thorax.Models.Result extends Thorax.Model
     index = hqmf.SpecificsManager.indexLookup[@measure.get('data_criteria')[dataCriteriaKey].source_data_criteria]
     goodElements = (row[index] for row in @get('finalSpecifics')[populationCriteriaKey]) if index? and @get('finalSpecifics')?[populationCriteriaKey]?
 
-
+  # adds specific rationale results to the toJSON output.
   toJSON: ->
     _(super).extend({specificsRationale: @specificsRationale()}) 
 
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -98,6 +98,7 @@ class PatientsController < ApplicationController
   end
 
   def excel_export
+
     # Grab all records for the given measure
     measure = Measure.where(hqmf_set_id: params[:hqmf_set_id], user_id: current_user.id).first
     records = Record.by_user(current_user).where({:measure_ids.in => [params[:hqmf_set_id]]})   
@@ -105,7 +106,7 @@ class PatientsController < ApplicationController
     # Only generate excel document if there are patients for the given measure.
     if records.length > 0
       cookies[:fileDownload] = "true" # We need to set this cookie for jquery.fileDownload
-      package = PatientExport.export_excel_file(measure, records)
+      package = PatientExport.export_excel_file(measure, records, params['results'])
       send_data package.to_stream.read, type: "application/xlsx", filename: "#{measure.cms_id}.xlsx"
     else
       render nothing: true
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -106,7 +106,7 @@ class PatientsController < ApplicationController
     # Only generate excel document if there are patients for the given measure.
     if records.length > 0
       cookies[:fileDownload] = "true" # We need to set this cookie for jquery.fileDownload
-      package = PatientExport.export_excel_file(measure, records, params['results'])
+      package = PatientExport.export_excel_file(measure, records, params[:results])
       send_data package.to_stream.read, type: "application/xlsx", filename: "#{measure.cms_id}.xlsx"
     else
       render nothing: true
--- a/lib/measure_exported_results.rb
+++ b/lib/measure_exported_results.rb
@@ -25,16 +25,25 @@ class MeasureExportedResults
 
   def get_criteria_value(criteria_key, population_type)
     value = @patient['rationale'][criteria_key]
-    #Change value if specific rationale is involved.
-    if @patient.key?('specificsRationale')
-      if @patient['specificsRationale'].key?(population_type)
-        value = @patient['specificsRationale'][population_type][criteria_key]
+    #value could be true, false, or nil.
+    if value != nil && value != "false"
+      value = "TRUE"
+    elsif value == "false"
+      value = "FALSE"
+    end
+
+    #Change value if specific rationale is involved. 
+    if @patient['specificsRationale'] && @patient['specificsRationale'][population_type] 
+      value = @patient['specificsRationale'][population_type][criteria_key]
+      
+      #value could be "false", nil, "true"
+      if value = "false"
+        value = "SPECIFICALLY FALSE"
+      elsif value = "true"
+        value = "SPECIFICALLY TRUE"
       end
     end
-    if value == "false"
-      return false
-    else
-      return value
-    end
+
+    return value
   end
 end
--- a/lib/measure_exported_results.rb
+++ b/lib/measure_exported_results.rb
@@ -7,15 +7,15 @@ class MeasureExportedResults
 
   def match_patient_to_patient_id(patient_id, population_index)
     patients = @results[population_index.to_s]
-    #Iterate over each of the patients to match the patient_id
-    patients.map{ |key, value| value if value['patient_id'] == patient_id }.compact.try(:first)
+    # Iterate over each of the patients to match the patient_id
+    patients.map{ |key, value| value if value[:patient_id] == patient_id }.compact.try(:first)
   end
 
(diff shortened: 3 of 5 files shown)


edeyoung:
**NEVER MIND! (I didn't restart the server)**

~~I tried running this using the simple specific occurrence example I drafted a while ago. I got the following error dialog:~~

image

~~The server showed:~~

```
ArgumentError (wrong number of arguments (3 for 2)):
  lib/patient_export.rb:16:in `export_excel_file'
  app/controllers/patients_controller.rb:109:in `excel_export'
```

~~The patients I created were:~~

P1
- encounter during MP
- encounter outside of MP with a diagnosis

P2
- encounter during MP with a diagnosis

P3
- encounter during MP

edeyoung:
In the export, it looks like something that is specifically false isn't differentiated from something that's false. What about adding a differentiator, like an asterisk ("**FALSE")?

lib/measure_exported_results.rb
```diff
@@ -0,0 +1,40 @@
...
+      if @patient['specificsRationale'].key?(population_type)
+        value = @patient['specificsRationale'][population_type][criteria_key]
+      end
+    end
+    if value == "false"
+      return false
+    else
+      return value
+    end
+  end
```
edeyoung, line 48: It seems like possible values here are: `false`, `nil`, `{"results"...}`, (other?).
It seems like four values would suffice: nil, false, true, specifically false.
1. Can you change the `specificsRationale` outcome to signify specifically false?
2. Can you make the returned value consistent somehow - symbols maybe? e.g. `:true`, `:false`, `:specifically_false`, `:none` (or as strings).

lib/patient_export.rb
```diff
@@ -211,14 +199,21 @@ def self.generate_rows(sheet, records, measure, population_index, expected_value
...
+          population_hash[index_key] = key
+          index_key = index_key + 1
         end
-        if value != false && value != nil && value != 'X'
+      end
+
+       # Populate the values of each row, in the order that the headers were generated.
+      population_criteria_keys.each_with_index do |key, index|
+        value = exported_results.get_criteria_value(key, population_hash[index])  
+        if value != false && value != nil
```
edeyoung: if all non-`false` and non-`nil` values will be `true`, consider moving the translation logic into `get_criteria_value` so you don't have to do it here and can just do a `patient_row.push(value)`.

edeyoung: Maybe pushing strings would work better than booleans here. If you push strings, you can note that something is specifically false. You can have `get_criteria_value` output the correct strings and just push that value onto here.

lib/patient_export.rb
```diff
@@ -211,14 +199,21 @@ def self.generate_rows(sheet, records, measure, population_index, expected_value
...
-        else
-          value = result['rationale'][key]
+      #Create a hash of population type and index of the list of criteria
+      population_hash = Hash.new
+      index_key = 0
+
+      criteria_keys_by_population.each do |key, list|
+        list.each do |value| 
+          population_hash[index_key] = key
+          index_key = index_key + 1
```
edeyoung: consider `index_key += 1`

lib/patient_export.rb
```diff
@@ -211,14 +199,21 @@ def self.generate_rows(sheet, records, measure, population_index, expected_value
...
       end
 
-      # Populate the values of each row, in the order that the headers were generated.
-      population_criteria_keys.each do |key|
-        if setup_exception || calculation_exception
-          value = 'X'
-        else
-          value = result['rationale'][key]
+      #Create a hash of population type and index of the list of criteria
+      population_hash = Hash.new
```
edeyoung: does this need to be a hash or can it be an array?

app/controllers/patients_controller.rb
```diff
@@ -105,7 +106,7 @@ def excel_export
     # Only generate excel document if there are patients for the given measure.
     if records.length > 0
       cookies[:fileDownload] = "true" # We need to set this cookie for jquery.fileDownload
-      package = PatientExport.export_excel_file(measure, records)
+      package = PatientExport.export_excel_file(measure, records, params['results'])
```
edeyoung: should this be `params[:results]`?

app/assets/javascripts/models/result.js.coffee
```diff
@@ -200,6 +200,11 @@ class Thorax.Models.Result extends Thorax.Model
     index = hqmf.SpecificsManager.indexLookup[@measure.get('data_criteria')[dataCriteriaKey].source_data_criteria]
     goodElements = (row[index] for row in @get('finalSpecifics')[populationCriteriaKey]) if index? and @get('finalSpecifics')?[populationCriteriaKey]?
 
+
```
edeyoung: consider adding a comment explaining the specifics rationale business below.

lib/patient_export.rb
```diff
@@ -160,19 +162,11 @@ def self.create_criteria_key_header_lookup(measure, criteria_keys_by_population)
...
-      calculator.set_measure_and_population(measure, population_index, rationale: true)
-    rescue => e
-      setup_exception = "Measure setup exception: #{e.message}"
-    end
-
+  def self.generate_rows(sheet, records, measure, population_index, expected_value_keys, population_criteria_keys, results, criteria_keys_by_population)
     # Populates the patient data
     records.each do |patient|
-
+      #Removes \" from beignning and end of the patient_id string
```
edeyoung: style guidance says to add a space after the `#` for comments

lib/patient_export.rb
```diff
@@ -160,19 +162,11 @@ def self.create_criteria_key_header_lookup(measure, criteria_keys_by_population)
...
-
+  def self.generate_rows(sheet, records, measure, population_index, expected_value_keys, population_criteria_keys, results, criteria_keys_by_population)
     # Populates the patient data
     records.each do |patient|
-
+      #Removes \" from beignning and end of the patient_id string 
+      exported_results = MeasureExportedResults.new(patient.id.to_json.tr('\"',''), population_index, results)
       patient_row = []
 
       expected_value_keys.each do |value|
```
edeyoung: for this loop, the variable is `value`. For the loop below, the variable is `key`. Consider making these consistent.

Also, consider renaming `expected_value_keys` to something like `population_categories` and then have the variable be something like `population_category` for both loops.

lib/patient_export.rb
```diff
@@ -160,19 +162,11 @@ def self.create_criteria_key_header_lookup(measure, criteria_keys_by_population)
...
-      setup_exception = "Measure setup exception: #{e.message}"
-    end
-
+  def self.generate_rows(sheet, records, measure, population_index, expected_value_keys, population_criteria_keys, results, criteria_keys_by_population)
     # Populates the patient data
     records.each do |patient|
-
+      #Removes \" from beignning and end of the patient_id string 
+      exported_results = MeasureExportedResults.new(patient.id.to_json.tr('\"',''), population_index, results)
       patient_row = []
```
edeyoung: Consider adding a comment for the loop below explaining that you're getting the expected values for each population.

lib/patient_export.rb
```diff
@@ -183,18 +177,12 @@ def self.generate_rows(sheet, records, measure, population_index, expected_value
         end
       end
 
-      # Generate the calculated rationale for each patient against the measure.
-      begin
-        result = calculator.calculate(patient)
-      rescue Exception => e
-        calculation_exception = "Measure calculation exception: #{e.message}"
-      end
-
```
edeyoung, line 192: Consider adding a comment for the loop below explaining that you're getting the actual value for each population.

lib/measure_exported_results.rb
```diff
@@ -0,0 +1,49 @@
...
+        return eval @patient['values'].to_s.gsub('"', '')
+      else
+        return 0
+      end
+    end
+  	@patient[population_type]
+  end
+
+  def get_criteria_value(criteria_key, population_type)
+    value = @patient['rationale'][criteria_key]
```
edeyoung: it looks like a best practice is to use symbols rather than strings for hash keys. Consider changing this to a symbol (`:rationale`). Similar for other keys in the changes.

lib/measure_exported_results.rb
```diff
@@ -0,0 +1,49 @@
...
+    patients = @results[population_index.to_s]
+    #Iterate over each of the patients to match the patient_id
+    patients.map{ |key, value| value if value['patient_id'] == patient_id }.compact.try(:first)
+  end
+
+  def value_for_population_type(population_type)
+    if population_type == 'OBSERV'
+      if @patient.key?('values') && @patient['rationale'].key?('OBSERV')
+        #Use eval to remove quotes around numbers ['75'] -> [75] 
+        return eval @patient['values'].to_s.gsub('"', '')
```
edeyoung: is `eval` the right way to go about this? what is the output of `@patient['values']`?

lib/measure_exported_results.rb
```diff
@@ -0,0 +1,49 @@
...
+    elsif value == "false"
+      value = "FALSE"
+    end
+
+    #Change value if specific rationale is involved. 
+    if @patient['specificsRationale'] && @patient['specificsRationale'][population_type] 
+      value = @patient['specificsRationale'][population_type][criteria_key]
+      
+      #value could be "false", nil, "true"
+      if value = "false"
```
edeyoung: change `=` to `==`

lib/measure_exported_results.rb
```diff
@@ -0,0 +1,49 @@
...
+        value = "SPECIFICALLY FALSE"
+      elsif value = "true"
```
edeyoung: change `=` to `==`

lib/measure_exported_results.rb
edeyoung: i don't think it's enough to see if a value is in the specifics. The original value also has to be the opposite (e.g. the original value is `true` and specifics say `false`). When the specifics rationale is created, it looks at a leaf clause that's specifically false, adds that to the specifics rationale, and then if appropriate adds parent clauses to the specifics rationale. Sometimes, however, the parent clause is false anyways. In this case, the parent clauses should still just be false rather than specifically false even if it's in the specifics rationale.

You can see this with the simple specifics example and the following patient:
- a single encounter during the measurement period

app/controllers/patients_controller.rb
pkmitre: params is a HashWithIndifferentAccess, so the symbol version is not strictly required, but the typical usage is to use the symbol version ie :results

lib/measure_exported_results.rb
pkmitre: Actually, using eval is a pretty major security issue, so we definitely need to find another way to do this

PR #511, opened 2016-02-06
Remove validation that prevents an event from ending after a patient expires

pkmitre:


edeyoung:
do we want a rule that something can't start after death? The rules that something can't end after death and something can't start after it ends implicitly covered this case before.

pkmitre:
I think that rule is already present:

```
if start_date && deathdate && start_date.isAfter deathdate
  errors.push [sdc.cid, 'start_date', "#{sdc.get('title')} start date must be before patient date of death"]
```

PR #516, opened 2016-02-12
Excel export changes

jbradl11:
Fix for BONNIE-140. Expired now is set to "FALSE" if the value is "nil". 

Also a small update to remove duplicate clauses in each population type.

### DIFF
--- a/lib/patient_export.rb
+++ b/lib/patient_export.rb
@@ -196,6 +196,8 @@ class PatientExport
         elsif value == 'birthdate' || value == 'deathdate'
           time = Time.at(patient[value]).strftime("%m/%d/%Y") unless patient[value].nil?
           patient_row.push(time)
+        elsif value == 'expired' && patient[value] == nil
+          patient_row.push(false)
         else
           patient_row.push(patient[value])
         end
--- a/lib/patient_export.rb
+++ b/lib/patient_export.rb
@@ -16,6 +16,12 @@ class PatientExport
   def self.export_excel_file(measure, records, results)
 
     criteria_keys_by_population = measure.criteria_keys_by_population
+
+    #Remove duplicates by population type
+    criteria_keys_by_population.each do | population_type, values | 
+      values.uniq!
+    end
+
     criteria_key_header_lookup = self.create_criteria_key_header_lookup(measure, criteria_keys_by_population)
 
     Axlsx::Package.new do |package|
--- a/lib/patient_export.rb
+++ b/lib/patient_export.rb
@@ -17,7 +17,7 @@ class PatientExport
 
     criteria_keys_by_population = measure.criteria_keys_by_population
 
-    #Remove duplicates by population type
+    # Remove duplicates by population type
     criteria_keys_by_population.each do | population_type, values | 
       values.uniq!
     end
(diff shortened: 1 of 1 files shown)


jbradl11:
Updated pull request with a Fix for bonnie-149 (Error grabbing expected values) and commented out code for removing duplicates from each population type.

lib/patient_export.rb
```diff
@@ -171,8 +178,11 @@ def self.generate_rows(sheet, records, measure, population_index, population_cat
 
       # populates the array with expected values for each population
       population_categories.each do |population_category|
-        if patient[:expected_values] && patient[:expected_values][population_index] && patient[:expected_values][population_index][population_category]
-          patient_row.push(patient[:expected_values][population_index][population_category])
+        # Filter out the expected values that match the measure hqmf_set_id. Return the first object in the array.
+        expected_values = patient[:expected_values].select{ | expected_values | expected_values[:measure_id] == measure.hqmf_set_id && 
+                                                                                expected_values[:population_index] == population_index.to_i }.compact.try(:first)
```
edeyoung: is `population_index` ever not an integer? is `to_i` necessary?

edeyoung: you have `try(:first)`, so it will return `nil` if the expected value set isn't found. There should be a `nil` check below.

edeyoung: `compact` is being used here, which looks like it removes `nil` values from an array. Would a `nil` value ever be added to this array?

lib/patient_export.rb
```diff
@@ -171,8 +178,11 @@ def self.generate_rows(sheet, records, measure, population_index, population_cat
 
       # populates the array with expected values for each population
       population_categories.each do |population_category|
-        if patient[:expected_values] && patient[:expected_values][population_index] && patient[:expected_values][population_index][population_category]
-          patient_row.push(patient[:expected_values][population_index][population_category])
+        # Filter out the expected values that match the measure hqmf_set_id. Return the first object in the array.
+        expected_values = patient[:expected_values].select{ | expected_values | expected_values[:measure_id] == measure.hqmf_set_id &&
```
edeyoung: I don't see any explicit style guidance, but all of the ruby code in the style guide does not have spaces between the `|`, so this would be: `{ |expected_values|`

lib/patient_export.rb
```diff
@@ -171,8 +178,11 @@ def self.generate_rows(sheet, records, measure, population_index, population_cat
 
       # populates the array with expected values for each population
       population_categories.each do |population_category|
-        if patient[:expected_values] && patient[:expected_values][population_index] && patient[:expected_values][population_index][population_category]
-          patient_row.push(patient[:expected_values][population_index][population_category])
+        # Filter out the expected values that match the measure hqmf_set_id. Return the first object in the array.
+        expected_values = patient[:expected_values].select{ | expected_values | expected_values[:measure_id] == measure.hqmf_set_id && 
+                                                                                expected_values[:population_index] == population_index.to_i }.compact.try(:first)
+        if expected_values[population_category]  
+          patient_row.push(expected_values[population_category])
```
edeyoung, line 186: there are some cute ways of doing these types of things so you can skip the `if/else`. I have mixed feelings about cute approaches, but ruby people seem to like them. Here's an example (untested):

```
patient_row.push(expected_values[population_category] || 0)
```

edeyoung, line 186: to handle the `nil` check mentioned above, you might even be able to do something like this (also untested):

```
patient_row.push(expected_values?[population_category] || 0)
```

pkmitre, line 186: The existential operator (?) is a coffeescript thing :->

PR #517, opened 2016-02-18
Adding a test case number field, per end user request

notpace:
screen shot 2016-02-18 at 4 18 44 pm

This came up in the last end user group session I sat in on.  Since measure developers generally use test case numbers (rather than patient names) in submissions for NQF endorsement, they would like to be able to create patients with test case numbers.

I left it as an optional field to avoid breaking any current patients in Bonnie, and I'm sure there will be situations where a test case number isn't necessary/wanted.

screen shot 2016-02-18 at 4 19 05 pm

Sorting will be a bit interesting, since it's currently done using an array of fail/pass status, first name, and last name.  Since test case number is optional, I'm adding "000" if it doesn't exist, since the measure developers seemed to be using 3-digit zero-padded numbers for their current test cases.

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
+++ b/app/assets/javascripts/templates/patient_builder/patient_builder.hbs
@@ -29,6 +29,11 @@
 <div class="row">
   <form class="col-left" role="form">
     <div class="form-group">
+      <label for="caseNumber" class="control-label">Test Case Number</label>
+      <input type="text" id="caseNumber" name="caseNumber" class="form-control" placeholder="#">
+    </div>
+    
+    <div class="form-group">
       <label for="last" class="control-label">Last Name</label>
       <input type="text" id="last" name="last" class="form-control" placeholder="Smith">
     </div>
--- a/app/assets/javascripts/templates/patient_bank/patient_bank.hbs
+++ b/app/assets/javascripts/templates/patient_bank/patient_bank.hbs
@@ -26,7 +26,7 @@
             <div class="patient">
               <div class="patient-checkbox">
                 <label>
-                  <span class="sr-only">Select {{patient.last}} {{patient.first}}</span>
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/templates/patient_builder/patient_builder.hbs
```diff
@@ -29,6 +29,11 @@
 <div class="row">
   <form class="col-left" role="form">
     <div class="form-group">
+      <label for="caseNumber" class="control-label">Test Case Number</label>
+      <input type="text" id="caseNumber" name="caseNumber" class="form-control" placeholder="#">
```
thecristen, line 33: Should probably be `type="number"` to enforce entry of number. Especially since I can imagine people typing a last name into this box from muscle memory...

notpace, line 33: I thought about doing that, but there are two problems with it:
1. The test cases are zero-padded (e.g., 001, 002) and `type="number"` might strip them out
2. One of the measure developers expressed a desire to place alpha characters after the number (e.g., if there were multiple patients to a given test case)

It will ultimately be up to the measure developers to enforce valid test case numbers, which is not great UX, but the sorting algorithm is not as flexible as I'd like (I thought about using an OID-like system for sorting, but that led me down a terrible rabbit hole) and I imagine the measure developers are happier leaving it alphanumeric.

thecristen, line 33: Fair points! OK, your decision makes even more sense now.
I'm a little concerned that its prominence as the very first field on the page will make people feel like they have to fill it out (right now we don't have a visible differentiation of required vs optional fields). That and it shifts the name fields further down, which will impact users who are used to filling out that section quickly. Can we consider moving the field to below the two name fields?
Otherwise I think this is great :+1:

notpace, line 33: Absolutely - I have no allegiance to the field order :)

notpace:
Hute has provided a series of ideas that are very likely to obviate the need for this pull request.

PR #519, opened 2016-02-23
fix for cached vsac draft vs. public slider

holmesie:


hossenlopp:
Looks like this works for reenabling the the VSAC options. But I have noticed since the VSAC ticket update that the VSAC specific errors are no longer showing. An error for MAT Export measures shows up instead. Here are two examples from entering the wrong VSAC password. 
From before:
screen shot 2016-02-23 at 10 15 01 am
Current:
screen shot 2016-02-23 at 10 16 19 am

hossenlopp:
Looks good now. Properly errors out when the VSAC creds are bad.

hossenlopp:
@pkmitre Should I merge this or do you want to?

pkmitre:
@hossenlopp I'll merge since I'm here, but in the future you can certainly merge once you're satisfied that it looks good (just making sure it's a fast forward merge)

PR #520, opened 2016-02-23
Adding required and/or aria-required to all required fields in Bonnie

notpace:
- `devise` doesn't automatically add required/aria-required indicators, so I added them there.  Also, email is now required for new users.
- On the measure upload modal, the only inputs that need some kind of input and are empty by default are the VSAC username and password.  Since they're persistent but hidden, I left out the required attribute and just used aria-required.
- For the patient builder, I added the required/aria-required attributes and an asterisk on first name, last name, and date of birth, since they are the only required fields.

### DIFF
--- a/app/assets/javascripts/templates/import/import_measure.hbs
+++ b/app/assets/javascripts/templates/import/import_measure.hbs
@@ -43,13 +43,13 @@
             <div class="form-group">
               <label for="vsacUser" class="col-sm-{{titleSize}} control-label">VSAC Username</label>
               <div class="col-sm-{{dataSize}}">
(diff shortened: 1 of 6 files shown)


pkmitre:
When editing a account attributes, this pull request requires the user to change their password, even if all they want to do is adjust their name.

pkmitre:
Looks like this pull request wound up with some un-related commits included (during a rebase?), I think we need to get that sorted out before looking at it again.

notpace:
I think this is fixed now - no unrelated commits and it should merge cleanly into master

PR #524, opened 2016-04-05
Fix for highlighting

mulcahyk:
- Added a new type of VAR to variable_logic_view.js.coffee.
- In highlightPatientData (population_logic_view.js.coffee) add logic to let VARiables bypass the test for being in the finalSpecifics.
- In data_criteria_logic_view.js.coffee added logic to check for Thorax.Views.VariablesLogic so that it will get the right type for VARiables.
- In data_criteria.hbs removed the if condition so that highlight-target gets added to derived elements.

### DIFF
--- a/app/assets/javascripts/views/logic/data_criteria_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/data_criteria_logic_view.js.coffee
@@ -64,8 +64,8 @@ class Thorax.Views.DataCriteriaLogic extends Thorax.Views.BonnieView
     populationView = @populationCriteriaView()
     return unless populationView
     populationCriteriaKey = populationView.population.type  # IPP, DEN, etc. including VAR(iables)
-    populationView.parent.highlightPatientData(dataCriteriaKey, populationCriteriaKey)  
-    
+    populationView.parent.highlightPatientData(dataCriteriaKey, populationCriteriaKey)
+
   clearHighlightEntry: (e) ->
     if view = @populationCriteriaView()
       view.parent.clearHighlightPatientData()
--- a/app/assets/javascripts/views/logic/data_criteria_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/data_criteria_logic_view.js.coffee
@@ -63,15 +63,15 @@ class Thorax.Views.DataCriteriaLogic extends Thorax.Views.BonnieView
     dataCriteriaKey = @dataCriteria.key
     populationView = @populationCriteriaView()
     return unless populationView
-    populationCriteriaKey = populationView.population.type
-    populationView.parent.highlightPatientData(dataCriteriaKey, populationCriteriaKey)
-
+    populationCriteriaKey = populationView.population.type  # IPP, DEN, etc. including VAR(iables)
+    populationView.parent.highlightPatientData(dataCriteriaKey, populationCriteriaKey)  
+    
   clearHighlightEntry: (e) ->
     if view = @populationCriteriaView()
       view.parent.clearHighlightPatientData()
 
   populationCriteriaView: ->
     parent = @parent
-    until !parent || parent instanceof Thorax.Views.PopulationCriteriaLogic
+    until !parent || parent instanceof Thorax.Views.PopulationCriteriaLogic || parent instanceof Thorax.Views.VariablesLogic
       parent = parent.parent
     parent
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -66,15 +67,17 @@ class Thorax.Views.PopulationLogic extends Thorax.Views.BonnieView
 
         targetrect.attr "class", (index, classNames) -> "#{classNames} #{targetClass}" #add styling to svg without removing all the other classes
 
-        targettext.addClass(targetClass)
+        targettext.addClass(targetClass)  #<<-- When this is executed the coloring is applied to the right on the right
```
edeyoung: the comment that was added here isn't clear to me. What do you mean by "applied to the right on the right"?

The comment styling we're using is a space after the pound with the text immediately following (e.g. `# When this is...`).

app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -66,15 +67,17 @@ class Thorax.Views.PopulationLogic extends Thorax.Views.BonnieView
...
         targettext.closest('.panel-heading').addClass(targetPanelClass)
         targettext.children('.sr-highlight-status').html(srTitle)
         # this second line is there to fix an issue with sr-only in Chrome making text in inline elements not display
         # by having the sr-only span and the DC title wrapped in a criteria-title span, the odd behavior goes away.
         targettext.children('.criteria-title').children('.sr-highlight-status').html(srTitle)
 
   highlightPatientData: (dataCriteriaKey, populationCriteriaKey) ->
-    if @latestResult?.get('finalSpecifics')?[populationCriteriaKey]
+    isVariablePopulation = populationCriteriaKey == 'VAR'
```
edeyoung: This is a situation where I like to use parens for readability.

Consider changing to `isVariablePopulation = (populatinoCriteriaKey == 'VAR')`

app/assets/javascripts/views/logic/variable_logic_view.js.coffee
```diff
@@ -6,3 +6,7 @@ class Thorax.Views.VariablesLogic extends Thorax.Views.BonnieView
   initialize: ->
     @variables = new Thorax.Collections.MeasureDataCriteria this.measure.get('source_data_criteria').select (dc) -> dc.get('variable') && !dc.get('specific_occurrence')
     @hasVariables = @variables.length > 0
+
+# Need to add this so that variables will have a type when it comes to handling the mouseover hightlighting
```
edeyoung: align the comment with the code

app/assets/javascripts/views/logic/population_logic_view.js.coffee
```diff
@@ -52,6 +52,7 @@ class Thorax.Views.PopulationLogic extends Thorax.Views.BonnieView
       @showRationaleForPopulation('variables', rationale, updatedRationale)
 
   showRationaleForPopulation: (code, rationale, updatedRationale) ->
+    #This is the logic that does the highlighting for the text on the right
```
edeyoung: see below - comment styling we use is to have a space after the `#`.

edeyoung: what do you mean by "highlighting for the text on the right"? Are you talking within the context of the measure view?
this chunk of code is related specifically to the logic part of that view. The logic part can be put in any page in any position so might not always be the "text on the right" (if I'm understanding what you meant here). Maybe rephrase to something like, "This is the code for managing the logic highlighting"

app/assets/javascripts/views/logic/data_criteria_logic_view.js.coffee
```diff
@@ -63,15 +63,15 @@ class Thorax.Views.DataCriteriaLogic extends Thorax.Views.BonnieView
     dataCriteriaKey = @dataCriteria.key
     populationView = @populationCriteriaView()
     return unless populationView
-    populationCriteriaKey = populationView.population.type
-    populationView.parent.highlightPatientData(dataCriteriaKey, populationCriteriaKey)
-
+    populationCriteriaKey = populationView.population.type  # IPP, DEN, etc. including VAR(iables)
+    populationView.parent.highlightPatientData(dataCriteriaKey, populationCriteriaKey)
```
edeyoung: Watch whitespace changes in your pushes to the master branch. When there are a lot of whitespace-only changes, it can get confusing reviewing the pull request.
Also, where possible, also try to avoid trailing whitespace. Using `git diff` before committing (which is a general good practice if you're not doing it already) actually highlights these for you, which is helpful.

edeyoung:
logic-wise it looks good. Have a few comments regarding styling and comments.

edeyoung:
@pkmitre what are your thoughts on adding test cases for testing highlighting? do we even have a test measure with variables in it? If not, we should definitely update that set.

pkmitre:
I like the idea of a test case for highlighting; my first instinct was that this would be tricky, but thinking about it more I think it's plausible. Though, I think I'd focus first on the basic logic highlighting, rather than the hover highlighting...

PR #527, opened 2016-04-20
Added principal diagnosis field value code selection for portfolio users

holmesie:
Allows customized code selections for Principal Diagnosis field values for portfolio users.

### DIFF
--- a/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
+++ b/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
@@ -71,7 +71,11 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
       values: @model.get('references')
       criteriaType: @model.get('type')
       vals: JSON.stringify(@model.get('references'))
-    @editCodeSelectionView = new Thorax.Views.CodeSelectionView criteria: @model
+    codes = @model.get('codes')
+    concepts = @model.valueSet().get('concepts')
+    codes.on 'add remove', => @model.set 'code_source', (if codes.isEmpty() then 'DEFAULT' else 'USER_DEFINED'), silent: true
+    @editCodeSelectionView = new Thorax.Views.CodeSelectionView codes: codes
+    @editCodeSelectionView.updateConcepts(concepts)
     @editFulfillmentHistoryView = new Thorax.Views.MedicationFulfillmentsView
       model: new Thorax.Model
       criteria: @model
@@ -216,9 +220,6 @@ class Thorax.Views.CodeSelectionView extends Thorax.Views.BuilderChildView
 
   initialize: ->
     @model = new Thorax.Model
-    @codes = @criteria.get('codes')
-    @codes.on 'add remove', => @criteria.set 'code_source', (if @codes.isEmpty() then 'DEFAULT' else 'USER_DEFINED'), silent: true
-    @codeSets = _(concept.code_system_name for concept in @criteria.valueSet()?.get('concepts') || []).uniq()
 
   validateForAddition: ->
     attributes = @serialize(set: false) # Gets copy of attributes from form without setting model
@@ -235,9 +236,14 @@ class Thorax.Views.CodeSelectionView extends Thorax.Views.BuilderChildView
     if codeSet isnt 'custom'
       blankEntry = if codeSet is '' then '--' else "Choose a #{codeSet} code"
       $codeList.append("<option value>#{blankEntry}</option>")
-      for concept in @criteria.valueSet().get('concepts') when concept.code_system_name is codeSet and !concept.black_list
+      for concept in @concepts when concept.code_system_name is codeSet and !concept.black_list
         $('<option>').attr('value', concept.code).text("#{concept.code} (#{concept.display_name})").appendTo $codeList
     @$('.codelist-control').focus()
+
+  updateConcepts: (concepts) ->
+    @concepts = concepts
+    @codeSets = _(concept.code_system_name for concept in @concepts || []).uniq()
+    @render()
 
   addCode: (e) ->
     e.preventDefault()
@@ -250,7 +256,7 @@ class Thorax.Views.CodeSelectionView extends Thorax.Views.BuilderChildView
     # Reset model to default values
     @model.clear()
     @$('select').val('')
-    # # Let the selectBoxIt() select box know that its value may have changed
+    # Let the selectBoxIt() select box know that its value may have changed
     @$('select[name=codeset]').change()
     @triggerMaterialize()
     @$(':focusable:visible:first').focus()
@@ -304,6 +310,7 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
     @model.clear()
     @triggerMaterialize()
 
+
 class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
   className: -> "#{if @fieldValue then 'field-' else ''}value-formset"
 
@@ -311,6 +318,9 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
 
   initialize: ->
     @model.set('type', 'CD')
+    @fieldValueCodesCollection = new Thorax.Collections.Codes {}, parse: true
+    @showAddCodesButton = false
+    @showAddCodes = false
 
   context: ->
     _(super).extend
@@ -328,21 +338,63 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
       delete attr.start_time
       title = @measure?.valueSets().findWhere(oid: attr.code_list_id)?.get('display_name')
       attr.title = title if title
+      attr.codes = @fieldValueCodesCollection.toJSON() unless jQuery.isEmptyObject(@fieldValueCodesCollection.toJSON())
     rendered: ->
+      @codeSelectionViewForFieldValues = new Thorax.Views.CodeSelectionView codes: @fieldValueCodesCollection
       @$("select[name=type]").selectBoxIt('native': true)
       @$('.date-picker').datepicker().on 'changeDate', _.bind(@validateForAddition, this)
       @$('.time-picker').timepicker(template: false).on 'changeTime.timepicker', _.bind(@validateForAddition, this)
     'change select[name=type]': (e) ->
       @model.set type: $(e.target).val()
+      @toggleAddCodesButton()
       @validateForAddition()
       @advanceFocusToInput()
+    'change select[name=code_list_id]': ->
+      @toggleAddCodesButton()
+      @validateForAddition()
     'change select': ->
+      @toggleAddCodesButton()
       @validateForAddition()
       @advanceFocusToInput()
     'keyup input': 'validateForAddition'
     'change select[name=key]': 'changeFieldValueKey'
     # hide date-picker if it's still visible and focus is not on a .date-picker input (occurs with JAWS SR arrow-key navigation)
     'focus .form-control': (e) -> if not @$(e.target).hasClass('date-picker') and $('.datepicker').is(':visible') then @$('.date-picker').datepicker('hide')
+    'click #addCodes': (e) ->
+      @showFieldValueCodeSelection(e)
+      @validateForAddition()
+
+  canSelectFieldValueCode: (concepts, key) ->
+    return bonnie.isPortfolio and @fieldValue and key in ['PRINCIPAL_DIAGNOSIS', 'DIAGNOSIS'] and concepts and @$("select[name=type]").val() == "CD"
+
+  getConcepts: (code_list_id) ->
+    return @measure?.valueSets().findWhere(oid: code_list_id)?.get('concepts')
+
+  toggleAddCodesButton: ->
+    attributes = @serialize(set: false)
+    if @canSelectFieldValueCode(@getConcepts(attributes.code_list_id), attributes.key)
+      # Show code selection for field value
+      @showAddCodesButton = true
+      @fieldValueCodesCollection.reset()
+    else
+      @showAddCodesButton = false
+    @showAddCodes = false
+    @render()
+
+  showFieldValueCodeSelection: (e) ->
+    attributes = @serialize(set: false)
+    @codeSelectionViewForFieldValues.updateConcepts(@getConcepts(attributes.code_list_id))
+    e.preventDefault()
+    @showAddCodesButton = false
+    @showAddCodes = true
+    @render()
+
+  removeFieldValueCode: (e) ->
+    e.preventDefault()
+    codeSet = $(e.target).model()?.attributes['codeset']
+    codeToRemove = $(e.target).model()?.attributes['code']
+    codeModel = @fieldValueCodesCollection.find (model) -> model.get('code') is codeToRemove and model.get('codeset') is codeSet
+    @fieldValueCodesCollection.remove(codeModel) if codeModel
 
   advanceFocusToInput: ->
     switch @model.get('type')
@@ -394,6 +446,9 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
     @$('select[name=type]').change()
     @triggerMaterialize()
     @$(':focusable:visible:first').focus()
+    @fieldValueCodesCollection.reset()
+    @showAddCodes = false
+    @render()
 
  class Thorax.Views.EditCriteriaReferenceView extends Thorax.Views.EditCriteriaValueView
   className: -> "#{if @fieldValue then 'field-' else ''}value-formset"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,7 +45,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/patientapi.git
-  revision: 74689decd9a79769c995664b368270acbe0d6d38
+  revision: 48dfd82014ee99e48e76c119a193b290dcad4adf
   branch: master
   specs:
     hquery-patient-api (1.0.4)
--- a/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
@@ -170,6 +170,11 @@
               {{value}} {{unit}}
               {{title}}
             {{/if}}
+            {{#if codes}} - Codes:
+              {{#each codes}}
+                {{@key}}: {{this}}
+              {{/each}}
+            {{/if}}
             {{#button "removeValue" class="btn btn-link close delete"}}&times; <span class="sr-only">Delete</span>{{/button}}
           </span>
         {{/collection}}
--- a/app/assets/javascripts/templates/patient_builder/edit_value.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_value.hbs
@@ -65,4 +65,21 @@
     {{/button}}
   </div>
 </div>
+
+{{#if showAddCodesButton}}
+  <br><a href="#" id="addCodes" class="btn btn-default btn-sm">Add Codes</a>
+{{/if}}
+
+{{#if showAddCodes}}
+  <br>
+  {{#collection fieldValueCodesCollection class="existing-values"}}
+  <span>
+    {{codeset}}:
+    {{code}}
+    {{#button "removeFieldValueCode" class="btn btn-link close delete"}}&times; <span class="sr-only">Delete</span>{{/button}}
+  </span>
+  {{/collection}}
+  {{view codeSelectionViewForFieldValues}}
+{{/if}}
+
 {{#unless fieldValue}}</div>{{/unless}}
--- a/lib/measures/patient_builder.rb
+++ b/lib/measures/patient_builder.rb
@@ -214,7 +214,17 @@ module Measures
 
         # Format the field to be stored in a Record.
         if field.type == "CD"
-          field_value = value["code"] || Measures::PatientBuilder.select_code(field.code_list_id, value_sets)
+          if value["codes"]
+            # Multiple codes were specified
+            field_value = {}
+            field_value["codes"] = value["codes"]
+          elsif value["code"] && !value["code"].empty?
+            # A single code was specified
+            field_value = value["code"]
+          else
+            # No codes specified, default to first possible code
+            field_value = Measures::PatientBuilder.select_code(field.code_list_id, value_sets)
+          end
           field_value["title"] = Measures::PatientBuilder.select_value_sets(field.code_list_id, value_sets)["display_name"] if field_value
         else
           field_value = field.format


app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,85 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
     @model.clear()
     @triggerMaterialize()
 
+class Thorax.Views.CodeSelectionViewForFieldValues extends Thorax.Views.BuilderChildView
```
edeyoung: This class is almost a duplicate of `Thorax.Views.CodeSelectionView`. Would it be possible to either make `Thorax.Views.CodeSelectionView` multi-purpose to handle the field value use as well, or would it be possible to create a subclass of `Thorax.Views.CodeSelectionView` to use with field values? Or maybe create a parent class that both `CodeSelectionView` and `CodeSelectionViewForFieldValues` would subclass?

edeyoung:
This only seems to apply to Principal Diagnosis. Would it be good to have for Diagnosis as well?

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,85 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
   className: -> "#{if @fieldValue then 'field-' else ''}value-formset"
 
   template: JST['patient_builder/edit_value']
 
   initialize: ->
     @model.set('type', 'CD')
+    @codeSelectionViewForFieldValues = new Thorax.Views.CodeSelectionViewForFieldValues
+    @codeSelectionViewForFieldValues.editCriteriaView = this
+    @fieldValueCodes = {}
+    @fieldValueCodesCollection = new Thorax.Collections.Codes @fieldValueCodes, parse: true
```
edeyoung: why do you need both `fieldValueCodes` and `fieldValueCodesCollection`? They seem like they represent the same thing. Can you pare this down to a single representation?

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +400,73 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+
+  selectedCodeListId: ->
+    attributes = @serialize(set: false)
+    # If the this is a field value and the selected key was a principal diagnosis, allow for code selection
+    concepts = @measure?.valueSets().findWhere(oid: attributes.code_list_id)?.get('concepts')
+    if @fieldValue and attributes.key == 'PRINCIPAL_DIAGNOSIS' and concepts and @$("select[name=type]").val() == "CD"
+      concepts = @measure?.valueSets().findWhere(oid: attributes.code_list_id)?.get('concepts')
+      @codeSelectionViewForFieldValues.updateConcepts(concepts)
+
+  updateCodesForFieldValue: (codes) ->
```
edeyoung: I'm not a fan of having the child view call into the parent view using this method. 

What about structuring it along the lines of this:
- you use `fieldValueCodesCollection` as your one representation
- you pass this into `CodeSelectionViewForFieldValues` as `@codes`
- you add an event handler onto `fieldValueCodesCollection` (e.g. on line 220: `@codes.on 'add remove'`)
- from this event handler you manage your re-rendering

This seems cleaner to me on a few levels: you don't have the child calling an action on the parent and you're using one code list instead of copying them between views.

Thoughts?

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +400,73 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+    @fieldValueCodes = codes
+    @fieldValueCodesCollection = new Thorax.Collections.Codes @fieldValueCodes, parse: true
+    @render()
+    @$('#codeSelection').show()
```
edeyoung: you're messing with the dom directly here. Is there a way you can do this using thorax/handlebars/render calls/etc.?

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +400,73 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
       delete attr.start_time
       title = @measure?.valueSets().findWhere(oid: attr.code_list_id)?.get('display_name')
       attr.title = title if title
+      attr.codes = @fieldValueCodes unless jQuery.isEmptyObject(@fieldValueCodes)
     rendered: ->
       @$("select[name=type]").selectBoxIt('native': true)
       @$('.date-picker').datepicker().on 'changeDate', _.bind(@validateForAddition, this)
       @$('.time-picker').timepicker(template: false).on 'changeTime.timepicker', _.bind(@validateForAddition, this)
+      @$('#codeSelection').hide()
```
edeyoung: dom elements should be manipulated directly as little as possible in this code. for all the `show`, `hide`, etc., dom actions, can you redo this using handlebars/thorax/etc.? See comment below as well.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +400,73 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
       @advanceFocusToInput()
+      @showAddCodesOption()
     'keyup input': 'validateForAddition'
     'change select[name=key]': 'changeFieldValueKey'
     # hide date-picker if it's still visible and focus is not on a .date-picker input (occurs with JAWS SR arrow-key navigation)
     'focus .form-control': (e) -> if not @$(e.target).hasClass('date-picker') and $('.datepicker').is(':visible') then @$('.date-picker').datepicker('hide')
+    'click #addCodes': 'showFieldValueCodeSelection'
+
+  showAddCodesOption: ->
+    attributes = @serialize(set: false)
```
edeyoung: why serialize here? consider using `@model` and then use the `get` operator for attributes later.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +400,73 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+    if bonnie.isPortfolio and @fieldValue and attributes.key == 'PRINCIPAL_DIAGNOSIS' and concepts and @$("select[name=type]").val() == "CD"
+      # Show code selection for field value
+      @$('#addCodesOption').fadeIn()
+      @$('#codeSelection').hide()
+    else
+      # Hide code selection for field value
+      @$('#addCodesOption').hide()
+      @$('#codeSelection').hide()
+
+  showFieldValueCodeSelection: (e) ->
```
edeyoung: this seems like it might end up being unnecessary if the show/hide actions are made more thorax-y and if the `@codeSelectionViewForFieldValues.codes` are tied directly to `updateCodesForFieldValue`

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +400,73 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+    e.preventDefault()
+    @codeSelectionViewForFieldValues.codes = {}
+    @updateCodesForFieldValue({})
+    @$('#addCodesOption').hide()
+    @$('#codeSelection').fadeIn()
+
+  selectedCodeListId: ->
+    attributes = @serialize(set: false)
```
edeyoung: this seems to duplicate the initial logic of `showAddCodesOption`. What about renaming `showAddCodesOption` and then adding the `@codeSelectionViewForFieldValues.updateConcepts(concepts)` to method and removing this method?

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +400,73 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+    # If the this is a field value and the selected key was a principal diagnosis, allow for code selection
+    concepts = @measure?.valueSets().findWhere(oid: attributes.code_list_id)?.get('concepts')
+    if @fieldValue and attributes.key == 'PRINCIPAL_DIAGNOSIS' and concepts and @$("select[name=type]").val() == "CD"
+      concepts = @measure?.valueSets().findWhere(oid: attributes.code_list_id)?.get('concepts')
```
edeyoung: `concepts` is defined above

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,62 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
+      $codeList.append("<option value>#{blankEntry}</option>")
+      for concept in @concepts when concept.code_system_name is codeSet and !concept.black_list
+        $('<option>').attr('value', concept.code).text("#{concept.code} (#{concept.display_name})").appendTo $codeList
+    else
+      @isCustom = true
+    @$('.codelist-control').focus()
+
+  addCode: (e) ->
+    e.preventDefault()
+    attributes = @serialize()
```
edeyoung: `serialize` takes the form data and puts it into the model. If you do `set: false`, then it just returns the attribute set without putting the form data into the model.

Given that, I'd do one of two things here:

Take the attributes from the serialization and don't set the model

```
attributes = @serialize(set: false)
... attributes.codeset ...
```

OR

Just call serialize and use the model

```
@serialize()
... model.get('codeset') ...
```

What we need to figure out is, do we want to set the model here or not. Because `@model.clear()` is being called below, I'm guessing that we don't want to be setting the model.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,62 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
+    # add the code unless there is a pre-existing code with the same codeset/code
+    if attributes['codeset'] is 'custom'
```
edeyoung: Consider changing to `attributes.codeset` here and for similar cases

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,62 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
+      attributes['codeset'] = attributes['custom_codeset']
+      attributes['code'] = attributes['custom_code']
+    codes = @fieldValueCodesCollection.toJSON()
+    if codes[attributes['codeset']]? and attributes['code'] in codes[attributes['codeset']]
+      return
```
edeyoung: Do we want to be returning here?
Do we want to be skipping all the rest of the logic (e.g. `@model.clear()`, `@triggerMaterialize()`, etc.)? Maybe we do, but I just want to be sure of that.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,62 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
+
+  addCode: (e) ->
+    e.preventDefault()
+    attributes = @serialize()
+    # add the code unless there is a pre-existing code with the same codeset/code
+    if attributes['codeset'] is 'custom'
+      attributes['codeset'] = attributes['custom_codeset']
+      attributes['code'] = attributes['custom_code']
+    codes = @fieldValueCodesCollection.toJSON()
+    if codes[attributes['codeset']]? and attributes['code'] in codes[attributes['codeset']]
```
edeyoung: There seems to be some repeated logic in this section. If we decide we don't need the `return` statement (see comment below), what about something like:

```
codes[attributes.codeset] = [] unless codes[attribute.codeset]?
codes[attributes.codeset].push attributes.code unless attributes.code in codes[attributes.codeset]
```

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,62 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
+    @$('.codelist-control').focus()
+
+  addCode: (e) ->
+    e.preventDefault()
+    attributes = @serialize()
+    # add the code unless there is a pre-existing code with the same codeset/code
+    if attributes['codeset'] is 'custom'
+      attributes['codeset'] = attributes['custom_codeset']
+      attributes['code'] = attributes['custom_code']
+    codes = @fieldValueCodesCollection.toJSON()
```
edeyoung: Look into doing everything with `@fieldValueCodesCollection` rather than serializing and deserializing into JSON. The Thorax/backbone structures have very helpful underscore methods associated with them so you don't have to go into the abstraction layers (like `@fieldValueCodesCollection.models...`). Take a look here: http://backbonejs.org/#Collection

Using these, you can probably do similar check/addition for new codes without having to do the de/serialization steps.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,62 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
+    if codes[attributes['codeset']]? and attributes['code'] in codes[attributes['codeset']]
+      return
+    if codes[attributes['codeset']]?
+      codes[attributes['codeset']].push attributes['code']
+    else
+      codes[attributes['codeset']] = []
+      codes[attributes['codeset']].push attributes['code']
+    @fieldValueCodesCollection.reset().add(@fieldValueCodesCollection.parse(codes))
```
edeyoung: See comment above about the de/serialization into JSON. If you keep things with `@fieldValueCodesCollection`, you won't need to reset the collection and then reload the set of codes, you can just add the new code.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +377,71 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+    if @shouldAllowFieldValueCodeSelection()
+      # Show code selection for field value
+      @showAddCodesButton = true
+      @showAddCodes = false
+    else
+      @showAddCodesButton = false
+      @showAddCodes = false
+    @fieldValueCodesCollection.reset()
+    @render()
+    @validateForAddition()
```
edeyoung: Consider removing `@validateForAddition` from this method. This is always called in the same context as `toggleAddCodesButton` is called so you get redundant calls.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +377,71 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+
+  toggleAddCodesButton: ->
+    if @shouldAllowFieldValueCodeSelection()
+      # Show code selection for field value
+      @showAddCodesButton = true
+      @showAddCodes = false
+    else
+      @showAddCodesButton = false
+      @showAddCodes = false
+    @fieldValueCodesCollection.reset()
```
edeyoung: What about moving `@fieldValueCodesCollection.reset()` into the first `if` statement? It seems like this probably only needs to get called when you'll be displaying the "add codes" button.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +377,71 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+    return bonnie.isPortfolio and @fieldValue and attributes.key in ['PRINCIPAL_DIAGNOSIS', 'DIAGNOSIS'] and @concepts and @$("select[name=type]").val() == "CD"
+
+  toggleAddCodesButton: ->
+    if @shouldAllowFieldValueCodeSelection()
+      # Show code selection for field value
+      @showAddCodesButton = true
+      @showAddCodes = false
+    else
+      @showAddCodesButton = false
+      @showAddCodes = false
```
edeyoung: since `@showAddCodes = false` gets called in either case, consider moving outside of the `if` clause.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +377,71 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+    @fieldValueCodesCollection.reset()
+    @render()
+    @validateForAddition()
+
+  showFieldValueCodeSelection: (e) ->
+    e.preventDefault()
+    @showAddCodesButton = false
+    @showAddCodes = true
+    @render()
+    @validateForAddition()
```
edeyoung: consider moving `@validateForAddition()` out of this method and into the event call above, similar to how the `select` events are done.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +377,71 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+
+  selectedCodeListId: ->
+    if @shouldAllowFieldValueCodeSelection()
+      @codeSelectionViewForFieldValues.updateConcepts(@concepts)
```
edeyoung: In this method, you're using `@concepts`, whose existence is dependent on `@shouldAllowFieldValueCodeSelection` being called. This is side-effecty.
Also, it looks like updateConcepts should only (always?) be called when `@shouldAllowFiledValueCodeSelection` is true.

What about restructuring things along these lines:

<pre>
  shouldAllowFieldValueCodeSelection: <b>(concepts)</b> ->
    attributes = @serialize(set: false)
    return bonnie.isPortfolio and @fieldValue and attributes.key in ['PRINCIPAL_DIAGNOSIS', 'DIAGNOSIS'] and concepts and @$("select[name=type]").val() == "CD"

  toggleAddCodesButton: ->
    <b>concepts = @measure?.valueSets().findWhere(oid: attributes.code_list_id)?.get('concepts')</b>
    if @shouldAllowFieldValueCodeSelection<b>(concepts)</b>
      # Show code selection for field value
      @showAddCodesButton = true
      @showAddCodes = false
      <b>@codeSelectionViewForFieldValues.updateConcepts(concepts)</b>
    else
      @showAddCodesButton = false
      @showAddCodes = false
    @fieldValueCodesCollection.reset()
    @render()
    @validateForAddition()
</pre>


This way, `concepts` is getting created in a known/seen way and you reduce the need for the other function.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +377,71 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+      @toggleAddCodesButton()
       @validateForAddition()
       @advanceFocusToInput()
     'keyup input': 'validateForAddition'
     'change select[name=key]': 'changeFieldValueKey'
     # hide date-picker if it's still visible and focus is not on a .date-picker input (occurs with JAWS SR arrow-key navigation)
     'focus .form-control': (e) -> if not @$(e.target).hasClass('date-picker') and $('.datepicker').is(':visible') then @$('.date-picker').datepicker('hide')
+    'click #addCodes': 'showFieldValueCodeSelection'
+
+  shouldAllowFieldValueCodeSelection: ->
```
edeyoung: consider renaming to `canSelectFieldValueCode` or `isFieldValueCodeSelectionAllowed` or some sort of question-based name to match the boolean output

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -328,21 +377,71 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+    @render()
+    @validateForAddition()
+
+  selectedCodeListId: ->
+    if @shouldAllowFieldValueCodeSelection()
+      @codeSelectionViewForFieldValues.updateConcepts(@concepts)
+
+  removeFieldValueCode: (e) ->
+    e.preventDefault()
+    codeSets = @fieldValueCodesCollection.toJSON()
```
edeyoung: see comments above regarding de/serializing the collection into JSON and using the built-in methods (http://backbonejs.org/#Collection) instead

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,62 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
...
+class Thorax.Views.CodeSelectionViewForFieldValues extends Thorax.Views.CodeSelectionView
+  initialize: ->
+    @model = new Thorax.Model
+
+  updateConcepts: (concepts) ->
+    @concepts = concepts
+    @codeSets = _(concept.code_system_name for concept in @concepts || []).uniq()
+    @render()
+
+  changeConcepts: (e) ->
```
edeyoung: It looks like this is a mirror of the parent class's `changeConcepts` except that it sets `@isCustom` and uses `@concepts` rather than `@criteria.valueSet().get('concepts')`

Would it be possible in the parent class to have `@concepts` which gets set to `@criteria.valueSet().get('concepts')` in the parent, but then this class sets it differently (in `updateConcepts`), and then just use the parent class method?

Is `@isCustom` used anywhere? I didn't see it, but if it is, this could also be added to the parent class method.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -304,13 +304,62 @@ class Thorax.Views.MedicationFulfillmentsView extends Thorax.Views.BuilderChildV
     @model.clear()
     @triggerMaterialize()
 
+class Thorax.Views.CodeSelectionViewForFieldValues extends Thorax.Views.CodeSelectionView
+  initialize: ->
```
edeyoung: what about instead of using `@fieldValueCodesCollection`, you used the `@codes` from the parent class, but you just set it differently in the initializer for this class (set it to the field value codes rather than to the model codes)? 

If you did that, could you get rid of the `addCode` method here? It seems like the only difference is the `@fieldValueCodesCollection` and `@codes`. I think these are the same structure type (a `Thorax.Collection.Codes` type), so this should be doable.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
holmesie: Yes, since we do not want the user to be able to add the same code/codeset twice, we want to be returning here.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
holmesie: While we do need the return, this makes sense (with a slight modification, i.e. without the last unless).

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -216,9 +216,11 @@ class Thorax.Views.CodeSelectionView extends Thorax.Views.BuilderChildView
 
   initialize: ->
     @model = new Thorax.Model
-    @codes = @criteria.get('codes')
-    @codes.on 'add remove', => @criteria.set 'code_source', (if @codes.isEmpty() then 'DEFAULT' else 'USER_DEFINED'), silent: true
-    @codeSets = _(concept.code_system_name for concept in @criteria.valueSet()?.get('concepts') || []).uniq()
+    @concepts = @criteria.valueSet().get('concepts') if @criteria
+    unless @codes
+      @codes = @criteria.get('codes') unless @codes
```
edeyoung: this `unless` statement is redundant given the statement above

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -216,9 +216,11 @@ class Thorax.Views.CodeSelectionView extends Thorax.Views.BuilderChildView
 
   initialize: ->
     @model = new Thorax.Model
-    @codes = @criteria.get('codes')
-    @codes.on 'add remove', => @criteria.set 'code_source', (if @codes.isEmpty() then 'DEFAULT' else 'USER_DEFINED'), silent: true
-    @codeSets = _(concept.code_system_name for concept in @criteria.valueSet()?.get('concepts') || []).uniq()
+    @concepts = @criteria.valueSet().get('concepts') if @criteria
+    unless @codes
```
edeyoung: We discussed something a bit different yesterday.

Here, `concepts` is getting set to `@criteria.valueSet().get('concepts')` in what is now a general purpose class. `@criteria.valueSet().get('concepts')` is the correct value for `concepts` in `EditCriteriaView`, but it's not the correct value in `EditCriteriaValueView`. I see that `if @criteria` is used, which makes it so this isn't actually set when created in `EditCriteriaValueView`, but that's implicit and dependent on how the method gets called.

This is similar for `codes`.

What we want is something where the values are explicitly set by whatever is initializing it so if `CodeSelectionView` gets reused again, there won't be confusion/issues/surprises (e.g. if someone set `criteria` on initialization, but didn't want `concepts` to be `@criteria.valueSet().get('concepts')`).

For example, you could have (untested):

```
class Thorax.Views.CodeSelectionView extends Thorax.Views.BuilderChildView
  ...
  initialize: ->
    @model = new Thorax.Model
    if @criteria
      @codes.on 'add remove', => @criteria.set 'code_source', (if @codes.isEmpty() then 'DEFAULT' else 'USER_DEFINED'), silent: true
  ...

class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
  ...
    @editCodeSelectionView = new Thorax.Views.CodeSelectionView criteria: @model, codes: @model.get('codes')
    @editCodeSelectionView.updateConcepts(@model.valueSet().get('concepts'))
  ...

class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
  ...
    @codeSelectionViewForFieldValues = new Thorax.Views.CodeSelectionView codes: @fieldValueCodesCollection
  ...
```

You may even want to put `@codes.on 'add remove', => @criteria.set 'code_source', (if @codes.isEmpty() then 'DEFAULT' else 'USER_DEFINED'), silent: true` outside of the class itself and not even bother with `criteria`:

```
class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
  ...
    codes = @model.get('codes')
    codes.on 'add remove', => @model.set 'code_source', (if codes.isEmpty() then 'DEFAULT' else 'USER_DEFINED'), silent: true
    @editCodeSelectionView = new Thorax.Views.CodeSelectionView codes: codes
    @editCodeSelectionView.updateConcepts(@model.valueSet().get('concepts'))
  ...
```

edeyoung: actually, i think last example makes the most sense and you should pull `criteria` out of the class. `criteria` is only used in `CodeSelectionView` to set an event on `codes`, and this isn't used by all classes using `CodeSelectionView`.

PR #537, opened 2016-06-06
Add expected values to the fields exported in the QRDA file.

mulcahyk:
This request came from a Bonnie user via Rute asking for the ability to include the expected values in QRDA export.  

I was able to get all but the continuous variables into the export.  I could not find anywhere in the QRDA spec to include them.

There is a companion pull request in Health Data Standards ([#386](https://github.com/projectcypress/health-data-standards/pull/386)) that needs to be accepted first.

### DIFF
--- a/lib/ext/record.rb
+++ b/lib/ext/record.rb
@@ -43,4 +43,30 @@ class Record
     end
   end
 
+  # Supports the exporting of the expected values in the QRDA export of the patients
+  # Note: There is logic in health-data-standards _measures.cat1.erb for further formatting.
+  # The special handling of the key 'DENEX' is do to the fact that the description of this population
+  #  stored in many of the measures is incorrect (it is often just 'Denominator' instead of
+  #  'Denominator Exclusion').
+  def expected_values_for_qrda_export(measure)
+    qrda_expected_values = []
+    expected_pop_names = HQMF::PopulationCriteria::ALL_POPULATION_CODES - %w{STRAT OBSERV}
+    measure.populations.each_with_index do |pop, idx|
+      pop.each do |pkey, pvalue|
+        next unless expected_pop_names.include?(pkey)
+        this_ev = {}
+        this_ev[:hqmf_id] = measure.population_criteria[pvalue.to_s]['hqmf_id']
+        if pkey == 'DENEX'
+          this_ev[:display_name] = 'Denominator Exclusions'
+        else
+          this_ev[:display_name] = measure.population_criteria[pvalue.to_s]['title']
+        end
+        this_ev[:code] = pkey
+        this_ev[:expected_value] = self.expected_values[idx][pkey].to_s
+        qrda_expected_values << this_ev
+      end
+    end
+    qrda_expected_values
+  end
+
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,7 @@ GIT
 
 GIT
   remote: https://github.com/projectcypress/health-data-standards.git
-  revision: 80551e5326937769d68dac5b3176ea79b60ce8f9
+  revision: 9e827235a53cd6d2719b48d50bfe820a8f997b38
   branch: master
   specs:
     health-data-standards (3.6.1)
@@ -23,7 +23,7 @@ GIT
       memoist (~> 0.9.1)
       mongoid (~> 4.0.0)
       mongoid-tree (~> 1.0.4)
-      nokogiri (~> 1.6.7.1)
+      nokogiri (~> 1.6.8)
       protected_attributes (~> 1.0.5)
       rest-client (~> 1.8.0)
       rubyzip (= 0.9.9)
@@ -133,7 +133,7 @@ GEM
       warden (~> 1.2.3)
     diffy (3.0.7)
     docile (1.1.5)
-    domain_name (0.5.20160310)
+    domain_name (0.5.20160615)
       unf (>= 0.0.5, < 1.0.0)
     erubis (2.7.0)
     eventmachine (1.0.7)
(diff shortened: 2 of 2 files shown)


app/controllers/patients_controller.rb
```diff
@@ -69,6 +69,7 @@ def qrda_export
         patient_measure = measure || get_associated_measure(patient)
         # attach the QRDA export, or the error
         begin
+          patient['qrda_expected_values'] = patient.qrda_expected_values(patient_measure)
```
pkmitre: I feel like there's a better approach than setting an attribute based on a function call that's on the same object; why not just build function calls that get used within the template?

lib/ext/record.rb
```diff
@@ -43,4 +43,29 @@ def rebuild!(payer=nil)
     end
   end
 
+  def qrda_expected_values(measure)
+    evs = []
+    measure.first.populations.each_with_index do |pop, idx|
```
pkmitre: This hard codes it to the first measure, which is a mismatch against the QRDA: the QRDA iterates through the measures, adding a section per-measure; why not pass the measure in within the QRDA template so that that iteration gets leveraged?

lib/ext/record.rb
```diff
@@ -43,4 +43,29 @@ def rebuild!(payer=nil)
...
+      pop.each do |pkey, pvalue|
+        if pkey == 'id' || pkey == 'title' || pkey.downcase.match(/^strat/) || pkey == 'OBSERV'
```
pkmitre: I'd consider white listing instead of black listing for your iterator skipping; also, this is a good opportunity for using a single line

```
next if ...
```

approach

lib/ext/record.rb
```diff
@@ -43,4 +43,29 @@ def rebuild!(payer=nil)
...
+        this_ev[:hqmf_id] = measure.first.population_criteria[pvalue.to_s]['hqmf_id']
+        if pkey == 'DENEX'
+          this_ev[:display_name] = 'Denominator Exclusions'
+        else
+          this_ev[:display_name] = measure.first.population_criteria[pvalue.to_s]['title']
+        end
+        this_ev[:code] = pkey == 'IPP' ? 'IPOP' : pkey
+        # Going to assume that any expected_value > 0 is a TRUE
+        # As per HL7 CDA R2:QRDA I, R1, STU Release 3.1, US Realm, Vol. 2 —Templates and Supporting Material
+        # Set negationInd="false" to assert true (that the criteria are met by the included data). Set negationInd="true" to assert false (that the criteria are not met by the included data).
```
pkmitre: I think the inversion logic and commentary is more appropriate for within the template, i'd prefer to minimize the QRDA-specificity of code in the patient model.

lib/ext/record.rb
mulcahyk: `def qrda_patient_export(patient, measure)
    start_time = Time.new(Time.zone.at(APP_CONFIG['measure_period_start']).year, 1, 1)
    end_time = Time.new(Time.zone.at(APP_CONFIG['measure_period_start']).year, 12, 31)
    qrda_exporter = HealthDataStandards::Export::Cat1.new
    qrda_exporter.export(patient, measure, start_time, end_time)
  end`
I read this function as only taking one measure at a time.  If there are multiple patients it will iteration over the patients and just get the first measure for each patient.
`def get_associated_measure(patient)
    Measure.where(hqmf_set_id: patient.measure_ids.first)
  end`

pkmitre: Definitely a case of bad variable names and odd coding choices in the original code:

```
Measure.where(hqmf_set_id: patient.measure_ids.first)
```

actually returns a set of measures; in this case we'd expect it to only contain one, but it's an array so should probably either be called "measures" or the first element should be selected from the set and returned.

I think the original issue is that the QRDA exporter actually expects a set of measures (since a QRDA file can be created in the context of more than one measure); ie the call to

```
    qrda_exporter.export(patient, measure, start_time, end_time)
```

the second argument is expected to be a set of measures.

Please feel free to refactor the code in app/controllers/patients_controller.rb to clear this up!

lib/ext/record.rb
```diff
@@ -43,4 +43,27 @@ def rebuild!(payer=nil)
     end
   end
 
+  # Supports the exporting of the expected values in the QRDA export of the patients
+  # Note: There is logic in health-data-standards _measures.cat1.erb for further formatting.
+  def expected_values_for_qrda_export(measure)
+    qrda_expected_values = []
+    expected_pop_names = HQMF::PopulationCriteria::ALL_POPULATION_CODES - %w{(STRAT OBSERV)}
```
pkmitre: The extra parenthesis around STRAT and OBSERV should be removed, they get included and lead to ["(STRAT", "OBSERV)"], we just want %w{STRAT OBSERV}.

lib/ext/record.rb
```diff
@@ -43,4 +43,27 @@ def rebuild!(payer=nil)
...
+    measure.populations.each_with_index do |pop, idx|
+      pop.each do |pkey, pvalue|
+        next unless expected_pop_names.include?(pkey)
+        this_ev = { code: nil, display_name: nil, hqmf_id: nil, expected_value: nil }
```
pkmitre: We don't need to initialize the hash keys with empty values, this_ev = {} works fine and is clearer.

lib/ext/record.rb
```diff
@@ -43,4 +43,27 @@ def rebuild!(payer=nil)
...
+        this_ev[:hqmf_id] = measure.population_criteria[pvalue.to_s]['hqmf_id']
+        if pkey == 'DENEX'
+          this_ev[:display_name] = 'Denominator Exclusions'
```
pkmitre: For measures that I've looked at, the title for the DENEX is 'Denominator Exclusions', why is this specific test needed?

mulcahyk: I forget the exact root at the moment but I found that in the definition of the measures that description isn’t in there.

pkmitre: OK, so it looks like the definition is there, but in the case of the DENEX the title is sometimes (more than 50% of the time!) incorrect, reading Denominator rather than Denominator Exclusion; I think it would be a good idea to add a comment for special case handling (specifically here and generally) indicating why the special handling is necessary.

PR #538, opened 2016-06-08
Patient Builder Improvements

DPC15038:


app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -21,6 +21,45 @@
...
+      {{#ifCond fulfillment_information.length ">" 0}}
+        <div style="line-height:135%">&nbsp;</div> {{!Adds a thin blank line (thinner than <br>) between each item}}
+        <label class="control-label"><i class="fa fa-medkit" aria-hidden="true"></i>&nbsp;Filled<input class="sr-only" disabled></label>
+          {{#each fulfillment_information}}
+          <div> <strong> &ndash; </strong> &nbsp; {{this}} </div> 
+          {{/each}}
+      {{/ifCond}}          
+      {{!=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=}}
+      {{#ifCond result_information.length ">" 0}}
+        <div style="line-height:135%">&nbsp;</div> {{!Adds a thin blank line (thinner than <br>) between each item}}
```
thecristen: An `<hr>` might make more sense than a pseudo-empty `<div>` here. I'd also avoid using the `style` attribute here and add a CSS class to adjust the appearance.

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -21,6 +21,45 @@
           {{/button}}
         </div>
       </div>
+      <div class="row">
+        <div class="col-md-12">
+      {{!The following ifConds display information when the window is minimized}}
+      {{#ifCond fulfillment_information.length ">" 0}}
```
thecristen: Would `{{#if fulfillment_information}}` work just as well here?

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -21,6 +21,45 @@
...
+        <div style="line-height:135%">&nbsp;</div> {{!Adds a thin blank line (thinner than <br>) between each item}}
+        <label class="control-label"><i class="fa fa-bar-chart-o" aria-hidden="true"></i>&nbsp;Result<input class="sr-only" disabled></label>
+          {{#each result_information}}
+          <div> <strong> &ndash; </strong> &nbsp; {{this}} </div> 
+          {{/each}}
+      {{/ifCond}}          
+      {{!=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=}}
+      {{#ifCond field_value_information.length ">" 0}}
+        <div style="line-height:135%">&nbsp;</div> {{!Adds a thin blank line (thinner than <br>) between each item}}
+        <label class="control-label"><i class="fa fa-bar-chart-o" aria-hidden="true"></i>&nbsp;Field Values<input class="sr-only" disabled></label>
```
thecristen: Secret font-awesome tip: adding the `fa-fw` class to the icon makes it fixed width, which will reduce the need to include `&nbsp;` after every icon.

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -21,6 +21,45 @@
...
+          {{#each field_value_information}}
+          <div> <strong> &ndash; </strong> &nbsp; {{this}} </div>
```
thecristen: This is more of a preference thing, but maybe replace`<strong> &ndash; </strong> &nbsp;` with `<i class="fa fa-fw fa-minus" aria-hidden="true"></i>` (or any other icon)? [Using list icons might work well here too](http://fontawesome.io/examples/#list).

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -95,6 +95,68 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
...
+        array_of_result_Ids.push(each_fulfillment) #Create an array of Fulfillment Ids (eg. c377, c836...etc)
+      if ((Object.keys(@model.attributes.fulfillments._byId).length) > 0) #If there are actually fulfillments (Medications)...
+        for this_specific_element in [0..(Object.keys(@model.attributes.fulfillments._byId).length)-1] #Loop that runs from 0 to the number of fulfillment values 
+          that = @model.attributes.fulfillments._byId["#{array_of_result_Ids[this_specific_element]}"].attributes
+          that = ((that.quantity_dispensed_value + " " + that.quantity_dispensed_unit + " " + that.dispense_date + " ")\
+          .replace(/_/g, ' ').replace(/\w\S*/g, (txt) ->  txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase()) + that.dispense_time)
+          #format the QRDA(or whatever healthcare language this is written in) by replacing underscores with spaces and capitalizing each word
+          
+          fulfillments_information_array.push(that)
+          #Add this formatted string to an array containing all the other fulfillment values
```
thecristen: I notice that a lot of this is building an array of strings corresponding to an array of things, and I see a great opportunity to use the map function here. Map basically takes an array, performs a function on every item in the array, and returns a new array. For instance, I think replacing the above section with this should produce the same result:

``` coffeescript
if @model.attributes.references?
  references_information_array = @model.attributes.references.map((r) ->
     # using .get to retreive attributes since r is a Thorax object
     if r.get('reference_type')
       return r.get('reference_type').replace(/_/g, ' ').replace(/\w\S*/g, (txt) ->  txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase()) + ": " + r.get('description') + ": " + r.get('start_date')
     else
       return r.get('description') + ": " + r.get('start_date'))
```

Tagging @edeyoung here since she might have more comments in this area.

DPC15038:
screen shot 2016-06-08 at 1 48 43 pm

rutemartins:
Looks good! Is there a way to reduce the amount of space each data element takes up? Most measures will have more than 10 elements, which makes the "see at a glance" a bit more difficult. The more we can show without scrolling, the better.

DPC15038:
screen shot 2016-06-09 at 9 35 27 am

edeyoung:
Would we want this to be the default view or would we still want to show the very minimal view, and people can choose to display the additional information?

edeyoung:
and I agree, this looks really great!

edeyoung:
I'd also be curious to see what this looks like on an actual test case for measure testing. Maybe they don't use nearly as many fields so it's a much more manageable view.

thecristen:
Future user stories include letting the user be able to filter/hide/show certain types of elements, sort elements, etc. With that additional level control, I think having this info displayed by default is OK.

DPC15038:
example of sort by elements

DPC15038:
screen shot 2016-06-16 at 4 36 49 pm

DPC15038:
screen shot 2016-06-16 at 4 38 22 pm

DPC15038:
duration for data criteria
Example of Duration

DPC15038:
code s now display when minimized
Codes display when minimized

DPC15038:
age now displays in patient builder
Age now displays in Patient Builder

edeyoung:
We recently got an email about needing to specify CPT codes in a very particular way in order to comply with a license agreement with AMA - so this may need to change. Don't know what that way is yet.

Also, it looks like the code numbers are no longer being displayed. I'd display the code number and the code description.

DPC15038:
@edeyoung The code numbers are still being displayed when the data criteria is open, but not when it is closed. Should I change it to include, 1. Codeset, 2. Code Number, and 3. Code Description when open?

edeyoung:
Everything should display when open (code set, number, description). When closed, I think it should either display code set and number (to save space) or code set, number, and description. In all cases, the code number should be shown.

thecristen:
>  In all cases, the code number should be shown.

@edeyoung I'm not sure the code number should get such priority. We've heard from users that that number is more noise than signal. 

If we do include the number (when closed), maybe we can omit the code set - if the associated code set can be inferred from the length/format/value of the `number`.

edeyoung:
I know the Cypress folks were specifically interested in the description, but I didn't hear that they weren't interested in the code. Has this come up in other conversations also?

I'm concerned about omitting a piece of data that the users have become accustomed to seeing.

rutemartins:
I agree with @edeyoung, the code can be helpful when debugging a test cases when the cause is value set related.

thecristen:
Please note that closed criteria showed _no information_ on codes, so this isn't "omitting" anything that was previously available.

rutemartins:
ah - wasn't thinking of the closed criteria. When open, it used to show codes only. I'm OK with not showing anything related to the code when the data criterion is closed. Usually I'm looking for attributes and date/time stamps.

edeyoung:
I think it's fine to not show anything about codes in the closed view if it doesn't make sense from a user perspective. If we are going to show code-related information, I think we should show the code number to maintain consistency and utility with what they're familiar with.

DPC15038:
Here's a side-by-side of without code number and with code number. Thoughts?
side by side code comparison

edeyoung:
With the code. But I think Rute makes a good point - maybe we don't need the codes in the summary view at all.

What about losing the space between the codeset and the code?
e.g.: CPT:99365 - Initial comprehensive...

DPC15038:
What do you think? screen shot 2016-06-28 at 7 59 54 am

edeyoung:
Looks good!!!

The "Show More" alignment seems to be a bit off.

DPC15038:
Examples of Age spanning Measure Periodexample of birth date

DPC15038:
Multiple **Show More** Buttons now format properlymultiple show mores

DPC15038:
Data Criteria Turns Red When Something is Not Performed 
example of notperformed

DPC15038:
Changed the Way Age and Durations are Displayed

## screen shot 2016-07-05 at 4 31 45 pm

screen shot 2016-07-05 at 4 31 41 pm

DPC15038:
Changed the Red to Yellow
screen shot 2016-07-06 at 12 14 21 pm

app/assets/javascripts/helpers.js.coffee
```diff
@@ -42,14 +46,10 @@ Handlebars.registerHelper 'ifCond', (v1, operator, v2, options) ->
...
-    out = ''
-    for index in [1..n]
-      out += opts.fn(this, data: { index: index })
-  else
-    out = opts.inverse(this)
-
-  return out
+# Helper to retrieve the element at a given position in an array
+Handlebars.registerHelper 'lookup', (array, index) ->
+  return array[index] || undefined
```
pkmitre: If the array element at index is 0 or false, this will return undefined instead of the actual value. You could use the actual handlebars implementation (from the latest version of handlebars, which we don't currently use in Bonnie) from here: https://github.com/wycats/handlebars.js/blob/master/lib/handlebars/helpers/lookup.js

app/assets/javascripts/helpers.js.coffee
```diff
@@ -42,14 +46,10 @@ Handlebars.registerHelper 'ifCond', (v1, operator, v2, options) ->
...
+  
+# Helper to make use of moment date formatting
+Handlebars.registerHelper 'dateFormat', (date, fmt) ->
```
pkmitre: We already have a date formatting helper, called "moment", defined at the top of this file.

DPC15038: Back when I was making the modal for measure-upload-summary dialog for test-case-history there was a helper called lookup I was using, but that helper wasn't available in master, so i just went into test case history, and copied out the entire helper file and pasted it in my own directory
EDIT: I originally needed the lookup helper for displaying the Codes, but since Ruth reccomended I remove them, I don't use the lookup helper anymore, so I don't need to change the helpers.js files at all

app/assets/javascripts/helpers.js.coffee
pkmitre: Actually, looks like Lizzie already added that implementation to the patient dashboard branch, back in January of 2016... it's of course not in master yet, there will be some merging needed in the future!

app/assets/javascripts/helpers.js.coffee
```diff
@@ -19,8 +19,12 @@ Handlebars.registerHelper 'complexityIcon', (score) ->
 # elements to display, not trustable for security purposes
 Handlebars.registerHelper 'ifAdmin', (options) ->
   if bonnie.isAdmin then options.fn(this) else options.inverse(this)
+
 Handlebars.registerHelper 'ifPortfolio', (options) ->
   if bonnie.isPortfolio then options.fn(this) else options.inverse(this)
+  
+Handlebars.registerHelper "times", (n, block) ->
+  (block.fn(i) for i in [0...n]).join("")
```
pkmitre: This implementation of times has different behavior from the one it's replacing in cases where n is 0 or otherwise falsy, the previous implementation appears to allow an else block to fire in those cases

DPC15038:
Changed the location of the "Utility" Buttons, and added two new buttons
screen shot 2016-07-18 at 2 33 27 pm

pkmitre:
The new interface is definitely an improvement, but there are still a few things that feel odd about it from a design perspective, that I think make the interface feel less cohesive than it could be:
- The SHOW INFORMATION / HIDE INFORMATION button still feels like it's in an odd place; some controls are at the top of the PATIENT HISTORY and some are at the bottom of the ELEMENTS, what's the reasoning behind the placements? In particular, when there are a lot of elements, a user would have to scroll to find that button.
- There are three different controls that basically toggle between two states (SHOW INFORMATION / HIDE INFORMATION, OPEN ALL / CLOSE ALL, and ORDER BY DATA / ORDER BY ELEMENTS) and three different mechanisms that get used to do the toggling (a button that changes, two buttons to click that don't show state, and two buttons to click that do show state).
- Only the OPEN ALL and CLOSE ALL are behind the gear, it seems off that some items are behind a gear and some (ORDER BY) are not.
- What does ORDER BY "ELEMENTS" mean? Alphabetical order by name? If so, why not just say "NAME"?

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+            , 0)
+
+  previewInformation: (e) ->
+  # Just like with the "Order By" buttons, if you spam/double click the "Preview/Hide Information"
+  # button it performs the action for each click. On patients with many events this can
+  # easily mean 20-30 seconds of "lag" if the user has clicked 5+ times
+  # Only allow the previewing/hiding to occur if they haven't clicked in .75 seconds
+  # TODO: The previewing/hiding button shares the same loading_spinner as the "Order By" buttons
+  # Maybe give it its own loading spinner?
+    if @timeSinceInfoPreviewClick + 750 < e.timeStamp
```
pkmitre: A better way to prevent multiple clicks is typically to disable the button as the first action after the click and then re-enable it once the action is complete (which you may get for free when the button changes). I think it's as easy as setting the disabled="true" property on the button, ie like @$('button#<idSelector>').prop('disabled', true).

pkmitre: Also, why is the lag so long? There's a pretty noticeable lag even with just a few elements, which surprises me... I'll poke around a bit.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -79,6 +78,26 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
     @editFulfillmentHistoryView = new Thorax.Views.MedicationFulfillmentsView
       model: new Thorax.Model
       criteria: @model
+    @showPatientInformation = if @builderView.previousStateWasHideInfo() then 1 else 0
```
pkmitre: Two things when dealing with booleans:

1) avoid using `0` and `1`, use the `false` and `true` literals instead

2) avoid testing a condition and then assigning true or false as a result, ie for this particular line use the following:

```
@showPatientInformation = @builderView.previousStateWasHideInfo()
```

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -89,16 +108,54 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
...
     cmsIdParts = @model.get("cms_id").match(/CMS(\d+)(V\d+)/i)
-    
     desc = @model.get('description').split(/, (.*:.*)/)?[1] or @model.get('description')
+    field_value_information_array = []
+    results_information_array = []
+    references_information_array = []
+    fulfillments_information_array = []
+
+    if @showPatientInformation # only perform the calculations if the patient is previewing the information
+      if @model.attributes.fulfillments? # LOGIC FOR FULFILLMENTS
```
pkmitre: In Thorax, it's bad practice to refer to the attributes field directly, this should be 

```
if @model.has('fulfillments')
```

And then, if you need to see the actual value someplace:

```
@model.get('fulfillments')
```

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -112,9 +169,18 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
...
+      negationCode: @getNegationCodeIDAsText(@model.get('negation_code_list_id'))
       hasStopTime: @model.hasStopTime()
       startLabel: @model.startLabel()
       stopLabel: @model.stopLabel()
+      showPatientInfo: @showPatientInformation
+      field_value_information: field_value_information_array
+      references_information: references_information_array
+      result_information: results_information_array
+      fulfillment_information: fulfillments_information_array
+      duration_between_element_start_and_finish: element_duration
```
pkmitre: For many of these lines above: when the same variable appears in multiple contexts, it's generally a lot easier to follow the code if the variable name is the same in both contexts.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+        when "sort_by_elements"
+          if @sortButton.buttonLastClicked != "Elements Button" || @sortButton.secondsSinceLastClick + 1000 < e.timeStamp
+            # only proceed if the last button they clicked wasn't Elements OR it has been 1 second since the last click
+            loadingSpinnerSelector.removeClass('hidden')
+            sortByElementsSelector.addClass('active')
+            sortByElementsSelector.html('<i class="fa fa-chevron-right"></i> <b>ELEMENTS</b>') # add bold tag and arrow to more visibly show which button is selected
+            sortByDateSelector.removeClass('active')
+            sortByDateSelector.html('DATE') # Remove the chevon arrow and bold tag from the other button, leave just the text
+            # Force these previous lines to occur before the sorting begins by waiting 0 seconds
+            setTimeout( =>
```
pkmitre: Can you explain a bit more about why the extensive use of setTimeout? It's a pretty unusual construct in view code (and maybe something we don't want to be doing), so I'd like to understand the reasoning more.

DPC15038: Hoss helped me with this bit, essentially I wanted the loading spinner to appear before anything else happened so the user could see that something was happening if/when the screen lagged. What was happening was that the loading spinner would appear after the lag started, so the screen would never update. Then when the lag stopped, the spinner would dissapear - AKA the user never sees the loading spinner and just sees lag. Hoss's solution was to set a timeout for 0 seconds, to force the code to show the spinner, then do everything else

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -21,6 +46,47 @@
...
         </div>
       </div>
+      {{#ifCond showPatientInfo "==" 1}} {{! If the user wishes to preview the information }}
+        <div class="row">
+          <div class="col-md-12">
+
+            {{#if fulfillment_information}}
+              <hr class= "hr-with-thin-margins"> {{! Thin line break between each element/value }}
+              {{#if isNegated}}<label>{{else}}<label class="control-label">{{/if}}<i class="fa fa-fw fa-medkit" aria-hidden="true"></i>Filled<input class="sr-only" disabled></label>
+                {{#each fulfillment_information}}
```
pkmitre: This #each looks like it's indented two spaces too far, which at first glance makes it look like it's controlled by the #if one line above it, which isn't the case.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+    if @timeSinceInfoPreviewClick + 750 < e.timeStamp
+      loadingSpinnerSelector = @$('#loading_spinner')
+      @timeSinceInfoPreviewClick = e.timeStamp # Overwrite previous timestamp with new one
+      loadingSpinnerSelector.removeClass('hidden')
+
+      if @previousStateWasHideInfo()
+        @$('#preview_information').text('Preview Information') # sets the button's new text
+        # Force these previous lines to occur before previewing patient info by waiting 0 seconds
+        setTimeout( =>
+          @trigger "hide_information_in_patient_builder"
```
pkmitre: The convention for event names is to use camel case with the first letter in lower case (hideInformationInPatientBuilder) rather than snake case.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
pkmitre: That makes sense, tweaking the comment could make this clearer, ie something like

```
# Put JS code that takes a long time to run into a setTimeout so the UI thread is
# free to update the display before processing starts
```

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
 
+  sortPatientEventsBy: (e) ->
+    # if the user accidentally clicks on DATE or ELEMENTS multiple times, it will sort multiple times
+    # even if everything is already sorted. By tracking sortButton, we can prevent the user from
+    # accidentally double clicking and waiting the extra few seconds that sorting takes.
+    loadingSpinnerSelector = @$('#loading_spinner')
+    sortByElementsSelector = @$('#sort_by_elements')
+    sortByDateSelector = @$('#sort_by_date')
+    if e.target.id?
+      switch e.target.id
```
pkmitre: This is a pretty complex dance of looking at the ID of the element to figure out the action, a better way to do this would be to have two separate functions, one for each button (ie sortPatientEventsByDate and sortPatientEventsByName), that then call a common function for any common code.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+        when "sort_by_elements"
+          if @sortButton.buttonLastClicked != "Elements Button" || @sortButton.secondsSinceLastClick + 1000 < e.timeStamp
+            # only proceed if the last button they clicked wasn't Elements OR it has been 1 second since the last click
+            loadingSpinnerSelector.removeClass('hidden')
+            sortByElementsSelector.addClass('active')
+            sortByElementsSelector.html('<i class="fa fa-chevron-right"></i> <b>ELEMENTS</b>') # add bold tag and arrow to more visibly show which button is selected
+            sortByDateSelector.removeClass('active')
+            sortByDateSelector.html('DATE') # Remove the chevon arrow and bold tag from the other button, leave just the text
```
pkmitre: Once you refactor this into two functions, it will probably also be cleaner to make use of jQuery selectors directly rather than assigning to variables, and perhaps chaining operators. ie

```
@$('#sort_by_date').removeClass('active').html('DATE')
```

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
       top: shiftDown
       bottom: $logic.nextAll(':visible').height() || 0
 
+  sortPatientEventsBy: (e) ->
+    # if the user accidentally clicks on DATE or ELEMENTS multiple times, it will sort multiple times
+    # even if everything is already sorted. By tracking sortButton, we can prevent the user from
+    # accidentally double clicking and waiting the extra few seconds that sorting takes.
+    loadingSpinnerSelector = @$('#loading_spinner')
```
pkmitre: The convention for HTML IDs is to use camel case with the first character lowercase, ie loadingSpinner

pkmitre:
The location of the age language seems a bit odd, it's wedged over to the right with the living status:

screen shot 2016-07-19 at 10 31 11 am

Seems like it would make more sense to include it under the birth date, since it's usually more closely related to that data, plus it makes better use of vertical real-estate:

screen shot 2016-07-19 at 10 44 06 am

app/assets/javascripts/templates/patient_builder/patient_builder.hbs
```diff
@@ -108,9 +86,37 @@
               </label>
             </div>
           {{/if}}
+          <hr class="hr-with-thin-margins">
+          <h5><span class="patient-age-label">{{!The text that goes here is added via jQuery in patient_builder.js.coffee}}</span>
+            <span class="patient-age">{{patientAge}}</span>
+          <h5>
```
pkmitre: Closing h5 is missing /

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+      patientAgeLabelSelector.text("") # Clear Age Label Text
+    else # Meaning the birthdate comes before deathdate
+      patientAgeLabelSelector.removeClass('fa fa-fw fa-exclamation-circle')
+      patientAgeLabelSelector.removeClass('fail fa fa-fw fa-times-circle')
+      patientAgeSelector.removeClass('fail')
+
+      if @patientStatus.patientIsAlive
+        patientAgeLabelSelector.text("Age at Start of Measure Period:")
+      else
+        patientAgeLabelSelector.text("Age at Time of Death: ")
```
pkmitre: Setting text via jQuery isn't a good pattern to use within Thorax code, it should really be set by using {{}} in the view and passing the data via the context.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+      patientAgeSelector.html("<br> #{@patientStatus.patientAge}") # Patient Age now displays all on one line (looks cleaner)
```
pkmitre: HTML (like &lt;br&gt;) should never appear in the view code, HTML should only appear in the template.

(Also see above re: setting text via jQuery)

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -130,8 +130,9 @@ class Thorax.Models.Patient extends Thorax.Model
   # Sort criteria by any number of attributes, first given highest priority
   sortCriteriaBy: (attributes...) ->
     originalComparator = @get('source_data_criteria').comparator
-    @get('source_data_criteria').comparator = (crit) -> _(attributes).map((attr) -> crit.get(attr))
-    @get('source_data_criteria').sort()
+    for current_attribute in attributes.reverse() #.reverse() gives the highest priority to the beggining of the array
```
pkmitre: Variable names in JavaScript use the camel-case-with-first-letter-lowercase style, ie currentAttribute

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
thecristen: Note: that requires a full re-rendering in order to display a new value. So going with that approach, one should create a little Thorax view so that one could just re-render that view rather than the entire page.

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -130,8 +130,9 @@ class Thorax.Models.Patient extends Thorax.Model
...
+      @get('source_data_criteria').comparator = current_attribute #.comparator cannot act on an array, it must be a single element
```
pkmitre: I don't believe the statement "comparator cannot act on an array, it must be a single element" is true, I tested out the approach used in the original code and it works fine; it's also more efficient than this approach of sorting multiple times.

pkmitre: To be clear: to sort on an array, the array has to be provided in the context of a sort function, not as a literal array, ie

```
@get('source_data_criteria').comparator = [attribute1, attribute2]
```

won't work but

```
@get('source_data_criteria').comparator = (model) -> [model.get('attribute1'), model.get('attribute2')]
```

will.

DPC15038: > comparator cannot act on an array, it must be a single element

This might be false, but it is true that the original code doesn't work under all circumstances. It can sort by name, however, it cannot sort dates under 2000
Edit: Discovered where/why the error is occuring, see post below

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+          if @sortButton.buttonLastClicked != "Date Button" || @sortButton.secondsSinceLastClick + 1000 < e.timeStamp
+            # only proceed if the last button they clicked wasn't Date OR it has been 1 second since the last click
+            loadingSpinnerSelector.removeClass('hidden')
+            sortByDateSelector.addClass('active')
+            sortByDateSelector.html('<i class="fa fa-chevron-right"></i> <b>DATE</b>') # add bold tag and arrow to more visibly show which button is selected
+            sortByElementsSelector.removeClass('active')
+            sortByElementsSelector.html('ELEMENTS') # Remove the chevon arrow and bold tag from the other button, leave just the text
+            # Force these previous lines to occur before the sorting begins by waiting 0 seconds
+            setTimeout( =>
+              @model.sortCriteriaBy 'start_date'
```
pkmitre: Elsewhere where we sort by date we also include the end_date as a secondary sort option.

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -1,5 +1,10 @@
-<div class="criteria-type-marker">
+{{#if isNegated}}
+  <div class="criteria-type-marker-with-negation">
   <i class="fa fa-fw fa-lg {{faIcon}}" aria-hidden="true"></i>
+{{else}}
+  <div class="criteria-type-marker">
+  <i class="fa fa-fw fa-lg {{faIcon}}" aria-hidden="true"></i>
+{{/if}}
 </div>
```
pkmitre: The only thing that's different in the two code blocks is the class, so this could be simplified as

```
<div class="{{#if isNegated}}criteria-type-marker-with-negation{{else}}criteria-type-marker{{/if}}
  <i class="fa fa-fw fa-lg {{faIcon}}" aria-hidden="true"></i>
</div>
```

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
DPC15038: Yep! If my understanding is correct, secondary sort options only occur when the primary ones are equal. I'm not entirely sure how comparator and .sort() work, but even without the secondary option of 'end_date' two elements with the same start date automatically sort by this secondary condition. As a result, I intentionally didn't add that secondary sort option, as it just adds extra work
Example: **8:00 - 8:15** will come before **8:00 - 8:25**

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -89,16 +108,54 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
...
+    fulfillments_information_array = []
+
+    if @showPatientInformation # only perform the calculations if the patient is previewing the information
+      if @model.attributes.fulfillments? # LOGIC FOR FULFILLMENTS
+        fulfillments_information_array = @model.attributes.fulfillments.map((ful) ->
+           # using .get to retreive attributes since ful is a Thorax object
+             return ((ful.get('quantity_dispensed_value') + " " + ful.get('quantity_dispensed_unit') + " " + ful.get('dispense_date') + " ").replace(/_/g, ' ').replace(/\w\S*/g, (txt) ->  txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase()) + ful.get('dispense_time')))
+
+      if @model.attributes.field_values? # LOGIC FOR FIELD VALUES
+        that_model = @model #store this.model to be used inside the map
```
pkmitre: The reason you can't access @model inside the map is because you're using a thin arrow for the function, use a fat arrow (=>) and you won't need to juggle with a variable called "that_model". To read more about it look for the section called "Bound Functions" in http://coffeescript.org/

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
pkmitre: Good call, "microviews" can definitely be a win, depending on the size of the original view.

app/assets/javascripts/models/patient.js.coffee
pkmitre: Why can't it sort dates under 2000? The following two pieces of code seem like they should be equivalent:

```
@get('source_data_criteria').comparator = 'start_date'

@get('source_data_criteria').comparator = (model) -> [model.get('start_date')]
```

You're saying that the second version handles dates differently?

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
pkmitre: I suspect that what you saw was just a coincidence, since if you don't make it explicit what would actually cause that to happen? I just tested it, and it doesn't actually sort by the end_date when the start_dates are equal.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -112,9 +169,18 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
       faIcon: @model.faIcon()
       definition_title: definition_title
       canHaveNegation: @model.canHaveNegation()
+      isNegated: @model?.get('negation')
```
pkmitre: Why the existential operator (?) here? I don't think it's needed.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -204,6 +272,12 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
     type = @$(e.target).model().get('type')
     $(".#{type}-elements").focus()
 
+  getNegationCodeIDAsText: (value_set_oid) ->
+    if @model.measure()?.valueSets().where(oid: value_set_oid).length > 0
+      return @model.measure()?.valueSets().where(oid: value_set_oid)[0].attributes.display_name
```
pkmitre: I think we can drop the existential operator here too, particularly since if it's used the expressions will still blow up because the chaining continues a few steps beyond (ie if there's no measure() the where() will blow up)

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -15,7 +15,7 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
       @measureRibbon = new Thorax.Views.MeasureRibbon model: @model
     @editCriteriaCollectionView = new Thorax.CollectionView
       collection: @model.get('source_data_criteria')
-      itemView: (item) => new Thorax.Views.EditCriteriaView(model: item.model, measure: @measure)
+      itemView: (item) => new Thorax.Views.EditCriteriaView(model: item.model, measure: @measure, builderView: @)
```
pkmitre: Because EditCriteriaView is a child of BuilderView, you shouldn't need to pass builderView in, you should be able to access it via the @patientBuilder() function.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+          @patientStatus.patientAge = @getDuration(moment(@model.get('birthdate') * 1000).add(5, 'hours'), moment(new Date(bonnie.measurePeriod, 0,1,0,0,0,0)))
+
+  setPatientAge: ->
+    @getPatientAge() # Determines Patient Age and stores results in instanced hash @patientStatus
+    patientAgeLabelSelector = @$('.patient-age-label')
+    patientAgeSelector = @$('.patient-age')
+
+    if @patientStatus.patientAge == null # Meaning the Birthdate came after the deathdate or the measurePeriod
+      if @patientStatus.patientIsAlive # Changes Warning Message based on whether the patient is alive or not
+        patientAgeLabelSelector.removeClass('fail fa fa-fw fa-times-circle')
```
thecristen: removing and adding the `fa` and `fa-fw` classes is redundant; you can leave them out of this code and the classes will be left alone.

app/assets/stylesheets/builder.less
```diff
@@ -422,3 +439,24 @@
...
+
+.hr-with-thin-margins {
+    margin-top: 7px;
+    margin-bottom: 7px;
+    border: 0;
+    border-top: 1px solid #e6e6e6;
+}
+
+.btn-width-100-percent {
+  width:100%;
```
thecristen: There's a class in Bootstrap that does this! :D It's `.btn-block`.

app/assets/stylesheets/mixins.less
```diff
@@ -20,6 +20,9 @@
   color: @brand-primary;
 }
 
+.warning {
```
thecristen, line 23: You could probably eliminate this and use Bootstrap's `.text-warning` class to achieve this.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -77,7 +82,7 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
 
     rendered: ->
       @$('.draggable').draggable revert: 'invalid', helper: 'clone', appendTo: '.patient-builder', zIndex: 10
-
+      @setPatientAge() #Clicking the button that removes Death Date doesn't trigger materialize, it triggers "render"
```
pkmitre: It seems like the lack of a materialize when removing the death date is an issue, shouldn't it materialize?

app/assets/javascripts/templates/patient_builder/patient_builder.hbs
```diff
@@ -142,16 +148,47 @@
...
   </div>
   <div class="col-center criteria-container droppable">
     <h1 class="heading-primary">Patient History</h1>
+    <h2 class="heading-muted">
+
+      {{#button "showOrHidePatientBuilderUtilities" class="btn btn-sm btn-link"}}<i class="fa fa-fw fa-lg fa-cog large-cog"> </i>
+        <span class="sr-only">Patient Builder Utilities</span>
+      {{/button}}
+      {{#button "openAllDataCriteria" id="open_all" class="btn btn-sm btn-primary hidden"}}Open All
+        <span class="sr-only">Open all Data Criteria</span>
```
thecristen: FYI, the screen reader will read both the visible "Open all" and the hidden "Open all" from "Open all Data Criteria". We probably just need the sr-only span to say "data criteria"

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -139,6 +144,7 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
 
   materialize: ->
     @serializeWithChildren()
+    @setPatientAge() # When a birthdate or deathdate is selected, materialize() is called and sets the Patient's Age
```
pkmitre: Rather than hooking into materialize(), which happens every time any change is made to the patient record, it might make more sense to just set the patient age specifically when the birth date changes.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
thecristen: I second the use two functions idea..

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+            if durationInformationValue > 1
+              durationInformation.durationInformationKey = "#{durationInformationValue} #{durationInformationKey}"
+            else
+              durationInformation.durationInformationKey = "#{durationInformationValue} #{durationInformationKey.slice(0, -1)}" # if necessary, remove "s" for non-plural
+            elementDurationAsString += "#{durationInformation.durationInformationKey}, "
+            displayTwoValues++
+      elementDurationAsString = elementDurationAsString.slice(0,-2) # Chop off the trailing ", " from the string
+    else
+      elementDurationAsString = null # meaning the end date preceded the start date
+    return elementDurationAsString
```
pkmitre: This code could be much simpler. If this is the behavior we want, we could probably start with something like

```
getDuration: (start_date, end_date) ->
  duration = moment.duration(end_date.diff(start_date))
  return null if duration < 0
  terms = []
  for unit in ['years', 'months', 'days', 'hours', 'minutes']
    terms.push "#{duration.get(unit)} #{unit}" if duration.get(unit) > 0
    break if terms.length > 1
  return terms.join(', ')
```

(not fully tested)

If we're willing to simplify the output, we could even just use the humanize() method that moment.js provides for durations.

pkmitre: Also, it would be nice if we had the pluralization correctly match the number, ie 1 year vs 2 years

DPC15038: It currently has this feature

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+        # Since both birthdate and deathdate are off by 5 hours, they will calculate correctly without the ".add(5, 'hours')". However, better to account for it
+        @patientStatus.patientAge = @getDuration(moment(parseInt(@model.get('birthdate')) * 1000).add(5, 'hours'), moment(parseInt(@model.get('deathdate')) * 1000).add(5, 'hours'))
+      else
+        @patientStatus.patientIsAlive = true
+        if parseInt(moment(@model.get('birthdate') * 1000).format("YYYY")) == bonnie.measurePeriod
+          @patientStatus.patientAge = "Born during Measure Period" #Patient born during Measure Period
+        else
+          # Time picker wants to offset birthdate by 5 hours - not sure if problem with timepicker or Moment.js
+          # either way, adding 5 hours to birthdate fixes problem
+          @patientStatus.patientAge = @getDuration(moment(@model.get('birthdate') * 1000).add(5, 'hours'), moment(new Date(bonnie.measurePeriod, 0,1,0,0,0,0)))
```
pkmitre: The 5 hour offset is likely due to a timezone difference, ie EST vs GMT; the right way to handle that is to use moment.utc (search for that in the patient builder code).

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,181 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
...
+            else
+              durationInformation.durationInformationKey = "#{durationInformationValue} #{durationInformationKey.slice(0, -1)}" # if necessary, remove "s" for non-plural
+            elementDurationAsString += "#{durationInformation.durationInformationKey}, "
+            displayTwoValues++
+      elementDurationAsString = elementDurationAsString.slice(0,-2) # Chop off the trailing ", " from the string
+    else
+      elementDurationAsString = null # meaning the end date preceded the start date
+    return elementDurationAsString
+
+  getPatientAge: ->
```
pkmitre: The approach used in the method of building a big global status object to hold patient status that gets used elsewhere has a few issues (it's a lot like a global variable, it doesn't separate concerns, there's computation being done on model data in the view that would be better done in the model). A better approach would be to create individual functions on the patient model like isAlive() and age(), and just call those functions where needed (ie passing the context from the view to the template). Most of the code should be in the model and the layout, with likely only a bit of code in the view.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
pkmitre: Missed that! My suggested code would still need to be tweaked...

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
DPC15038: I did some testing and here are the results

screen shot 2016-07-21 at 1 10 31 pm

pkmitre: There's still a fundamental issue with the approach of "adding 5 seems to fix the problem". If Bonnie is installed in a different timezone, will this still work? We need to figure out what's going on and figure out a solution that doesn't use a "magic number".

DPC15038: Ooh I forgot about timezones, I'll look more into this, and come up with a dynamic solution

app/assets/stylesheets/mixins.less
DPC15038, line 23: The top is with .text-warning, and the bottom is what I currently have. Thoughts?
screen shot 2016-07-22 at 8 39 44 am

app/assets/javascripts/models/patient.js.coffee
DPC15038: Working on a solution now...
image

DPC15038:
Having weird computer issues, so pushing up changes now - No need to review these currently, many things still incomplete

app/assets/javascripts/models/data_criteria.js.coffee
```diff
@@ -35,7 +35,7 @@ class Thorax.Collections.MeasureDataCriteria extends Thorax.Collection
 # multiple criteria with the same ID.
 class Thorax.Models.PatientDataCriteria extends Thorax.Model
   idAttribute: null
-  
+
```
pkmitre: It looks like the only changes in this file are white space.

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -29,7 +29,18 @@ class Thorax.Models.Patient extends Thorax.Model
 
     new @constructor JSON.parse(json), parse: true
 
-  getBirthDate: -> new Date(@get('birthdate'))
+  getBirthDate: -> moment(moment(moment.unix(@get('birthdate'))).utc().format("YYYY-MM-DD HH:mm"))
```
pkmitre: This seems somewhat convoluted, what's the goal of this line?

app/assets/javascripts/models/data_criteria.js.coffee
DPC15038: One of my Atom packages auto removes trailing whitespace on save. Is this a problem? Should I revert the file?

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -29,7 +29,18 @@ class Thorax.Models.Patient extends Thorax.Model
...
+  isAlive: ->
+    if @get('expired')
+      false
+    else
+      true
```
pkmitre: This could be simplified to

```
!@get('expired')
```

app/assets/javascripts/models/patient.js.coffee
DPC15038: Formats the Date for Moment JS, and accounts for the timezone issue I found earlier. I'll see if I can refactor it a bit

app/assets/javascripts/models/result.js.coffee
```diff
@@ -188,7 +188,7 @@ class Thorax.Models.Result extends Thorax.Model
   # walk through data criteria to account for specific occurrences within a UNION
   calculateDataCriteriaOrCounts: (rationale) ->
     orCounts = {}
-    for key, dc of @measure.get('data_criteria') when dc.derivation_operator == 'UNION' && key.indexOf('UNION') != -1
+    for key, dc of @measure.get('data_criteria') when dc.derivation_operator == 'UNION' && key.match(/UNION|satisfiesAny/)
```
pkmitre: I'm pretty sure this change is on master, we should look into why it's showing up in the diff for your pull request

DPC15038: I rebased with master

app/assets/javascripts/utilities.js.coffee
```diff
@@ -0,0 +1,13 @@
+window.Bonnie ||= {}
+Bonnie.util ||= {}
+Bonnie.util.getDurationBetween = (start_date, end_date) ->
```
pkmitre, line 3: We should check the case and be consistent, I think I've seen Bonnie and bonnie

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -79,6 +78,24 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
...
+    )
+    @listenTo(@parentView, "hideInformationinDataCriteria", ->
+      @showElementInformation = false
+      @render()
+    )
+    @listenTo(@parentView, "closeAllDataCriteria", ->
+      @render() # (Re)rendering the view causes each specific data criteria to close
+    )
+    @listenTo(@parentView, "openAllDataCriteria", ->
+      e = arguments[0] # e is only passed within the arguments[] object
```
pkmitre: This should just be

```
@listenTo(@parentView, "openAllDataCriteria", (e) ->
```

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -238,6 +244,58 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
       top: shiftDown
       bottom: $logic.nextAll(':visible').height() || 0
 
+  # Allows us to remember state after sorting by name or date
+  displayElementInformation: ->
+    if @$('#previewPatientInformation').is(":checked")
+      true
+    else
+      false
```
pkmitre: We should avoid using the DOM to store application state.

(Also, avoid explicit references to true and false)

app/assets/javascripts/models/data_criteria.js.coffee
pkmitre: It makes code reviews more difficult; revert this file (since there are no actual code changes) and adjust atom to not automatically remove trailing white space

app/assets/javascripts/models/result.js.coffee
pkmitre: If code you didn't write is showing up in your pull request that means there's an issue with how you rebased; we can take a look later today

app/assets/javascripts/models/data_criteria.js.coffee
```diff
@@ -96,6 +96,12 @@ class Thorax.Models.PatientDataCriteria extends Thorax.Model
     criteriaType in ['diagnostic_study_performed', 'functional_status_performed', 'intervention_performed', 'laboratory_test_performed',
                      'physical_exam_performed', 'procedure_performed', 'risk_category_assessment']
 
+  getNegationCodeText: ->
+    if @get('negation')
+      @.measure().valueSets().where(oid: @get('negation_code_list_id'))[0].get('display_name')
```
pkmitre: Should be @measure instead of @.measure

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -29,7 +29,15 @@ class Thorax.Models.Patient extends Thorax.Model
 
     new @constructor JSON.parse(json), parse: true
 
-  getBirthDate: -> new Date(@get('birthdate'))
+  # Returns the BirthDate formatted as a moment from Moment JS
+  getBirthDate: -> moment(moment.unix(@get('birthdate')).utc().format("YYYY-MM-DD HH:mm"))
```
pkmitre: I think this can be simplified, take a look at the way moment is used elsewhere in the code to deal with UTC and unix time. I think this could just be something like

```
moment.utc(@get('birthdate'), 'X')
```

If it's not quite right grab me and we can sit down and figure out what does make sense.

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -29,7 +29,15 @@ class Thorax.Models.Patient extends Thorax.Model
...
+  isAlive: -> !@get('expired')
+  getDeathDate: ->
+    if @isAlive()
+      null
+    else
+      return moment(moment(moment.unix(@get('deathdate'))).utc().format("YYYY-MM-DD HH:mm"))
```
pkmitre: See moment comment above.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -15,7 +15,7 @@ class Thorax.Views.PatientBuilder extends Thorax.Views.BonnieView
       @measureRibbon = new Thorax.Views.MeasureRibbon model: @model
     @editCriteriaCollectionView = new Thorax.CollectionView
       collection: @model.get('source_data_criteria')
-      itemView: (item) => new Thorax.Views.EditCriteriaView(model: item.model, measure: @measure)
+      itemView: (item) => new Thorax.Views.EditCriteriaView(model: item.model, measure: @measure, parentView: @)
```
pkmitre: We shouldn't need to pass the parent view. `EditCriteriaView` is an instance of `BuilderChildView`, which has a method called `patientBuilder()`

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -79,6 +79,23 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
     @editFulfillmentHistoryView = new Thorax.Views.MedicationFulfillmentsView
       model: new Thorax.Model
       criteria: @model
+    @previewElementInformation = @parentView.previewElementInformation # Allows us to remember state after sorting
```
pkmitre: This should just be set in the context rather than as an instance variable... that simplifies the code below quite a bit, to

```
@listenTo @parentView, "previewInformationinDataCriteria", => @render()
```

(Note that I'm changing `->` to `=>`, if I remember correctly `@listenTo` doesn't require `=>` to bind correctly to the current class but it's somewhat more explicit to use it).

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -79,6 +79,23 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
...
+
+    @listenTo(@parentView, "previewInformationinDataCriteria", ->
+      @previewElementInformation = true
+      @render()
+    )
```
pkmitre: The coffeescript style we use avoids parentheses in these cases, ie

```
@listenTo @parentView, "previewInformationinDataCriteria", =>
    @previewElementInformation = true
    @render()
```

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -251,3 +304,49 @@ class Thorax.Views.BuilderPopulationLogic extends Thorax.LayoutView
     _(super).extend
       title: if @model.collection.parent.get('populations').length > 1 then (@model.get('title') || @model.get('sub_id')) else ''
       cms_id: @model.collection.parent.get('cms_id')
+
+class Thorax.Views.PatientAge extends Thorax.Views.BuilderChildView
+  template: JST['patient_builder/patient_age']
+
+  initialize: ->
+    #We shouldn't need to listen for the change event, but Thorax doesn't seem to be updating the model properly
```
pkmitre: Our code style includes a space between the # and the text, ie

```
# We shouldn't need to listen for the change event, but Thorax doesn't seem to be updating the model properly
```

pkmitre: Also, the comment isn't quite right, it should read more like

```
# We shouldn't need to listen for the change event, but Thorax doesn't seem to be updating the view given the change event
```

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -251,3 +304,49 @@ class Thorax.Views.BuilderPopulationLogic extends Thorax.LayoutView
...
+    @model.on 'change', =>  @render()
```
pkmitre: It's worth being consistent with spacing, ie

```
@model.on 'change', => @render()
```

instead of

```
@model.on 'change', =>  @render()
```

It doesn't have a big impact here, but there are definitely cases where it affects readability

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -251,3 +304,49 @@ class Thorax.Views.BuilderPopulationLogic extends Thorax.LayoutView
...
+
+
+  context: ->
+    if @model.isAlive()
+    # if the birthdate/deathdate fields are left blank they default to the epoch
+      unless @model.getBirthDate().isSame(moment("1970-01-01 00:00"))
```
pkmitre: I'm assuming that `getBirthdate()` is returning this date if it's `null`? That's not a great practice, what about a patient that actually has that birthdate? If the birthdate is `null`, `getBirthdate()` should probably return `null`.

app/assets/javascripts/views/patient_builder/patient_builder.js.coffee
```diff
@@ -251,3 +304,49 @@ class Thorax.Views.BuilderPopulationLogic extends Thorax.LayoutView
...
+          faIcon = "fa fa-fw fa-exclamation-circle"
+          patientAge = "Measure Period Precedes Date of Birth!"
+        else
+          errorWithDates = false
+          importantDate = false
+          faIcon = null
+          patientAge = "Age at Start of Measure Period: #{Bonnie.util.getDurationBetween(@model.getBirthDate(), moment([bonnie.measurePeriod]))}"
+    else
+      unless @model.getBirthDate().isSame(moment("1970-01-01 00:00")) || @model.getDeathDate().isSame(moment("1970-01-01 00:00"))
+        if Bonnie.util.getDurationBetween(@model.getBirthDate(), @model.getDeathDate())
```
pkmitre: This condition uses an implicit property of the `getDurationBetween()` method (ie returns null if the duration would be negative), which makes the code trickier to understand than it needs to be. Why not

```
if @model.getBirthDate() < @model.getDeathDate()
```

DPC15038:
@pkmitre I'd like to sit down with you and go over some of your comments/ask a few questions before you take another look

thecristen:
"End Date precedes Start Date" seems to be shown on a data element even when the end date is set equal to the start date

edeyoung:
I think the measure request to include icons for outdated codes might be a good opportunity to merge this PR in.

I think some of the UI should change slightly. See my screen shot below and the specific changes I'm interested in:

image

1. Make "Order By" a drop down box in case we want to add more options. Also, less visual clutter.
2. Remove "collapse all" and "expand all". With the preview information, this seems like less of a necessity.
3. Change wording and placement of the "show details" check box.
4. Include codes in the detail view.
5. (in expanded view) include time duration

DPC15038:
~~Oh god did I just include all the commits from everything...Should I cherry pick just mine, and worry about merging/git issues later?~~ See comment below.

c-monkey:
I think cherry picking does seem wise!

DPC15038:
Cherrypicked the correct commits, and force pushed to remove any unwanted commits in this PR.

PR #539, opened 2016-06-08
Remove extra padding from list inside a set operator nested within a …

thecristen:
…listed set operator.

Yeah!

This is a fix for JIRA issue 166 that affects several measures - CMS171v6, CMS185v5, CMS105v5, and CMS190v5 (bad indenting seen below with red for emphasis). 

image

This is caused by an extraneous `<ul>` element. The approach taken was to use CSS to target the extraneous element and remove the left padding.

### DIFF
--- a/app/assets/stylesheets/measure.less
+++ b/app/assets/stylesheets/measure.less
@@ -210,6 +210,10 @@
       li {
         .set-op-item {
           list-style-type: square;
+
+          .set-op-group > ul {
+            padding-left: 0; // addresses overly-nested list markup
+          }
         }
       }
     }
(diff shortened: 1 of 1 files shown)


edeyoung:
Nice!

It looks like the same thing is happening in the logic view in the patient dashboard: 
image
This must be under a different style hierarchy.

If there's an extra `<ul>`, should we change how the code is generated so this doesn't happen, or is the extra `<ul>` necessary?

thecristen:
Good catch, I'll look into that view!

IMO we should probably code up [the template](https://github.com/projecttacoma/bonnie/blob/master/app/assets/javascripts/templates/logic/data_criteria.hbs#L8) more intelligently, but right now it's painfully confusing to parse and much easier to try to target it with CSS..

PR #541, opened 2016-06-20
File rename in order for rake to work with new version of Bundler

mayerm94:
Rename test/simplecov.rb to test/simplecov_init.rb so that bundle exec rake works with bundler version 1.12.5


hossenlopp:
This makes Rake tests work with bundler 1.12.x. I tested this with bundler 1.11.2 and all was good. This also updates the simplecov library. Everything looks in order to me. @pkmitre, @edeyoung should we merge this? If we hold off then new developers will have to install bundler 1.11.2 to be able to run unit tests.

pkmitre:
@hossenlopp, thanks for taking a look, merged

PR #548, opened 2016-06-24
made it so the query for code concepts will return an empty array if …

edeyoung:
…no code concepts exist.

This is related to a bug seen in dogpark related to CMS172. When tested locally, if this measure was loaded using "Published", the code broke because of the empty concept list. This addresses that issue.


hossenlopp:
This looks like a quick fix for the issue of VSAC giving empty valuesets sometimes. Future fixes should include warnings for empty sets. We should also revamp our options the users get and how we interact with VSAC.

PR #549, opened 2016-06-30
Controller tests

mayerm94:
Added various controller tests and fixtures that stress different error-checks, added VCR gem for automatic VSAC auth in tests.


config/initializers/webmock.rb
```diff
@@ -0,0 +1,4 @@
+# Disable WebMock globally so it allows VSAC to work outside of tests
```
edeyoung: I'd add more wording to this comment. Maybe something like,

"WebMock is used to test the ~~MAT API~~ VSAC API. WebMock is disabled globally here. It is only enabled in the testing context."

Also, I feel like there should be a better way to handle this (but I don't know what that would be). @hossenlopp - any thoughts on how to make WebMock only enabled in the testing context?

test/fixtures/vcr_cassettes/invalid_vsac_response.yml
```diff
@@ -0,0 +1,101 @@
...
+      X-Vip-Info:
+      - 130.14.16.40:443
+      X-Pool-Info:
+      - "/Common/vsacweb_p 10.1.5.110 8080"
+    body:
+      encoding: UTF-8
+      string: ''
+    http_version: 
+  recorded_at: Tue, 28 Jun 2016 13:40:45 GMT
+- request:
```
edeyoung: what is the difference between the two requests?

test/fixtures/vcr_cassettes/invalid_vsac_response.yml
```diff
@@ -0,0 +1,101 @@
+---
```
edeyoung, line 3: Add some comments describing the purpose of this file and how/where it gets used

test/functional/admin_users_controller_test.rb
```diff
@@ -46,7 +46,7 @@ class Admin::UsersControllerTest  < ActionController::TestCase
     assert_response :success
     @user_unapproved.reload
     assert_equal true, @user_unapproved.approved?
-
+    #Verify transmission of new user email
```
edeyoung: For all comments, add a space after the initial `#`.

We try to conform to the style guidance outlined here: https://github.com/bbatsov/ruby-style-guide

test/functional/application_controller_test.rb
```diff
@@ -0,0 +1,20 @@
...
+class ApplicationControllerTest < ActionController::TestCase
+  include Devise::TestHelpers
+  setup do
+    dump_database
+    
+  end
+
+  test "page not found" do
+    get :page_not_found
+    assert_equal 404,response.status
```
edeyoung: Style fix: add spaces after all commas

config/initializers/webmock.rb
pkmitre: Perhaps just including webmock in the test and ci groups in the Gemfile

hossenlopp: If you want to run bonnie locally in development mode with all the gems for testing and have VSAC work, this may be needed. I'm surprised that WebMock even enables itself by default.

test/functional/application_controller_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'test_helper'
+
```
edeyoung, line 2: Only a few of the methods in `ApplicationController` are included here. Is it worth including a comment that the other functions are covered through other calls? What about including direct unit tests for those functions even though they're covered indirectly?

config/initializers/webmock.rb
hossenlopp: Looking at the WebMock github readme, it looks like it shouldn't be enabled by default. @mayerm94 have you given it a try without this file?

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,25 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
-  #   put :log_in_as, {id: @user_plain.id}
-  #   assert_response :redirect
-  # end
+  test "sign in as" do
+     sign_in @user_admin
+     pre_count = @user_plain.sign_in_count
+     post :log_in_as, {id: @user_plain.id}
+     assert_response :redirect
+     @user_plain.reload
+     assert_equal pre_count +1 ,@user_plain.sign_in_count
```
edeyoung: Style comment: watch the spacing. Should be: `assert equal pre_count + 1, @user_plain.sign_in_count`

test/functional/measures_controller_test.rb
```diff
@@ -14,6 +15,90 @@ class MeasuresControllerTest  < ActionController::TestCase
     sign_in @user
   end
 
+  test "upload xml with valid VSAC creds" do
```
edeyoung, line 18: To help out future developers, add comments here explaining how the VCR cassettes work for testing VSAC.

test/functional/measures_controller_test.rb
```diff
@@ -14,6 +15,90 @@ class MeasuresControllerTest  < ActionController::TestCase
...
+
+  test "force expire vsac session" do
+    #The ticket field was taken from the vcr_cassettes/valid_vsac_response file
+    session[:tgt] = {ticket: "ST-67360-HgEfelIvwUQ3zz3X39fg-cas", expires: Time.now + 27000}
+    get :vsac_auth_expire
+
+    assert_response :ok
+    assert_equal " ", response.body
+
+    assert_nil session[:tgt]
```
edeyoung, line 105: should this also include:

```
    get :vsac_auth_valid

    assert_response :ok
    assert_equal false, JSON.parse(response.body)['valid']
```

test/functional/measures_controller_test.rb
```diff
@@ -55,6 +140,93 @@ class << measure_file
...
+  test "upload measure already loaded" do
+    sign_in @user
+    measure_file = fixture_file_upload(File.join('test','fixtures','measure_exports','measure_initial.zip'),'application/zip')
+    class << measure_file
+      attr_reader :tempfile
+    end
+    post :create, {measure_file: measure_file, measure_type: 'eh', calculation_type: 'episode'}
+    assert_response :redirect
+    measure = Measure.where({hqmf_id: "40280381-3D27-5493-013D-4DCA4B826AE4"}).first
+    get :value_sets, {id: measure.id, format: :json}
```
edeyoung: above when you do a valid vsac auth upload, you check the existence of a measure by checking the measure's `hqmf_id` field. Why go through the `value_sets` here? If there's not a specific reason, consider making the existence check for measures consistent and just check the `hqmf_id` field.

test/functional/measures_controller_test.rb
```diff
@@ -55,6 +140,93 @@ class << measure_file
...
+    assert_response :success
+    assert_equal 29, JSON.parse(response.body).keys.count
+    #Now measure successfully uploaded, try to upload again
+    post :create, {measure_file: measure_file, measure_type: 'eh', calculation_type: 'episode'}
+    assert_equal "Error Loading Measure",flash[:error][:title]
+    assert_equal "A version of this measure is already loaded.", flash[:error][:summary]
+    assert_equal "You have a version of this measure loaded already.  Either update that measure with the update button, or delete that measure and re-upload it.", flash[:error][:body]
+    assert_response :redirect
+    #Verify measure has not been deleted or modified
```
edeyoung: this isn't checking whether or not the measure has been modified (other than the count for value sets). Reword the comment to `# Verify measure has not been deleted`

If you want to check if the measure has changed, I think you can convert it to json. If that's the case, you can do a comparison between the two json representations of the measure (after the first upload and after the second upload attempt).

Also, it looks like the exact same measure is being uploaded both times instead of different versions of the same measure. If you want to make sure the measure hasn't changed, I'd recommend uploading a different version of the same measure as your second upload.

test/functional/measures_controller_test.rb
```diff
@@ -55,6 +140,93 @@ class << measure_file
...
+
+  test "upload measure already loaded" do
+    sign_in @user
+    measure_file = fixture_file_upload(File.join('test','fixtures','measure_exports','measure_initial.zip'),'application/zip')
+    class << measure_file
+      attr_reader :tempfile
+    end
+    post :create, {measure_file: measure_file, measure_type: 'eh', calculation_type: 'episode'}
+    assert_response :redirect
+    measure = Measure.where({hqmf_id: "40280381-3D27-5493-013D-4DCA4B826AE4"}).first
```
edeyoung: I'd add a `nil` check for this `Measure.where` statement before the first upload. This will show that the measure that is uploaded actually corresponds to this `hqmf_id`.

test/functional/measures_controller_test.rb
```diff
@@ -55,6 +140,93 @@ class << measure_file
...
+    #Verify measure has not been deleted or modified
+    measure = Measure.where({hqmf_id: "40280381-3D27-5493-013D-4DCA4B826AE4"}).first
+    get :value_sets, {id: measure.id, format: :json}
+    assert_response :success
+    assert_equal 29, JSON.parse(response.body).keys.count
+  end
+
+  test "update with hqmf set id mismatch" do
+    #Upload the initial file
+    measure_file = fixture_file_upload(File.join('test','fixtures','measure_exports','measure_initial.zip'),'application/zip')
```
edeyoung: to show that you're actually uploading the measure with the referenced `hqmf_id` and `hqmf_set_id` below, consider doing an initial `nil` check, then after the first upload, do check that the measure's `hqmf_id` and `hqmf_set_id` match what you use later.

test/functional/measures_controller_test.rb
```diff
@@ -55,6 +140,93 @@ class << measure_file
...
+  test "upload missing value set missing oid" do
+    measure_file = fixture_file_upload(File.join('test','fixtures','measure_exports','measure_initial_valueset_oid_mismatch.zip'),'application/zip')
+    class << measure_file
+      attr_reader :tempfile
+    end
+    post :create, {measure_file: measure_file, measure_type: 'eh', calculation_type: 'episode'}
+
+    assert_equal "Measure is missing value sets",flash[:error][:title]
+    assert_equal "The measure you have tried to load is missing value sets.", flash[:error][:summary]
+    # The 'principal diagnosis' oid was changed in DischargedonAntithromboticThe_eMeasure.xml within measure_initial_valueset_oid_mismatch.zip from 2.16.840.1.113883.3.117.2.7.1.14 to 2.16.840.1.113883.3.117.2.7.1.15
```
edeyoung: I think this comment makes more sense on the line that does the `fixture_file_upload`.

test/functional/measures_controller_test.rb
```diff
@@ -73,7 +245,7 @@ class << measure_file
     class << measure_file
       attr_reader :tempfile
     end
-
+    #Initial create of measure
```
edeyoung: to add visual clarity to what's happening in this test (so a reader can automatically chunk the code and see the comment more easily), add a new line before the comment:

```
     class << measure_file
        attr_reader :tempfile
      end

    #Initial create of measure
```

test/functional/patients_controller_test.rb
```diff
@@ -2,26 +2,38 @@
...
+
+    get :index
+    assert_equal 4, JSON.parse(response.body).count
+
+  end
+
   test "create" do
 
     assert_equal 0,Record.count
-    @patient = {'first'=> 'Betty',
```
edeyoung: You should turn off "whitespace" in Atom. With this on, it makes it look like files have been changed that haven't actually, and can clog pull requests.

To turn off whitespace in atom, go to Atom -> Preferences, click Packages, click the whitespace package, and unclick “Remove Trailing Whitespace”.

test/functional/registrations_controller_test.rb
```diff
@@ -37,4 +37,36 @@ class RegistrationsControllerTest < ActionController::TestCase
...
+    assert_equal "You have signed up successfully. However, we could not sign "+
+    "you in because your account is not yet approved.  You will receive an email"+
+    " once your account has been approved.", flash[:notice]
+  end
+
+  test "destroy with valid password" do
+    sign_in @user
+    delete :destroy, { user: {current_password: 'Test1234!'}}
+    assert_response :redirect
+    deluser = User.by_email('bonnie@example.com').first
```
edeyoung: I see looking above that `@user` has the email bonnie@example.com. However, this would be more convincing to me if the `@user` object were used to get `deluser`. For example (untested):

```
deluser = User.by_email(@user.email).first
```

Similar for test below.

test/vcr_setup.rb
```diff
@@ -0,0 +1,14 @@
+require 'vcr'
+
+VCR.configure do |c|
```
edeyoung: Consider adding a comment here to explain vcr and how/why it's being used.

config/initializers/webmock.rb
mayerm94: @hossenlopp I removed this file all together and ran tests that both did and did not require VCR creation and usage. Everything appears to work fine without the webmock init file. The fact that webmock is disabled by default along with the fact that webmock is only in the test,developement,ci group in the gemfile renders this file pointless.

test/fixtures/vcr_cassettes/invalid_vsac_response.yml
mayerm94: invalid_vsac_response.yml uses 'invalidusername' and 'invalidpassword' as credentials for VSAC whereas valid_vsac_response.yml uses the ENV[VSAC_USERNAME] and ENV[VSAC_PASSWORD] which are valid vsac credentials. The invalid_vsac_response.yml is a recording of unsuccessful vsac authentication, valid_vsac_response.yml is a recording of successful vsac authentication.

test/fixtures/vcr_cassettes/invalid_vsac_response.yml
mayerm94, line 3: I added comments about the purpose of the file at the top of the cassettes, but these comments will be lost should someone decide to re-record the cassette for whatever reason. Because of this, I also put the comments describing the purpose where the cassette is used as well.

test/functional/measures_controller_test.rb
mayerm94: Removed retrieval of value_sets, for they were not used after being retrieved.

test/functional/measures_controller_test.rb
mayerm94: Changed the test to upload a different version of the measure and asserted that the JSON representations of the measure were the same before attempting to upload and after per your suggestions.

test/fixtures/vcr_cassettes/invalid_vsac_response.yml
edeyoung: I meant the two requests inside `invalid_vsac_response.yml`. I guess maybe the answer is that it's a recording, and if it has duplicates there, that's just how it recorded?

hossenlopp: I think this happens since there could be more than one test that tries using an invalid password. I don't think theres a way to have it not record these near duplicate requests.

test/functional/application_controller_test.rb
mayerm94, line 2: Of the 3 functions explicitly tested in ApplicationController, 2 are tested in ApplicationControllerTest I added a comment describing which test file the third function is tested in and what the purpose of said test is.

test/fixtures/vcr_cassettes/invalid_vsac_response.yml
edeyoung, line 3: Great! I didn't understand that the yml files were the actual recordings.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,25 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
-  #   put :log_in_as, {id: @user_plain.id}
-  #   assert_response :redirect
-  # end
+  test "sign in as" do
+     sign_in @user_admin
+     pre_count = @user_plain.sign_in_count
+     post :log_in_as, {id: @user_plain.id}
+     assert_response :redirect
+     @user_plain.reload
+     assert_equal pre_count +1 , @user_plain.sign_in_count
```
edeyoung: Take a look at the spacing here.
Should be `assert equal pre_count + 1, @user_plain.sign_in_count`

edeyoung:
This looks really good! I think the only outstanding comment is a styling issue that I just added a comment on. Once that gets addressed, I'll add this into master!

test/functional/admin_users_controller_test.rb
```diff
@@ -139,4 +139,47 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    assert_equal mail[-2].to.first, @user_plain.email
+    assert_equal mail[-3].to.first, @user_admin.email
+    assert_equal mail[-4].to.first, @user.email
+  end
+
+  test "email active" do
+    sign_in @user_admin
+    original_last_sign_in_time = @user_admin.last_sign_in_at
+    original_measure_count = @user_admin.measure_count
+    #=-=-=-=-=-=-=|The following tests greater than 6 months|-=-=-=-=-=-=-=-
```
mayerm94: Style guide suggests a space between # and the comment https://github.com/bbatsov/ruby-style-guide

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
-  #   put :log_in_as, {id: @user_plain.id}
-  #   assert_response :redirect
-  # end
+  test "sign in as" do
+     sign_in @user_admin
+     pre_count = @user_plain.sign_in_count
+     post :log_in_as, {id: @user_plain.id}
+     assert_response :redirect
+     @user_plain.reload
+     assert_equal pre_count +1, @user_plain.sign_in_count
```
edeyoung: Please add a space after the `+`.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    assert_equal mail[-2].to.first, @user_plain.email
+    assert_equal mail[-3].to.first, @user_admin.email
+    assert_equal mail[-4].to.first, @user.email
+  end
+
+  test "email active" do
+    sign_in @user_admin
+    original_last_sign_in_time = @user_admin.last_sign_in_at
+    original_measure_count = @user_admin.measure_count
+   # =-=-=-=-=-=-=|The following tests greater than 6 months|-=-=-=-=-=-=-=-
```
edeyoung: The comments should just include the comment text. To visually differentiate between sections, add whitespace rather than text separators.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+  test "email all" do
+    sign_in @user_admin
+    assert_equal @user_admin.admin?, true
+    post :email_all, {subject: "Test Email All Subject", body: "email all body", format: :json}
+    mail = ActionMailer::Base.deliveries
+    assert_equal "Test Email All Subject", mail.last.subject
+    assert_equal mail[-1].to.first, @user_unapproved.email
+    assert_equal mail[-2].to.first, @user_plain.email
+    assert_equal mail[-3].to.first, @user_admin.email
+    assert_equal mail[-4].to.first, @user.email
```
edeyoung: Are negative indices necessary here? What about: 

```
assert_equal mail[0].to.first, @user.email
...
```

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+  test "email active" do
+    sign_in @user_admin
+    original_last_sign_in_time = @user_admin.last_sign_in_at
+    original_measure_count = @user_admin.measure_count
+   # =-=-=-=-=-=-=|The following tests greater than 6 months|-=-=-=-=-=-=-=-
+    ActionMailer::Base.deliveries = [] # reset the list of email deliveries to ensure clean slate
+    assert_equal @user_admin.last_sign_in_at, User.where(email: @user_admin.email).first.last_sign_in_at
+    post :email_active, {subject: "Example Subject for Testing", body: "test body of email", format: :json}
+    mail = ActionMailer::Base.deliveries
+    assert_equal mail, [] # There shouldn't be any new mail because this patient is older than 6 months
```
edeyoung: Consider explicitly making the last sign in date more than six months ago for this user like it's done below to ensure that it's within six months. This will make it clear to future users and will also make this test robust to changes to the underlying user data.

It also looks like there's an assumption that all other users have last login dates more than six months ago. This would be good to set explicitly for the same reason.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    # ==-=--=-=|The following tests fewer than 6 months, but 0 measures|-=-=-
+    @user_admin.last_sign_in_at = Date.today - 1.months
+    @user_admin.save! # update the database with the test values
+    ActionMailer::Base.deliveries = [] # reset the list of email deliveries to ensure clean slate
+    post :email_active, {subject: "Example Subject for Testing", body: "test body of email", format: :json}
+    mail = ActionMailer::Base.deliveries #
+    assert_equal mail, [] # There shouldn't be any mail because, although the patient is younger than 6 months, they have 0 measures associated with them
+    # ==-=--=-=|The following tests fewer than 6 months, and 3 measures|-=-=-
+    associate_user_with_measures(@user_admin, Measure.all)
+    associate_user_with_patients(@user_admin, Record.all)
```
edeyoung: Adding in the `Records` adds in a confounding factor. The `email_active` should work even with no patients. I'd remove this line to clarify that `email_active` is actually acting as expected and doesn't have requirements that records exist.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    @user_admin.save! # update the database with the test values
+    ActionMailer::Base.deliveries = [] # reset the list of email deliveries to ensure clean slate
+    post :email_active, {subject: "Email Sent!", body: "test body of email", format: :json}
+    mail = ActionMailer::Base.deliveries
+    assert_equal mail.last.subject, "Email Sent!" # This test should pass because the user has more than 0 measures, and is younger than 6 months
+
+    @user_admin.last_sign_in_at = original_last_sign_in_time
```
edeyoung: because all tests are run in a fresh context, it not necessary to reset the data.

If you look at the top of the file, there's a `setup` method. This gets run before each test to setup the context.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    post :email_all, {subject: "Test Email All Subject", body: "email all body", format: :json}
+    mail = ActionMailer::Base.deliveries
+    assert_equal "Test Email All Subject", mail.last.subject
+    assert_equal mail[-1].to.first, @user_unapproved.email
+    assert_equal mail[-2].to.first, @user_plain.email
+    assert_equal mail[-3].to.first, @user_admin.email
+    assert_equal mail[-4].to.first, @user.email
+  end
+
+  test "email active" do
```
edeyoung: It looks like you're handling:
- no measures, over six months
- no measures, within six months
- measures, within six months

Please look into also adding
- measures, over six months

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    mail = ActionMailer::Base.deliveries #
+    assert_equal mail, [] # There shouldn't be any mail because, although the patient is younger than 6 months, they have 0 measures associated with them
+    # ==-=--=-=|The following tests fewer than 6 months, and 3 measures|-=-=-
+    associate_user_with_measures(@user_admin, Measure.all)
+    associate_user_with_patients(@user_admin, Record.all)
+    @user_admin.save! # update the database with the test values
+    ActionMailer::Base.deliveries = [] # reset the list of email deliveries to ensure clean slate
+    post :email_active, {subject: "Email Sent!", body: "test body of email", format: :json}
+    mail = ActionMailer::Base.deliveries
+    assert_equal mail.last.subject, "Email Sent!" # This test should pass because the user has more than 0 measures, and is younger than 6 months
```
edeyoung: To keep the tests parallel, consider checking that the `to` field contains the admin user ~~rather than~~ as well as checking the subject. Also, check that there's only one delivery recipient.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    assert_equal @user_admin.last_sign_in_at, User.where(email: @user_admin.email).first.last_sign_in_at
+    post :email_active, {subject: "Example Subject for Testing", body: "test body of email", format: :json}
+    mail = ActionMailer::Base.deliveries
+    assert_equal mail, [] # There shouldn't be any new mail because this patient is older than 6 months
+    # ==-=--=-=|The following tests fewer than 6 months, but 0 measures|-=-=-
+    @user_admin.last_sign_in_at = Date.today - 1.months
+    @user_admin.save! # update the database with the test values
+    ActionMailer::Base.deliveries = [] # reset the list of email deliveries to ensure clean slate
+    post :email_active, {subject: "Example Subject for Testing", body: "test body of email", format: :json}
+    mail = ActionMailer::Base.deliveries #
```
edeyoung: there's a hanging hashmark here

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+     sign_in @user_admin
+     pre_count = @user_plain.sign_in_count
+     post :log_in_as, {id: @user_plain.id}
+     assert_response :redirect
+     @user_plain.reload
+     assert_equal pre_count +1, @user_plain.sign_in_count
+  end
+
+  test "email all" do
+    sign_in @user_admin
```
edeyoung: Consider doing something where you check who all the users are in the DB, and make sure that this list matches the delivery list. As written, this will break if the `setup` method changes at all and makes assumptions that these are the only patients in the test DB.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,68 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    assert_equal mail[-3].to.first, @user_admin.email
+    assert_equal mail[-4].to.first, @user.email
+  end
+
+  test "email active" do
+    sign_in @user_admin
+    original_last_sign_in_time = @user_admin.last_sign_in_at
+    original_measure_count = @user_admin.measure_count
+   # =-=-=-=-=-=-=|The following tests greater than 6 months|-=-=-=-=-=-=-=-
+    ActionMailer::Base.deliveries = [] # reset the list of email deliveries to ensure clean slate
```
edeyoung: tests are run in their own context and are not necessarily run sequentially. If resetting this array seems like it's necessary (it shouldn't be), then it should also be reset for the `"email all"` test.

test/functional/populations_controller_test.rb
```diff
@@ -2,14 +2,13 @@
 
 class PopulationsControllerTest  < ActionController::TestCase
   include Devise::TestHelpers
-      
+
```
edeyoung, line 5: looks like this file may have been missed for managing spaces and commas.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,111 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+     assert_equal pre_count + 1, @user_plain.sign_in_count
+  end
+
+  test "email all" do
+    sign_in @user
+    assert_equal false, @user.admin?
+    not_authorized = assert_raises RuntimeError do
+      post :email_all, {subject: "Test Email All Subject", body: "email all body", format: :json}
+    end
+    assert_equal "User #{@user.email} requesting resource requiring admin access", not_authorized.message
```
edeyoung: Should check that `ActionMailer::Base.deliveries` is empty.

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,111 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+          number_of_users_who_recieved_emails += 1
+          assert_equal "Test Email All Subject", mail[1].subject
+        when mail[2].to.first
+          number_of_users_who_recieved_emails += 1
+          assert_equal "Test Email All Subject", mail[2].subject
+        when mail[3].to.first
+          number_of_users_who_recieved_emails += 1
+          assert_equal "Test Email All Subject", mail[3].subject
+        end
+    end
```
edeyoung: There's an assumption here that only 4 people have received mail. What you want to check is that for all users, each user has received an email. If you check that and you check the length of the mail list against the number of users, you know that each user has been sent one email.

Consider something like this (untested):

```
users_sent_emails = User.all? do |user|
  ActionMailer::Base.deliveries.any? do |email|
    email.to.first == user.email
  end
end

assert_true users_sent_email
assert_true ActionMailer::Base.deliveries.length, User.count()
```

`all?` and `any?` are very nifty functions. Check them out here: http://ruby-doc.org/core-2.3.1/Enumerable.html

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,111 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    assert_equal number_of_users_who_recieved_emails, User.count()
+    assert_equal mail.length, User.count()
+  end
+
+  test "email active" do
+    @user.last_sign_in_at = Date.today - 8.months # arbitrary date more than 6 months ago
```
edeyoung: There's an assumption here on the users again.

Consider something like:

```
# make sure each user's last sign in is greater than 6 months
User.each do |user|
  user.last_sign_in_at = Date.today - 8.months
  user.save!
end
```

test/functional/admin_users_controller_test.rb
```diff
@@ -118,23 +118,111 @@ class Admin::UsersControllerTest  < ActionController::TestCase
...
+    assert_equal [], ActionMailer::Base.deliveries # There shouldn't be any mail because, despite having 3 measures, the patient is greater than 6 months
+
+    # The following tests fewer than 6 months, and 3 measures, so we should recieve an  email
+    @user_admin.last_sign_in_at = Date.today - 1.months
+    @user_admin.save! # update the database with the test values
+    post :email_active, {subject: "Email Sent!", body: "test body of email", format: :json}
+    mail = ActionMailer::Base.deliveries.last
+    assert_equal 1, mail.to.length # ensure that only one email was sent
+    assert_equal "Email Sent!", mail.subject # This test should pass because the user has more than 0 measures, and is younger than 6 months
+    assert_equal @user_admin.email, mail.to.first
```
edeyoung: You can test against all of the users by using this loop structure:

```
# make sure each user's last sign in is greater than 6 months
User.each do |user|
  user.last_sign_in_at = Date.today - 8.months
  user.save!
end
```

for each of these cases. For the final case where you expect to have emails sent, you can use a similar structure to `email_all`.

PR #556, opened 2016-07-27
updated demo script with a more recent measure.

edeyoung:


pkmitre:
Why not use your SimpleDischargedOnAntithrombotic measure?

PR #561, opened 2016-08-08
Istanbul integration

mayerm94:
In this branch, the jasmine gem has been replaced with the teaspoon and teaspoon-jasmine gems. the jasmine-jquery-rails is no longer required, because we have jasmine-jquery in spec/helpers.

The jasmine-jquery-2.1.0.js (which was actually was version 1.5.1 and improperly named) file in spec/helpers has been updates to jasmine-jquery-2.1.1.js.

spec/teaspoon_env.rb configures teaspoon, and code coverage options and spec/javascripts/spec_helper.js defines what resources need to be loaded in the particular order required for the jasmine specs to run.

The teaspoon test runner uses istanbul for code coverage, so you will need to install istanbul 'npm install -g istanbul' in order to generate coverage reports.

See the bottom of [(https://github.com/projecttacoma/bonnie/wiki/Guide-to-Bonnie-Rake-Tasks)] for teaspoon usage instructions.


spec/javascripts/models/result_spec.js.coffee
```diff
@@ -1,6 +1,7 @@
 describe 'Result', ->
 
   beforeEach ->
+    jasmine.getJSONFixtures().clearCache()
```
edeyoung, line 5: The need to do this is surprising to me. Can you explain in more detail why this needs to be cleared? Are we doing something wrong somewhere else? This seems like it should be automatically cleared between tests.

mayerm94, line 5: The data returned from getJSONFixtures is correct the first time it is called, but becomes jumbled during following requests. getJSONFixtures is likely not doing a deep copy, meaning that any modification to the loaded fixture is not 'reset' before the next call to getJSONFixtures

spec/javascripts/integration/navigation_spec.js.coffee
```diff
@@ -1,5 +1,7 @@
 describe 'Navigation', ->
 	beforeEach ->
+		# Clear the fixtures cache so that getJSONFixture does not return stale/modified fixtures
```
edeyoung: indentation here seems funky (in the unedited code too). Can you change this so the indentation is correct?

spec/javascripts/models/patient_spec.js.coffee
```diff
@@ -1,6 +1,7 @@
 describe 'Patient', ->
 
   beforeEach ->
+    jasmine.getJSONFixtures().clearCache()
```
edeyoung: Looks like there's a missing comment here for explaining `clearCache`

spec/javascripts/models/result_spec.js.coffee
edeyoung, line 5: missing comment here for explaining `clearCache`

spec/javascripts/views/complexity_view_spec.js.coffee
```diff
@@ -26,6 +28,7 @@ describe 'ComplexityVizView', ->
     measureSet1 = "54cad5fc69702d74b4000000"
     measureSet2 = "54cad74a69702d755a000000"
     @complexityView = new Thorax.Views.Dashboard
+    jasmine.getJSONFixtures().clearCache()
```
edeyoung: missing comment

hossenlopp:
When running the teaspoon rake task it does not find phantomjs. Do I need to manually install it?

.travis.yml
```diff
@@ -6,7 +6,7 @@ before_install:
   - "sh -e /etc/init.d/xvfb start"
 services: mongodb
 bundler_args: --without test development
-script: export DISPLAY=:99.0 && bundle exec rake jasmine:ci && bundle exec rake
+script: export DISPLAY=:99.0 && bundle exec teaspoon && bundle exec rake
```
hossenlopp: Shouldn't the teaspoon runner part be `bundle exec rake teaspoon`?

mayerm94: This is a trivial change, both commands result in the same action. I have changed it to bundle exec rake teaspoon for the sake of consistency.

edeyoung (approved):
Previously reviewed before github made it's review changes.

PR #562, opened 2016-08-10
Improved Code Coverages

DPC15038:


test/unit/user_test.rb
```diff
@@ -37,6 +37,29 @@ class UserTest < ActiveSupport::TestCase
...
+    @user.reload
+    assert_equal true, @user.is_dashboard?
+    @user.revoke_dashboard
+    @user.reload
+    assert_equal false, @user.is_dashboard?
+  end
+
+  test "grant and revoke dashboard_set" do
+    assert_equal false, @user.is_dashboard_set?
+    assert_equal false, @user.is_approved?
```
edeyoung, line 61: setting `@user.approved` also happens when the dashboard/admin/portfolio/etc. are set. Because this is how `approved` is set within `user.rb` (always as part of granting one of the other permissions), it is an intentional side effect (although one I don't really agree with), so we should check for it everywhere.

edit: by everywhere, I mean in all of the other tests for granting privileges.

test/unit/user_test.rb
```diff
@@ -37,6 +37,29 @@ class UserTest < ActiveSupport::TestCase
...
+    assert_equal false, @user.is_dashboard?
+    @user.grant_dashboard
+    @user.reload
+    assert_equal true, @user.is_dashboard?
+    @user.revoke_dashboard
+    @user.reload
+    assert_equal false, @user.is_dashboard?
+  end
+
+  test "grant and revoke dashboard_set" do
```
edeyoung: @pkmitre : two questions:
1. the grant/revoke code here is only called in rake tasks. the rest of the code accesses the attributes directly. is there a reason for this?
2. what is `dashboard_set` used for? I don't think I've come across this anywhere else.

pkmitre: 1. The code in app/controllers/admin/users_controller.rb  that sets attributes uses an approach to limit what attributes can be set that works better by direct attribute access; otherwise, no specific reason that I know of.
2. The `dashboard_set` user accounts are used to represent the sets of measures available within the complexity dashboard; see app/controllers/complexity_dashboard/measure_sets_controller.rb and lib/tasks/dashboard.rake

edeyoung (approved):

PR #564, opened 2016-08-18
Made "Medication, Order" cumulative medication duration calculate correctly.

edeyoung:
This addresses https://oncprojectracking.healthit.gov/support/browse/BONNIE-172.

Usually, cumulative medication duration is calculated based off of the fulfillment history, dose, and regimen. However, for "Medication, Order", the medication has not yet been filled, so fulfillment history and dose don't make sense. Maybe more importantly, these fields are not available in the QRDA for the "Medication, Order" template.

Instead, cumulative medication duration should be calculated based on the maximum allowed administrations (`repeatNumber` in the QRDA) and the regimen (`effectiveTime` with `period` subelement in the QRDA).

Additionally, fulfillment information should not be allowed to be entered for "Medication, Order".

Specific changes for this repository include GUI changes so the appropriate fields are shown for "Medication, Order" data criteria and so that proper fields are populated in the patient model.

This pull request depends on the `medication_order_cmd` branches in **patient-api, hqmf2js, and health-data-standards.**

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
@@ -112,11 +112,9 @@
         <div class="row">
           {{#ifCond status "==" "ordered"}}
           <div class="form-group col-md-6">
-            <label for="administrations_value_{{@cid}}" class="control-label"><i class="fa fa-medkit" aria-hidden="true"></i> Allowed Administrations<span class="sr-only"> value</span></label>
+            <label for="administrations_value_{{@cid}}" class="control-label"><i class="fa fa-medkit" aria-hidden="true"></i> Number of Doses<span class="sr-only"> value</span></label>
             <div class="form-inline">
-              <div class="input-group">
-                <input type="text" name="administrations_value" id="adminstrations_value_{{@cid}}" class="form-control" placeholder="# of administrations">
-              </div>
+              <input type="number" min="0" name="administrations_value" id="adminstrations_value_{{@cid}}" class="form-control" placeholder="number of doses">
             </div>
           </div>
           {{else}}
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -110,6 +110,16 @@
 
       {{#ifCond definition "==" "medication"}}
         <div class="row">
+          {{#ifCond status "==" "ordered"}}
+          <div class="form-group col-md-6">
+            <label for="administrations_value_{{@cid}}" class="control-label"><i class="fa fa-medkit" aria-hidden="true"></i> Allowed Administrations<span class="sr-only"> value</span></label>
```
thecristen: Is the phrase "Allowed Administrations" universally understood?
If it's always in units of doses, it would more sense to refer to it in the UI as 'number of doses'.

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -110,6 +110,16 @@
...
+            <div class="form-inline">
+              <div class="input-group">
```
thecristen: You don't need the `.input-group` div if there's no `.input-group-addon` inside it.

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -110,6 +110,16 @@
...
+                <input type="text" name="administrations_value" id="adminstrations_value_{{@cid}}" class="form-control" placeholder="# of administrations">
```
thecristen: Consider changing this input type to "number"

thecristen (changes requested):

thecristen (approved):

PR #568, opened 2016-08-29
Patient Dashboard - Comments

holmesie:


app/assets/javascripts/helpers.js.coffee
```diff
@@ -53,3 +53,32 @@ Handlebars.registerHelper 'times', (n, opts) ->
...
+  else
+    options.inverse(this)
+
+###
+Takes a shorthand population name and renders it such
+that screen readers will read out the complete name
+###
+Handlebars.registerHelper 'population', (population) ->
+  return '' unless population?
+  population_map =
```
hossenlopp, line 71: This map will get rebuilt every single time this helper is called. Could the map be moved to the constants file app/assets/javascripts/constants.js.coffe.erb?

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    @BIRTHDATE = "birthdate"
+    @DEATHDATE = "deathdate"
+    @FIRST = "first"
+    @LAST = "last"
+    @GENDER = "gender"
+    @EXPECTED = "expected"
+    @ACTUAL = "actual"
+    @METADATA = "metadata"
+
+    # Column widths
```
edeyoung, line 17: seems like column width definitions belong in the view. (I think I put these here originally... but I think they should live in the view)

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    @COL_WIDTH_FREETEXT = 240
+    @COL_WIDTH_CRITERIA = 200
+    @COL_WIDTH_RESULT = 80
+
+    # Construct a mapping of populations to their data criteria keys
+    @criteriaKeysByPopulation = {}
+    for population in @populations
+      preconditions = @populationSet.get(population)?['preconditions']
+      if preconditions
+        @criteriaKeysByPopulation[population] = @preconditionCriteriaKeys(preconditions[0]).filter (ck) -> ck != 'MeasurePeriod'
```
edeyoung, line 32: what is the `preconditions` array and does it make sense for us to always pick the first item from this array?

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    @COL_WIDTH_META_HUGE = 200
+    @COL_WIDTH_META_LARGE = 110
+    @COL_WIDTH_META_MEDIUM = 100
+    @COL_WIDTH_META_SMALL = 40
+    @COL_WIDTH_FREETEXT = 240
+    @COL_WIDTH_CRITERIA = 200
+    @COL_WIDTH_RESULT = 80
+
+    # Construct a mapping of populations to their data criteria keys
+    @criteriaKeysByPopulation = {}
```
edeyoung, line 28: would be nice to have a longer, more detailed description. Maybe an example? Because I have a hard time understanding what this would look like.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    for population in @populations
+      preconditions = @populationSet.get(population)?['preconditions']
+      if preconditions
+        @criteriaKeysByPopulation[population] = @preconditionCriteriaKeys(preconditions[0]).filter (ck) -> ck != 'MeasurePeriod'
+      else
+        @criteriaKeysByPopulation[population] = []
+    @dataIndices = @getDataIndices(@populations, @criteriaKeysByPopulation)
```
edeyoung, line 35: because this is where the instance variables are defined, would be good to have comments describing what these are (even though it's on the method descriptions)

app/assets/javascripts/models/patient_dashboard.js.coffee
edeyoung, line 28: thoughts on making `@criteriaKeysByPopulation` a local variable? Is it used outside of this class? It's already being passed into functions.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    @dataCollections = @getDataCollections(@populations, @dataIndices, @criteriaKeysByPopulation)
+    @dataInfo = @getDataInfo(@populations, @dataIndices, @dataCollections)
```
edeyoung, line 37: We're passing in instance variables to all of these functions when we can reference without passing them in. Consider not passing these in.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+
+  ###
+  @returns {Array} an array of column names in the right order.
+  ###
+  getDataIndices: (populations, @criteriaKeysByPopulation) =>
```
edeyoung, line 42: `criteriaKeysByPopulation` gets passed in as an instance variable rather than a regular variable. If we do pass it in (see comment above), it should be passed in like a regular variable.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    dataIndices.push(@FIRST)
+    for population in populations
+      dataIndices.push(@ACTUAL + population)
+    for population in populations
+      dataIndices.push(@EXPECTED + population)
+    dataIndices.push(@DESCRIPTION)
+    dataIndices.push(@BIRTHDATE)
+    dataIndices.push(@DEATHDATE)
+    dataIndices.push(@GENDER)
+    for population in populations
```
edeyoung, line 56: with this iterator and the others above, might be nice to have a quick comment describing what's being generated.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+      if preconditions
+        @criteriaKeysByPopulation[population] = @preconditionCriteriaKeys(preconditions[0]).filter (ck) -> ck != 'MeasurePeriod'
+      else
+        @criteriaKeysByPopulation[population] = []
+    @dataIndices = @getDataIndices(@populations, @criteriaKeysByPopulation)
+    @dataCollections = @getDataCollections(@populations, @dataIndices, @criteriaKeysByPopulation)
+    @dataInfo = @getDataInfo(@populations, @dataIndices, @dataCollections)
+
+  ###
+  @returns {Array} an array of column names in the right order.
```
edeyoung, line 40: i'd say these are 'keys' more than names. i'd also mention the purpose of this array.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    dataIndices.push(@DEATHDATE)
+    dataIndices.push(@GENDER)
+    for population in populations
+      criteria = @criteriaKeysByPopulation[population]
+      for criterium in criteria
+        dataIndices.push(population + '_' + criterium)
+    dataIndices
+
+  ###
+  @returns {Array} an array with the top most header columns for the patient
```
edeyoung, line 63: this returns a hash

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+  dashboard table.
+  ###
+  getDataInfo: (populations, dataIndices, dataCollections) =>
+    dataInfo = {}
+    # Get the text for all of the data criteria, mapped by data criteria key
+    dataCriteriaText = {}
+    for reference in Object.keys(@measure.get('data_criteria'))
+      dataLogicView = new Thorax.Views.DataCriteriaLogic(reference: reference, measure: @measure)
+      dataLogicView.appendTo(@$el)
```
edeyoung, line 72: what is the purpose of this line?

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+        dataInfo[item] = name: dataCriteriaText[@getRealKey(item)], width: @COL_WIDTH_CRITERIA
+
+    # Add the index from dataIndices for each key
+    for key, data of dataInfo
+      data.index = dataIndices.indexOf(key)
+
+    return dataInfo
+
+  ###
+  @returns {Array} an array of the sub header columns for the patient dashboard
```
edeyoung, line 101: this returns a hash

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    dataIndices.push(@GENDER)
+    for population in populations
+      criteria = @criteriaKeysByPopulation[population]
+      for criterium in criteria
+        dataIndices.push(population + '_' + criterium)
+    dataIndices
+
+  ###
+  @returns {Array} an array with the top most header columns for the patient
+  dashboard table.
```
edeyoung, line 64: i think this comment could be more descriptive. Maybe something like: "A hash containing metadata information about each column. The metadata information includes the name, the width, and the column index."

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    for key, dataCollection of dataCollections
+      # This grabs the lowest index in the given items
+      dataCollection.firstIndex = _.min(dataCollection.items.map (item) -> dataIndices.indexOf(item))
+      dataCollection.lastIndex = dataCollection.firstIndex + dataCollection.items.length - 1
+    return dataCollections
+
+  ###
+  @returns {string} returns the data criteria id
+  ###
+  getRealKey: (dataKey) =>
```
edeyoung, line 123: I think i remember writing this and I think it's a bit of a hack. Is there a way to rewrite things so this isn't needed? If not, is there a way to make the intent of this function more clear? Maybe a new name?

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    dataInfo[@DEATHDATE] = name: "Deathdate", width: @COL_WIDTH_META_HUGE
+    dataInfo[@GENDER] = name: "Gender", width: @COL_WIDTH_META_MEDIUM
+
+    for population in populations
+      # Include the expected and actual values for each population (IPP/DENOM/etc.)
+      dataInfo[@EXPECTED + population] = name: population, width: @COL_WIDTH_POPULATION
+      dataInfo[@ACTUAL + population] = name: population, width: @COL_WIDTH_POPULATION
+      # Add the data criteria by data criteria key and the data criteria text
+      dataCollection = dataCollections[population]
+      for item in dataCollection.items
```
edeyoung, line 91: can this be done from `@criteriaKeysByPopulation` rather than from `dataCollections`? I wonder if that will make it so `getRealKey` doesn't need to be used?
Also, that will make it so this object can be passed into `getDataCollections` so `dataIndices.indexOf` won't have to be called twice (it's also called in `getDataCollections`).

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    dataCollections[@RESULT] = name: "Result", items: [@RESULT, @LAST, @FIRST]
+    dataCollections[@EXPECTED] = name: "Expected", items: populations.map (p) => @EXPECTED + p
+    dataCollections[@ACTUAL] = name: "Actual", items: populations.map (p) => @ACTUAL + p
+    dataCollections[@DESCRIPTION] = name: "Description", items: [@DESCRIPTION]
+    dataCollections[@METADATA] = name: "Metadata", items: [@BIRTHDATE, @DEATHDATE, @GENDER]
+    for population in populations
+      dataCollections[population] = name: population, items: criteria_keys_by_population[population].map (c) => population + '_' + c
+    for key, dataCollection of dataCollections
+      # This grabs the lowest index in the given items
+      dataCollection.firstIndex = _.min(dataCollection.items.map (item) -> dataIndices.indexOf(item))
```
edeyoung, line 116: See comment for line 91. We might be able to get rid of the `indexOf` call here by passing in the `dataInfo` object.

app/assets/javascripts/models/patient_dashboard.js.coffee
edeyoung, line 123: Maybe rename to "stripPopulationFromKey"?

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    else
+      keyValue = dataKey
+    return keyValue
+
+  ###
+  @returns {Array} given a data criteria, return the list of all data criteria
+  keys referenced within, either through children criteria or temporal
+  references; this includes the passed in criteria reference.
+  ###
+  dataCriteriaChildrenKeys: (criteria_reference) =>
```
edeyoung, line 140: watch variable naming. my guess is we're all over the place in the bonnie code generally, but each file should at least be self-consistent. i looked up the style guide. Looks like classes should be CamelCase and methods/variables should be snake_case.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    criteria_keys = [criteria_reference]
+    if criteria = @measure.get('data_criteria')[criteria_reference]
+      if criteria['children_criteria']?
+        criteria_keys = criteria_keys.concat(@dataCriteriaChildrenKeys(childCriteria) for childCriteria in criteria['children_criteria'])
```
edeyoung, line 144: take a look at using symbols for hash keys instead of strings. saves memory, time, etc. (although I'm doubting that will effect us a whole lot here, but it appears to be a ruby best practice.

So, `criteria['children_criteria']` -> `criteria[:children_criteria]`

edeyoung, line 144: it looks like `concat` directly changes the object, so you don't need to reassign it.
https://ruby-doc.org/core-2.2.0/Array.html#method-i-concat

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+        criteria_keys = _.flatten(criteria_keys)
```
edeyoung, line 145: is `_.flatten` actually necessary? as written, it doesn't seem like it should be. If it is, you can pull it out of the `if` statement and have it occur just above the `return` statement.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+  ###
+  getChildrenCriteria: (criteria_key) =>
+    if criteria_key?
+      children_criteria = @dataCriteriaChildrenKeys(criteria_key)
+      children_criteria = children_criteria.filter (ck) -> ck != 'MeasurePeriod'
+      dataCriteriaViewMap = {}
+      for reference in children_criteria
+        if reference != criteria_key # We want to ignore the data criteria itself when populating map.
+          dataLogicView = new Thorax.Views.DataCriteriaLogic(reference: reference, measure: @measure)
+          dataLogicView.appendTo(@$el)
```
edeyoung, line 176: what's the purpose of this line?

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    return criteria_keys
+
+  ###
+  @returns {Array} given a precondition, return the list of all data criteria
+  keys referenced within.
+  ###
+  preconditionCriteriaKeys: (precondition) =>
+    if precondition['preconditions'] && precondition['preconditions'].length > 0
+      results = (@preconditionCriteriaKeys(precondition) for precondition in precondition['preconditions'])
+      results = _.flatten(results)
```
edeyoung, line 159: same thoughts as above re hash key and `_.flatten`.

app/assets/javascripts/models/patient_dashboard.js.coffee
```diff
@@ -0,0 +1,178 @@
...
+    else if precondition['reference']
+      [precondition['reference']]
+    else
+      []
+
+  ###
+  @returns {Object} children criteria keys and criteria text
+  ###
+  getChildrenCriteria: (criteria_key) =>
```
edeyoung, line 168: it looks like the only place this is used is to determine whether or not a column has children criteria (in app/assets/javascripts/views/patient_dashboard/measure_patient_dashboard.js.coffee line 742). This is a lot of work for a boolean. Take a look at a different way of getting this information.

app/assets/javascripts/models/patient_dashboard_patient.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+    @saveExpectedResults()
+    @saveActualResults()
+    @savePopulationResults()
+
+  ###
+  Sets the expected results for each population as instance variables
+  of this object. These will later be accessed by DataTables when populating
+  the patient dashboard.
+  ###
+  saveExpectedResults: ->
```
edeyoung, line 31: consider renaming from `save` to `set`. `save` implies that this is an action to persist changes, but it looks like it's only used in the initializer.
Same with all the other `save` methods.

app/assets/javascripts/models/patient_dashboard_patient.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+    @last = @patient.get('last')
+    @description = if @patient.get('notes') then @patient.get('notes') else ''
+    @birthdate = @patient.get('birthdate')
+    @birthdate = moment.utc(@birthdate, 'X').format('MM/DD/YYYY hh:mm A')
+    @deathdate = @patient.get('deathdate')
+    @deathdate = if @deathdate then moment.utc(@deathdate, 'X').format('MM/DD/YYYY hh:mm A') else ''
+    @gender = @patient.get('gender')
+
+    # Get expected population results; check if patient is passing
+    @expected = @getExpectedResults()
```
edeyoung, line 16: are `@expected` and `@actual` accessed outside of this class? If not, consider making these local and passing into `saveExpectedResults` and `saveActualResults` (or maybe just call `@getExpectedResults` from those methods).
if so, should they be? we're storing all the actual and expected results in the instance variables created from `saveExpectedResults` and `saveActualResults`. This would be providing duplicate information.

app/assets/javascripts/models/patient_dashboard_patient.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+    for k, v of @pd.criteriaKeysByPopulation
+      for dc in v
+        @[k + '_' + dc] =  @getPatientCriteriaResult(dc, k)
+
+  ###
+  @returns {Object} a mapping of populations to their expected results
+  ###
+  getExpectedResults: ->
+    expectedResults = {}
+    expected_model = (model for model in @patient.get('expected_values').models when model.get('measure_id') == @measure.get('hqmf_set_id') && model.get('population_index') == @population.get('index'))[0]
```
edeyoung, line 59: this is a lot of logic in a single line (hard to debug). consider breaking into multiple lines.

app/assets/javascripts/models/patient_dashboard_patient.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+    for population in @populations
+      if population not in expected_model.keys()
+        expectedResults[population] = 0
+      else
+        expectedResults[population] = expected_model.get(population)
```
edeyoung, line 64: same comment here as in previous file: watch naming conventions and keep them internally consistent.

app/assets/javascripts/models/patient_dashboard_patient.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+
+  ###
+  @returns {String} describes the patient's result for a single data criteria
+  ###
+  getPatientCriteriaResult: (criteriaKey, populationKey) ->
+    if criteriaKey of @patientResult['rationale']
+      value = @patientResult['rationale'][criteriaKey]
+      if value != null && value != 'false' && value != false
+        result = 'TRUE'
+      else if value == 'false' || value == false
```
edeyoung, line 90: Could this just be an `else` statement? The only that's not covered by the `else if` is `value == null`. If `value==null` should it be `false`? If not, what should it be? it's currently ambiguous with how it's written. `result` never gets set if `value == null` but then `value` gets set to `result` later.

app/assets/javascripts/models/patient_dashboard_patient.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+        result = 'FALSE'
+      value = result
+      if 'specificsRationale' of @patientResult && populationKey of @patientResult['specificsRationale']
```
edeyoung, line 93: i think this would read easier with `@patientResult.key?('specificsRationale')` (or as mentioned earlier `:specificsRationale`) rather than `'specificsRationale' of @patientResult`. Same with `populationKey`.

app/assets/javascripts/models/patient_dashboard_patient.js.coffee
```diff
@@ -0,0 +1,110 @@
...
+
+  ###
+  Sets the results for each individual data criteria as instance variables
+  of this object. These will later be accessed by DataTables when populating
+  the patient dashboard.
+  ###
+  savePopulationResults: ->
+    for k, v of @pd.criteriaKeysByPopulation
+      for dc in v
+        @[k + '_' + dc] =  @getPatientCriteriaResult(dc, k)
```
edeyoung, line 52: consider making these variable names (and the ones above) more descriptive. e.g. `k` -> `population`, `v` -> `result`, etc.

app/assets/javascripts/models/population.js.coffee
```diff
@@ -14,6 +14,8 @@ class Thorax.Models.Population extends Thorax.Model
 
   calculationResults: -> new Thorax.Collections.Results @measure().get('patients').map (p) => @calculate(p)
 
+  calculateResult: (patient) -> new Thorax.Collections.Results @calculate(patient)
```
edeyoung, line 17: I found `calculateResults` hard to find (in terms of knowing this is how I managed doing delays due to calculations) and hard to wrap my head around initially. Consider adding some comments to both `calculateReuslt` and `calculateResults` to describe how they're used.

app/assets/javascripts/router.js.coffee
```diff
@@ -38,6 +40,15 @@
...
     @breadcrumb.addMeasurePeriod()
 
+  renderPatientDashboard: (hqmfSetId)->
+    @navigationSetup "Patient Dashboard", "patient-dashboard"
+    measure = @measures.findWhere({hqmf_set_id: hqmfSetId})
+    document.title += " - #{measure.get('cms_id')}" if measure?
+    measureLayoutView = new Thorax.Views.MeasureLayout(measure: measure, patients: @patients)
+    @mainView.setView(measureLayoutView)
+    measureLayoutView.showDashboard showFixedColumns: !/508_patient_dashboard/.test(Backbone.history.getFragment()) #Checks to see if 508_patient_dashboard is not in the url
+    @breadcrumb.addMeasure(measure)
```
edeyoung, line 50: should there be a patient dashboard breadcrumb?

app/assets/javascripts/router.js.coffee
```diff
@@ -65,7 +77,7 @@
     measure = @measures.findWhere({hqmf_set_id: measureHqmfSetId}) if measureHqmfSetId
     patient = if patientId? then @patients.get(patientId) else new Thorax.Models.Patient {measure_ids: [measure?.get('hqmf_set_id')]}, parse: true
     document.title += " - #{measure.get('cms_id')}" if measure?
-    patientBuilderView = new Thorax.Views.PatientBuilder(model: patient, measure: measure, patients: @patients, measures: @measures)
+    patientBuilderView = new Thorax.Views.PatientBuilder(model: patient, measure: measure, patients: @patients, measures: @measures, showCompleteView: true)
```
edeyoung, line 80: consider changing the name of `showCompleteView`. Can this be combined with the `routeToPatientDashboard` boolean? `showCompleteView` made sense when the modal view didn't let you edit meta data. however, it does now. all this does is change the location of the cancel/save. Maybe just name both of these booleans `inPatientDashboard` (or something along those lines)?

app/assets/javascripts/models/patient_dashboard.js.coffee
thecristen: Doesn't the `@` (coffeescript for the javascript `this.`) make it local? Or am I misunderstanding...

edeyoung: I meant have it be local to the method rather than be an instance variable

PR #577, opened 2016-09-23
add some new icons for some new categories of data criteria

thecristen:
These icons are what the user sees in the patient builder when selecting data criteria to add to a patient.

Here's an illustrated before/after of the icons in use:

image


edeyoung:
closing because travis test is failing and i think it was a one-off fail. will reopen so travis tests will run again.

edeyoung (approved):
Looks good!

PR #580, opened 2016-09-27
Learn CQL 👍

holmesie:
The Learn CQL feature is a tool to enable measure developers to test CQL logic segments against synthetic test patient records for an QDM-logic based measure. Currently this only supports the QDM 4.2 data model. This is a beta capability that is currently not intended for long term inclusion in Bonnie.

This commit is a combination of work by aholmes@mitre.org, jhbradley@mitre.org, and krautscheid@mitre.org.


edeyoung (approved):
This is a temporary beta feature. Components of this that get added permanently should go through a more formal code review.

PR #581, opened 2016-09-27
Correctly escape user input in multiple contexts to prevent XSS issues

pkmitre:


hossenlopp:
I just gave this a test. The escaping for the bootstrapping works properly. Escaping for viewing in the patient dashboard works. But when editing in patient dashboard the fields give the escaped data in the edit fields. So if you have a `<` in the description it gets turned into a `&lt;`. It also doesn't escape what shows up immediately in the table after pressing the green check to save it.

PR #582, opened 2016-09-28
Correctly escape user input in multiple contexts to prevent XSS issues

holmesie:


app/assets/javascripts/views/patient_dashboard/measure_patient_dashboard.js.coffee
```diff
@@ -219,6 +226,21 @@ class Thorax.Views.MeasurePopulationPatientDashboard extends Thorax.Views.Bonnie
...
+  Escapes HTML using built in datatables helper
+  ###
+  renderFreeText: (text) =>
+      $.fn.dataTable.render.text().display(text)
+
+  ###
+  Renders a free text cell
+  ###
+  renderCell: (data, type, row, meta) =>
+    if row?['editable']
```
edeyoung, line 241: add comment for why these cases are different - why do you not need to do the fancy rendering stuff for an editable cell.

edeyoung:
I haven't run it yet, but it would be nice to include comments describing why these things are there (XSS). That way, people may be more likely to do this when adding new components because they'll see it for pre-existing components. I can take a stab at adding the comments in because I know people are busy (or are no longer on Bonnie :) )

app/assets/javascripts/views/patient_dashboard/measure_patient_dashboard.js.coffee
holmesie: Okay, I've added an explanation for this and pushed it.

lib/templates/measure_logic.html.erb
```diff
@@ -8,9 +8,9 @@
       <%# logic is displayed using JavaScript, we need to load up all the app JavaScript in this %>
       <%# script tag, instantiate the measure, and run the appropriate display code %>
       <%= app_javascript %>
-      var measure = new Thorax.Models.Measure(<%= measure.to_json.html_safe %>, { parse: true });
+      var measure = new Thorax.Models.Measure(<%= ERB::Util.json_escape(measure.to_json).html_safe %>, { parse: true });
```
hossenlopp, line 11: Why is the ERB::Util namespace used here and not elsewhere?

hossenlopp:
It looks like the script execution issues in bootstrapping and patient dashboard are taken care of.

hossenlopp (approved):
Addendum to my last comment. I was not able to reproduce any of the issues we saw last night and this morning. I would consider this update good to go. The patient dashboard can now handle users using sensitive characters. (i.e. < and >) Without garbling them on edit or executing any of the fields as HTML or script.

My only code comment is about curiosity and consistency. I don't why the name space is needed in that one file.

lib/templates/measure_logic.html.erb
pkmitre: Rails includes ERB::Util automatically when rendering views, but when using bare ERB outside of the normal Rails rendering pipeline (as is the case here, since this is used in a Rake task) the reference needs to be explicit.

PR #584, opened 2016-10-04
Moved secret token to enviornment variable for production deployments

holmesie:
For production deployments, set the secret token in the .env file, as such:

SECRET_TOKEN=<30+ character random string>

For testing and development deployments, this pull request will default to a secret token of 30 \* 'x'.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,9 @@ gem 'less-rails'
 # We want non-digest versions of our assets for font-awesome
 gem "non-stupid-digest-assets"
 
+# Loads environment variables from the .env file
+gem 'dotenv-rails'
+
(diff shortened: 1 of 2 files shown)


pkmitre:
We already handle this issue in a different way. See config/deploy/local.rb, line 17:

```
set :linked_files, (fetch(:linked_files) || []).concat(%w{config/server.yml config/newrelic.yml config/initializers/secret_token.rb})
```

On deployment, this causes the secret_token.rb file to be a symlink to a file local to the production and staging servers, shared/config/initializers/secret_token.rb, which contains a unique secret (generated specific to each server).

If you decide to stick with the current approach, it should likely be documented in the checked in secret_token.rb file to avoid future confusion. It is already documented in the deployment/ubuntu-deployment-steps.txt file.

holmesie:
Aha, I didn't notice that. Thanks Pete!

PR #585, opened 2016-10-10
Backend fixture grouping

mayerm94:
Record sets can now be used by putting records into a sub directory of the records fixture folder and passing that path to the collections_fixtures test helper method as normal


test/test_helper.rb
```diff
@@ -20,6 +20,8 @@ def collection_fixtures(*collection_names)
         fixture_json = JSON.parse(File.read(json_fixture_file))
         convert_times(fixture_json)
         set_mongoid_ids(fixture_json)
+        # The collection (Record,User,Measures) is the root of the collection sets
```
edeyoung: can you add some additional comments here? why does this need to be split?

test/functional/admin_users_controller_test.rb
```diff
@@ -5,7 +5,8 @@ class Admin::UsersControllerTest  < ActionController::TestCase
 
   setup do
     dump_database
-    collection_fixtures("users", "records", "draft_measures")
+    records_set = File.join("records","set_1")
```
edeyoung: I think it makes sense to rename this to "base_tests" instead of "set_1" or something more descriptive. Then as we add tests, they can be in directories like, "measure_history_tests" (or something like that).
I also think it makes sense to do this for the users and the records, as you had suggested in an earlier standup.
Is it possible to combine sets together? Can I use records that would be in both "set_1" and "set_2" or would I need to copy the records I want from "set_1" into "set_2"?

edeyoung (changes requested):
Looks good - expand to users and draft_measures and rename "set_1".

test/functional/admin_users_controller_test.rb
mayerm94: Yes, you can include multiple sets. To include multiple sets you would need to do create a file path for each set (set_1 and set_2 for example) and pass both of the file paths to the collection_fixtures function. In this instance, all records in set_1 and set_2 would be added to the records collection

edeyoung (approved):

mulcahyk:
@mayerm94 can you give an example of how you would add fixtures other than what is in the base set?  It would be helpful to have that in the comments in test_helper.rb.

PR #592, opened 2016-11-21
Bonnie 157 fixture export task

alexanderelliott121:
Adds a fixture export task.

### DIFF
--- a/lib/tasks/generate_fixtures.rake
+++ b/lib/tasks/generate_fixtures.rake
@@ -1,38 +1,45 @@
 namespace :HDS do
   namespace :test do
     
-    task :generateFixture, [:user_email, :cms_id, :patient_first_name, :patient_last_name, :output_dir] => [:environment] do |t, args|
+    task :generateFixture, [:gen_test, :user_email, :cms_id, :patient_first_name, :patient_last_name, :test_set, :test_name] => [:environment] do |t, args|
+      test_name = args[:test_name]
+      if (test_name.nil?)
+        tmp_email = args[:user_email].split("@")[0]
+        test_name = "#{tmp_email}_#{args[:cms_id]}_#{args[:patient_first_name]}_#{args[:patient_last_name]}"
+        test_name["\."] = "_"
+      end
+      
       user = User.find_by email: args[:user_email]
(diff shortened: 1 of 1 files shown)


alexanderelliott121:
Chris Tohnline has brought up the subject of error reporting.
At the moment, failure utilizes the native errors that will be thrown by Mongoid when a search returns empty results.

lib/tasks/generate_fixtures.rake
```diff
@@ -0,0 +1,66 @@
...
+        test_name["\."] = "_"
+      end
+      
+      user = User.find_by email: args[:user_email]
+      if (args[:use_hqmf] == true)
+        measure = user.measures.find_by(hqmf_id: args[:cms_id])
+      else
+        measure = user.measures.find_by(cms_id: args[:cms_id])
+      end
+      patient = Record.find_by(user: user, first: args[:patient_first_name], last: args[:patient_last_name])
```
mulcahyk, line 28: You are searching by the whole `user` object here.  I think that you want to be searching by just the user's id so like this: `user: user.id`.

lib/tasks/generate_fixtures.rake
```diff
@@ -0,0 +1,66 @@
...
+      
+      fixtures_directory = File.join("test", "fixtures")
+      output_directory = File.join(fixtures_directory, args[:test_set], test_name)
+      Dir.mkdir(output_directory) unless Dir.exists? output_directory
```
mulcahyk: When I run I get this: `Errno::ENOENT: No such file or directory @ dir_s_mkdir - test/fixtures/measure_history/fake`.  It doesn't seem to matter what I pass in for the `test_set` and `test_name` parameters, I still get some kind of error when creating the directory.

lib/tasks/generate_fixtures.rake
alexanderelliott121: I believe that the user field invokes the scope: by_user in bonnie's extension of record.
I could look into this more deeply, but the mongoid documentation I've seen so far ha been unhelpful on this front.

lib/tasks/generate_fixtures.rake
alexanderelliott121: I've added support for recursive file creation.

PR #593, opened 2016-11-22
Removed hard coded email address

holmesie:
Removed hard coded email address; Made email config (email.yml) server specific.

### DIFF
--- a/config/bonnie.yml
+++ b/config/bonnie.yml
@@ -10,6 +10,8 @@ defaults: &defaults
     api_url: https://vsac.nlm.nih.gov/vsac/svs/RetrieveMultipleValueSets
     profiles_url: https://vsac.nlm.nih.gov/vsac/profiles
     profile: MU2 Update 2016-04-01
+  # Place holder email address for error alerts. This email is not valid.
+  # This email should be configured in a '<root>/config/email.yml' file.
   bonnie_email: bonnie@example.com
 
 development:
--- a/config/bonnie.yml
+++ b/config/bonnie.yml
@@ -10,6 +10,7 @@ defaults: &defaults
     api_url: https://vsac.nlm.nih.gov/vsac/svs/RetrieveMultipleValueSets
     profiles_url: https://vsac.nlm.nih.gov/vsac/profiles
     profile: MU2 Update 2016-04-01
+  bonnie_email: bonnie@example.com
 
 development:
   <<: *defaults
--- a/config/environment.rb
+++ b/config/environment.rb
@@ -4,6 +4,9 @@ require File.expand_path('../application', __FILE__)
 # Load config here, not in bonnie initializer, so config is available during initialization
 APP_CONFIG = YAML.load_file(Rails.root.join('config', 'bonnie.yml'))[Rails.env]
 
+# Also load in the email settings, if present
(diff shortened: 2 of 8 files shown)


config/environment.rb
```diff
@@ -4,6 +4,9 @@
 # Load config here, not in bonnie initializer, so config is available during initialization
 APP_CONFIG = YAML.load_file(Rails.root.join('config', 'bonnie.yml'))[Rails.env]
 
+# Also load in the email settings, if present
+APP_CONFIG.merge! YAML.load_file(Rails.root.join('config', 'email.yml')) if File.exists? Rails.root.join('config', 'email.yml')
```
c-monkey, line 8: server.yml is already optional, is there a reason you've split out email from this file?

holmesie: `server.yml` handles SMTP/hostname settings, i.e. things reliant on the actual machine the deployment is running on. I felt the email address we are sending errors/signups/etc is more server agnostic, so it should have its own config.

test/functional/admin_users_controller_test.rb
```diff
@@ -52,7 +52,6 @@ class Admin::UsersControllerTest  < ActionController::TestCase
     # Verify transmission of new user email
     mail = ActionMailer::Base.deliveries.last
     assert_equal @user_unapproved.email, mail.to.first
-    assert_equal "bonnie-feedback-list@lists.mitre.org", mail.from.first
```
mulcahyk, line 55: Do we have some way of testing to make sure that the correct email will show up when we need it?

config/bonnie.yml
```diff
@@ -10,6 +10,7 @@ defaults: &defaults
     api_url: https://vsac.nlm.nih.gov/vsac/svs/RetrieveMultipleValueSets
     profiles_url: https://vsac.nlm.nih.gov/vsac/profiles
     profile: MU2 Update 2016-04-01
+  bonnie_email: bonnie@example.com
```
mulcahyk: Do we have it documented somewhere what this value is for and what it should be changed to?
What happens if this is not a valid email address or is blank?

holmesie: This is the default value in case Bonnie was not deployed with an email configuration (`email.yml`). The email value for certain parts of Bonnie can not be set to null, so this needs to exist (since `email.yml` is optional, and local to a deployment server). If this field is not valid, it would be like sending an email to an invalid email address - not an issue.

The intention is to add information for this and `email.yml` for deployments in our deployment instructions/documentation, but I feel that should wait until this is merged.

test/functional/admin_users_controller_test.rb
holmesie: We could probably do that - I'll add that in.

config/bonnie.yml
edeyoung: I agree with @mulcahyk that this should have a comment associated with it, especially as `bonnie@example.com` is not a valid email address.

Maybe something like:
"Place holder email for error alerts. This email is not valid. This email should be configured in X file."

edeyoung (changes requested):
Add comment about email in config file and resolve Kristian's comment about adding a test to ensure the correct email will actually show up.

edeyoung (approved):

holmesie:
Reviving this change for CQL in a new pr.

PR #595, opened 2016-11-23
Measure history population fixes

mulcahyk:
Work dealing with the handling of changes to the number of populations from pre-upload to post-upload.


lib/ext/measure_upload_summary.rb
```diff
@@ -83,13 +84,23 @@ def before_measure_load_compare(patient, population_set_index, hqmf_set_id)
     end
   end
 
-  def self.collect_before_upload_state(measure, arch_measure)
-    patients = Record.where(user_id: measure.user_id, measure_ids: measure.hqmf_set_id)
+  def self.collect_before_upload_state(measure, arch_measure, measure_patients)
+    # patients = Record.where(user_id: measure.user_id, measure_ids: measure.hqmf_set_id)
```
edeyoung: remember to go back through and delete commented out code.

lib/ext/measure_upload_summary.rb
```diff
@@ -83,13 +84,23 @@ def before_measure_load_compare(patient, population_set_index, hqmf_set_id)
     end
   end
 
-  def self.collect_before_upload_state(measure, arch_measure)
-    patients = Record.where(user_id: measure.user_id, measure_ids: measure.hqmf_set_id)
+  def self.collect_before_upload_state(measure, arch_measure, measure_patients)
```
edeyoung: watch naming again. It took me two looks to get that `arch_measure` was `archived_measure` and I was honestly confused for a second about why we care about arches.

lib/ext/measure_upload_summary.rb
```diff
@@ -105,41 +116,62 @@ def self.collect_before_upload_state(measure, arch_measure)
...
-  def self.collect_after_upload_state(measure, upload_summary_id)
-    patient_snapshots_pre_measure_upload = MeasureSummary.where(id: upload_summary_id).first
-    patient_snapshots_pre_measure_upload.population_set_summaries.each_index do |population_set_index|
-      patient_snapshot_population_sets = patient_snapshots_pre_measure_upload.population_set_summaries[population_set_index]
-      patient_snapshot_population_sets[:patients].keys.each do |patient|
-        patient = Record.where(id: patient).first
-        post_upload_filtered_calc_results = (patient.calc_results.find { |result| result[:measure_id] == measure.hqmf_set_id && result[:population_index] == population_set_index }).slice(*ATTRIBUTE_FILTER) unless patient.results_exceed_storage
+  def self.collect_after_upload_state(measure, upload_summary_id, measure_patients)
+    measure_upload_summary = MeasureSummary.where(id: upload_summary_id).first
+    # if measure_upload_summary.population_set_summaries.count < measure.populations.count
```
edeyoung: delete old lines of code.

lib/ext/measure_upload_summary.rb
```diff
@@ -105,41 +116,62 @@ def self.collect_before_upload_state(measure, arch_measure)
...
+    if measure_upload_summary.measure_population_set_count[:pre_upload] < measure_upload_summary.measure_population_set_count[:post_upload]
+      
+      population_sets_to_add = measure.populations.count - measure_upload_summary.population_set_summaries.count
+      # Add the missing population sets
```
edeyoung: we're adding missing population sets. do we need to remove population sets that are no longer there?

lib/ext/measure_upload_summary.rb
```diff
@@ -105,41 +116,62 @@ def self.collect_before_upload_state(measure, arch_measure)
...
+      # Using downto so that the population sets can be added from where they are missing.
+      # It needs to be done this way as the order of the population sets is their population_index.
```
edeyoung: we now have three places in the code where we are getting populations and expected values to align.
Is there somewhere we can put this in a common place? Can this be a function on the patient model or the measure model?

lib/ext/measure_upload_summary.rb
```diff
@@ -105,41 +116,62 @@ def self.collect_before_upload_state(measure, arch_measure)
...
+      population_sets_to_add.downto(1) do |i|
+        new_population_set = PopulationSetSummary.new
+        missing_expected_values = {}
+        # Get the populations 
+        (measure.populations[-i].keys.to_a - ['id', 'title']).each { |population| missing_expected_values[population] = 0 }
```
edeyoung: `to_a` is unnecessary here. `keys` returns an array.

what about doing the slice with `HQMF::PopulationCriteria::ALL_POPULATION_CODES` instead here? this will ensure we're only dealing with populations.

lib/ext/measure_upload_summary.rb
```diff
@@ -105,41 +116,62 @@ def self.collect_before_upload_state(measure, arch_measure)
...
+        # Add all of the patients with their expected values
+        measure_patients.each { |patient| new_population_set['patients'][patient.id.to_s] = {:expected => missing_expected_values} }
```
edeyoung: does the `patient.id` need to be converted to a string for use as a key?

lib/ext/measure_upload_summary.rb
edeyoung: is this still necessary with the work to align populations and expected values done on measure load?

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+
+    assert_response :redirect
+    measure = Measure.where({hqmf_id: '40280381-4555-E1C1-0145-E20602FE49E8'}).first
+    assert_equal 'C0D72444-7C26-4863-9B51-8080F8928A85', measure.hqmf_set_id
+
+
+    collection_fixtures('records')
+    @patients = Record.where(measure_ids: 'C0D72444-7C26-4863-9B51-8080F8928A85')
+    associate_user_with_patients(@user, @patients)
+    # measure = Measure.where({cms_id: "CMS123v3"}).first
```
edeyoung: should this line be deleted?

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+    p = Record.where(first: 'John, III', measure_ids: 'C0D72444-7C26-4863-9B51-8080F8928A85').first
```
edeyoung: add comment about what you're doing here. Is the point to make this patient exceed the record size limit?

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+      attr_reader :tempfile
+    end
+
+    post :create, {measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional', hqmf_set_id: 'C0D72444-7C26-4863-9B51-8080F8928A85'}
+
+    measure = Measure.where(cms_id: 'CMS123v5').first
+    assert_equal 'C0D72444-7C26-4863-9B51-8080F8928A85', measure.hqmf_set_id
+    assert_equal '40280381-51F0-825B-0152-229B7B2F16F9', measure.hqmf_id
+
+    upl = UploadSummary::MeasureSummary.where(hqmf_id: '40280381-51F0-825B-0152-229B7B2F16F9').first
```
edeyoung: names :)

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+    assert_not_nil upl
+    assert_equal 'CMS123v5', upl.measure_cms_id_after
```
edeyoung: do we need an assert for `measure_cms_id_before`?

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+  end
+
+  test "walk through measure population_set change scenarios" do
+    # Load version 1 of the measure. Starts with two population sets
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'CMS704_v1.1.zip'), 'application/zip')
+    class << measure_file
+      attr_reader :tempfile
+    end
+
+    collection_fixtures('records')
```
edeyoung: add comment explaining why you are doing this. Also, does this make sense to put in a `setup` operation so it's done for all tests?

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+      attr_reader :tempfile
+    end
+
+    post :create, {measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional', hqmf_set_id: 'E37012AD-2B01-4738-8985-D5C0F4C3AE9F'}
+
+    assert_response :redirect
+    measure = Measure.where({hqmf_id: '40280382-5859-6598-0158-92BE72260A7E', user_id: @user.id}).first
+    assert_equal 'E37012AD-2B01-4738-8985-D5C0F4C3AE9F', measure.hqmf_set_id
+
+    upl = UploadSummary::MeasureSummary.where(hqmf_id: '40280382-5859-6598-0158-92BE72260A7E').first
```
edeyoung: name :)

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+    assert_nil measure
+
+    post :create, {measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional'}
+
+    assert_response :redirect
+    measure = Measure.where({cms_id: 'CMS704v1', user_id: @user.id}).first
+    assert_equal 'E37012AD-2B01-4738-8985-D5C0F4C3AE9F', measure.hqmf_set_id
+    assert_equal '40280382-5859-6598-0158-89421C4E071B', measure.hqmf_id
+    
+    # Go from two population sets to three
```
edeyoung: i don't think it's asserted anywhere above that there are 2 population sets. This should be added.

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+
+    post :create, {measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional', hqmf_set_id: 'E37012AD-2B01-4738-8985-D5C0F4C3AE9F'}
+
+    assert_response :redirect
+    measure = Measure.where({hqmf_id: '40280382-5859-6598-0158-92BE72260A7E', user_id: @user.id}).first
+    assert_equal 'E37012AD-2B01-4738-8985-D5C0F4C3AE9F', measure.hqmf_set_id
+
+    upl = UploadSummary::MeasureSummary.where(hqmf_id: '40280382-5859-6598-0158-92BE72260A7E').first
+    assert_not_nil upl # There is an upload summary
+    assert_equal 3, upl.population_set_summaries.count # Represents that there are 3 population sets
```
edeyoung: should this also check the before/after population set counts on the object?

test/functional/measures_controller_test.rb
```diff
@@ -439,4 +439,179 @@ class << measure_file
...
+    assert_not_nil upl.population_set_summaries[1]['patients'][test_patient.id.to_s]['pre_upload_results']
+    assert_not_nil upl.population_set_summaries[1]['patients'][test_patient.id.to_s]['post_upload_results']
+
+    # Check that for the patient the third population set has expected, and the expected match the populations for the 3 population set, post_upload_results
+    assert_not_nil upl.population_set_summaries[2]['patients'][test_patient.id.to_s]['expected']
+    # Make sure the number of and names of the populations match up
+    assert_equal 0, (upl.population_set_summaries[2]['patients'][test_patient.id.to_s]['expected'].keys - measure.populations[2].keys).count
+    # There should be nothing for the this population set.
+    assert_nil upl.population_set_summaries[2]['patients'][test_patient.id.to_s]['pre_upload_results']
+    assert_not_nil upl.population_set_summaries[2]['patients'][test_patient.id.to_s]['post_upload_results']
```
edeyoung: this assertion doesn't align with the comment. Clarify the comment.

edeyoung (changes requested):

lib/ext/measure_upload_summary.rb
mulcahyk: I wasn't sure if the measure had the OBSERV_UNITS in it or not.

lib/ext/measure_upload_summary.rb
mulcahyk: As it is `patient.id` is a BSON::ObjectId.  You have to make it a string to use it as a key.

test/functional/measures_controller_test.rb
mulcahyk: This was left as a reminder of sorts that when we merge into `master` we need to switch to using the new call for loading fixtures.  
None of the other tests in this file use a separate `setup` for each test.

test/functional/measures_controller_test.rb
mulcahyk: @edeyoung which *object* are you referring to here?

edeyoung: `upl`

it would be related to this type of thing:

```coffee
    measure_upload_summary.measure_population_set_count[:pre_upload] = before_upload_population_sets.count
    measure_upload_summary.measure_population_set_count[:post_upload] = measure.populations.count
```

test/functional/measures_controller_test.rb
mulcahyk: Yes, that is exactly it.  I have updated the comment.

lib/ext/measure_upload_summary.rb
mulcahyk: per our conversation we are going to leave the `patient.id` as a string for the time being.

edeyoung (approved):

PR #599, opened 2016-12-01
Measure history ldc2

edeyoung:


app/assets/javascripts/views/patient_builder/patient_version_compare.js.coffee
```diff
@@ -8,69 +8,64 @@ class Thorax.Views.PatientBuilderCompare extends Thorax.Views.BonnieView
...
+    populationLogicView
+  
+  # creates a patient logic view based off of the current patient state.
+  _getLiveView: (measure) =>
+    populationLogicView = new Thorax.Views.BuilderPopulationLogic(isCompareView: true, suppressDataCriteriaHighlight: true)
+    populationLogicView.setPopulation measure.get('displayedPopulation')
+    populationLogicView.showRationale @model
+    populationLogicView
   
 # Modified Thorax.Views.BuilderPopulationLogic that accepts new Thorax.Models.cachedResult
```
hossenlopp: Typo `cachedResult` => `CachedResult`

app/assets/javascripts/views/patient_builder/patient_version_compare.js.coffee
```diff
@@ -8,69 +8,64 @@ class Thorax.Views.PatientBuilderCompare extends Thorax.Views.BonnieView
...
     else
-      @populationPresentAfterUpload = false
+      @populationLogicViewAfter = @_getLiveView(@measure)
 
-    @render()  
+    @render()
+  
+  # creates a patient logic view based off of a patient snapshot rather than the current
+  # patient state.
+  _getSnapshotView: (population, summaryCachedResult) =>
```
hossenlopp, line 48: Is prepending a class method with an underscore a way to mark methods as 'private' or should only be used from inside the class?

hossenlopp (changes requested):
This looks good to me. I just had one question about coffeescript class conventions. One small typo in comment. I tested this. It looks like it doesn't break the patient builder highlighting and fixes the latest compare coming from measure history using the current state of the patient.

I did notice the highlighting in Measure Observations didn't work on CMS111 but I presume thats because the crazy stuff from it calculating median times there. I haven't tried it on master/production.

app/assets/javascripts/views/patient_builder/patient_version_compare.js.coffee
edeyoung: yes - i looked into how to do things as private in coffeescript. signalling seemed like the easiest way - or doing nothing at all. curious what you think we should do here or if we should just do nothing.

hossenlopp: I'm cool with `_` meaning only for use in this class. It's a convention I have used elsewhere.

hossenlopp (approved):

PR #600, opened 2016-12-01
Tone down the output of the calc rake task

mulcahyk:
It was putting out an message every time a patient was double processed.  This was overwhelming all of the other messages being generated.
Also added a quick check to get through measures that don't have patient faster.

### DIFF
--- a/lib/tasks/setup_history_tracks.rake
+++ b/lib/tasks/setup_history_tracks.rake
@@ -38,8 +38,8 @@ namespace :bonnie do
           processed_patients_array = []
           patients.each_with_index do |patient, patient_index|
             if processed_patients_array.include?(patient)
-              # puts "\nPatient repeated for #{measure.user.email} measure #{measure.cms_id} population set #{population_index}."
-              # puts "\tID: #{patient.id}\tName: #{patient.first} #{patient.last}"
+              # For some reason patients are added to the list again as the iteration goes along.
+              # This checks for that and skips those patients.
               next
             end
             processed_patients_array << patient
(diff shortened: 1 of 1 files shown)


lib/tasks/setup_history_tracks.rake
```diff
@@ -33,8 +38,8 @@ namespace :bonnie do
           processed_patients_array = []
           patients.each_with_index do |patient, patient_index|
             if processed_patients_array.include?(patient)
-              puts "\nPatient repeated for #{measure.user.email} measure #{measure.cms_id} population set #{population_index}."
-              puts "\tID: #{patient.id}\tName: #{patient.first} #{patient.last}"
+              # puts "\nPatient repeated for #{measure.user.email} measure #{measure.cms_id} population set #{population_index}."
+              # puts "\tID: #{patient.id}\tName: #{patient.first} #{patient.last}"
```
edeyoung: if we're not going to output anything, I'd suggest removing the check entirely. so `processed_patients_array` would be removed above and the entire `if` block and the array push below would be removed as well.

mulcahyk: The check is still valid.  It helps to speed up the processing by not reprocessing the patient again.  I agree in taking out the comments.

edeyoung: agreed. I'd add a comment at the beginning explaining that for some reason patients get repeated and you're skipping patients to not duplicate calculations.

edeyoung:
a new pull request is in with these changes for the rebased branch.

PR #604, opened 2016-12-03
Measure history r quiet down setup

edeyoung:


mulcahyk:
I agree with the changes you made @edeyoung .

mulcahyk:
Updates look good.

PR #605, opened 2016-12-06
fixed issue with displayed population index.

edeyoung:


hossenlopp (approved):
Looks good to me! Would have been embarrassing to let this one through.

PR #610, opened 2016-12-07
Measure history rebased tests

mayerm94:


edeyoung:
Can you send out an email to the team summarizing the folder structure that you're using for these tests?

edeyoung:
Also, why is this archived measure file in a 'base_set' directory?

test/fixtures/archived_measures/base_set/CMS104v2_archived.json

spec/javascripts/fixtures/json/measure_history_reupload_capture/upload_summaries.json
```diff
@@ -0,0 +1,61 @@
+[{
```
edeyoung: It looks like because of how the json files need to be built, there is a directory for each set of tests so it can have its own json file.

I think it makes sense to further break down the directories. Instead of having `json/measure_history_reupload_capture/` and `json/measure_history_version_display/`, what about `json/measure_history/reupload_capture/` and `json/measure_history/version_display/`?

spec/javascripts/fixtures/json/measure_history_version_display/patients.json
```diff
@@ -0,0 +1,342 @@
+[
```
edeyoung: in the backend fixtures, the fixtures are organized by type and then by test set:
`fixtures/draft_measures/base_set/...`. Should these be similarly organized? e.g. `fixtures/json/patients/measure_history/version_display/patients.json`?

edeyoung: Should the "base set" fixtures be added to a `base_set` directory like they are on the back end?

spec/javascripts/models/archived_measure_spec.js.coffee
```diff
@@ -0,0 +1,24 @@
...
+    expect(archived.get('hqmf_id')).toEqual '40280381-3D27-5493-013D-4DCA4B826AE4'
+    expect(archived.get('hqmf_set_id')).toEqual '42BF391F-38A3-4C0F-9ECE-DCD47E9609D9'
+    expect(archived.get('id')).toEqual '584089f91841113b96000868'
+    
+  it 'has measure available', ->
+    archived = @measure.get('archived_measures').models[0]
+    archived_measure = archived.get('measure_content')
+    expect(archived_measure.measure_logic).toExist()
+    expect(archived_measure.value_set_oids).toExist()
+    expect(archived_measure.data_criteria).toExist()
```
edeyoung: Both of the tests in this spec seem to be testing artifical actions: artificially adding the artifact to the measure (not as it's done through the code) and then testing for existence things artificially included in the artifact (from the spec). It doesn't seem like any actual logic is being tested here. What are the operations that happen on the archive measure? How is an archive measure generated from an actual measure?

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -0,0 +1,17 @@
...
+    @uploadSummaries.add(new Thorax.Models.UploadSummary uploadSummaryFixture) for uploadSummaryFixture in getJSONFixture('measure_history_version_display/upload_summary.json')
+    @patients = new Thorax.Collections.Patients()
+    @patients.add(new Thorax.Models.Patient patientFixture) for patientFixture in getJSONFixture('measure_history_version_display/patients.json')
+    @measure.attributes.upload_summaries = @uploadSummaries
+    
+  it 'Ensure that version is displayed properly', ->
+    @measure_history_view = new Thorax.Views.MeasureHistoryView model: @measure, patients: @patients, upload_summaries: @uploadSummaries
+    @measure_history_view.render()
+    expect(@measure_history_view.$('tr[data-upload-id="584187bae76e94d175000227"]')).toContainText "v5.4.000"
+    expect(@measure_history_view.$('tr[data-upload-id="5841877ee76e94d175000100"]')).toContainText "v5.3.000"
```
edeyoung: @mulcahyk does this address your thoughts about this needing to be more detailed?

edeyoung: does order of presentation matter here? I know it doesn't matter in terms of the base test ("is the version being displayed correctly) but does it matter overall or does that make this too brittle?

spec/javascripts/fixtures/json/measure_history_reupload_capture/upload_summaries.json
mayerm94: I agree, having separate nested directories makes sense. I am also adding a base_set directory as suggested in the comments below.

spec/javascripts/fixtures/json/measure_history_version_display/patients.json
mayerm94: This is a good idea. I think it would be good to have a consistent structure between the ruby and javascript tests.

spec/javascripts/models/archived_measure_spec.js.coffee
mayerm94: Good point... I think these assertions would be more appropriate in a back end test to test the  from_measure function of the ArchivedMeasure class

spec/javascripts/fixtures/json/archived_measures/measure_history_set/upload_summaries.json
```diff
@@ -0,0 +1,61 @@
+[{
```
edeyoung: should this file be in an `upload_summaries` folder rather than an `archived_measures` folder?

spec/javascripts/fixtures/json/measure_data/measure_history_set/version_display/measure.json
```diff
@@ -0,0 +1,1709 @@
+{
```
edeyoung: should this be named `measures.json` instead of `measure.json` even if there's just one?

spec/javascripts/fixtures/json/upload_summaries/measure_history_set/upload_summary.json
```diff
@@ -0,0 +1,321 @@
+[
```
edeyoung: what's the difference between this file and the `upload_summaries.json` file that I thought should be moved to this directory that's listed above?

test/fixtures/upload_summaries/CMS104v2_Upload_Summary.json
```diff
@@ -0,0 +1,30 @@
+{
```
edeyoung, line 1: should these be in subfiles (e.g. `measure_history_set`)?

edeyoung (commented):
There are some organizational comments to address.

in spec/javascript/fixtures/json/measure_data, should these be organized differently? One file is directly related to measures, so should perhaps go in a `measure` (or `draft_measure` if we're staying tightly aligned with back-end tests) folder. Maybe the rest stay in the `measure_data` folder? Also, I think for some tests that Alex was working on, he had to do a backend fixture for value_sets. How was that organized?

test/fixtures/upload_summaries/CMS104v2_Upload_Summary.json
mayerm94: I decided against using "draft_measure" as the title because that would suggest that the directory only contains measures. I also thought it was a good idea to keep the measures and valuesets fixture file in measure_date/base_set because they are both used in the setup file measures.js.erb

edeyoung (approved):
There are a few folder structure issues, but these are also present in the pull request that Alex is working on and I've asked him to address them there. I think it makes sense to let them stay here and they'll be addressed once Alex's changes are merged in.

PR #615, opened 2016-12-09
Measure History New Controllers Comments and Fixes

hossenlopp:
- Added comments to controllers and new routes
- Removed skip_before_action. This was bad copy pasting.
- Made it so limited attribute database calls include _id.
  - @mayerm94 showed me these being excluded are causing silent errors that appear in tests
- Removed extraneous routes to controller methods that don't exist.

### DIFF
--- a/app/controllers/archived_measures_controller.rb
+++ b/app/controllers/archived_measures_controller.rb
@@ -1,16 +1,24 @@
-# Methods for handling archived measures
+##
+# This controller provides access to the ArchivedMeasures. Backbone accesses these functions to lazy
+# load ArchivedMeasures.
+
 class ArchivedMeasuresController < ApplicationController
-  
-  skip_before_action :verify_authenticity_token, only: [:show, :value_sets]
   
   respond_to :json, :js, :html
   
-  # List the measure_db_id for all of versions of a measure using the
-  # hqmf_set_id as the measure identifier (so as to version agnostic)
+  ##
+  # GET /measures/:measure_id/archived_measures
+  #
+  # Lists the measure_db_id for all ArchivedMeasures for a specific measure. This is an index function that only returns
+  # a list of measure_db_ids.
+  
   def index
     begin
-      @measure = Measure.by_user(current_user).only(:hqmf_set_id).find(params[:measure_id])
-      @archived_measures = ArchivedMeasure.by_user(current_user).only(:measure_db_id).where(hqmf_set_id: @measure.hqmf_set_id)
+      # Fetch only hqmf_set_id of the measure so we can use it to grab the archived measures for that hqmf_set_id.
+      @measure = Measure.by_user(current_user).only(:hqmf_set_id, :_id).find(params[:measure_id])
+      
+      # Fetch only the measure_db_id and _id fields of the archived measures.
+      @archived_measures = ArchivedMeasure.by_user(current_user).only(:measure_db_id, :_id).where(hqmf_set_id: @measure.hqmf_set_id)
       
       respond_with @archived_measure do |format|
         format.json { render json: @archived_measures }
@@ -20,7 +28,11 @@ class ArchivedMeasuresController < ApplicationController
     end
   end
   
-  # Show the archived content of a measure at specific point in time
+  ##
+  # GET /measures/:measure_id/archived_measures/:id
+  #
+  # Gets a specific archived measure by its measure_db_id.
+  
   def show
     skippable_fields = [:map_fns, :record_ids, :measure_attributes]
     @archived_measure = ArchivedMeasure.by_user(current_user).where(measure_db_id: params[:id]).first
--- a/app/controllers/archived_measures_controller.rb
+++ b/app/controllers/archived_measures_controller.rb
@@ -14,7 +14,8 @@ class ArchivedMeasuresController < ApplicationController
   
   def index
     begin
-      # Fetch only hqmf_set_id of the measure so we can use it to grab the archived measures for that hqmf_set_id.
+      # Fetch only hqmf_set_id and _id fields of the measure so we can use it to grab the archived measures for that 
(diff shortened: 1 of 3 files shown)


mulcahyk (approved):

app/controllers/archived_measures_controller.rb
```diff
@@ -1,16 +1,24 @@
...
+  # GET /measures/:measure_id/archived_measures
+  #
+  # Lists the measure_db_id for all ArchivedMeasures for a specific measure. This is an index function that only returns
+  # a list of measure_db_ids.
+  
   def index
     begin
-      @measure = Measure.by_user(current_user).only(:hqmf_set_id).find(params[:measure_id])
-      @archived_measures = ArchivedMeasure.by_user(current_user).only(:measure_db_id).where(hqmf_set_id: @measure.hqmf_set_id)
+      # Fetch only hqmf_set_id of the measure so we can use it to grab the archived measures for that hqmf_set_id.
```
mayerm94: Fetch only hwmf_set_id **and** _id

app/controllers/upload_summaries_controller.rb
```diff
@@ -1,12 +1,23 @@
...
+  ##
+  # GET /measures/:measure_id/upload_summaries
+  #
+  # Lists the _id and created_at fields for every upload summary for a measure ordered by created_at in a descending
+  # order.
+  
   def index
     begin
-      @measure = Measure.by_user(current_user).only(:hqmf_set_id).find(params[:measure_id])
+      # Fetch only hqmf_set_id of the measure so we can use it to grab the upload summaries for that hqmf_set_id.
```
mayerm94: Same as above, fetch hqmf_set_id **and** _id

app/controllers/archived_measures_controller.rb
```diff
@@ -1,16 +1,24 @@
-# Methods for handling archived measures
+##
+# This controller provides access to the ArchivedMeasures. Backbone accesses these functions to lazy
+# load ArchivedMeasures.
+
```
edeyoung, line 4: this is very minor, but I actually find it confusing when there's a space between a function comment and the function itself. It makes me think the comment is not associated with the function. Is this a recommended style? If not, can you remove the extra space for all these function comments?

edeyoung (changes requested):
address Matt's comments on the comments. Consider my comment about removing space between comments and functions.

app/controllers/archived_measures_controller.rb
hossenlopp: I was just following the RDoc style. https://github.com/rdoc/rdoc#writing-documentation

I don't like the space either, I was just trying to be correct.

app/controllers/archived_measures_controller.rb
hossenlopp: FIXED in 5f1d6a57c84b46f0075fd445d4097ead7517bee6.

app/controllers/upload_summaries_controller.rb
hossenlopp: FIXED in 5f1d6a57c84b46f0075fd445d4097ead7517bee6.

app/controllers/archived_measures_controller.rb
edeyoung: Let's keep it then. Maybe at some point we'll make the rest of the files conform so it won't look as different (code debt task!).

c-monkey: The lazy programmer's solution here is probably the best. Have the team consent to a standard pretty printer format that can be automated. Once the configuration for the pretty printing is shared with the team, then we all conform at the press of a button.

hossenlopp: So... waiting on this PR until we decide a format? I thought we did the research months ago and chose RDoc.

edeyoung: @c-monkey can you clarify your comment?
@hossenlopp no, this is not on hold based on the comment format. Please keep the comments as they are.

c-monkey: Sorry! Wasn't trying to cause confusion, I was just musing that if the team agrees upon a format that can be automatically applied, then any member can be confident they are producing properly formatted code by running the pretty printer.

I am not advocating for any *particular* format, as long as it can be automated.

hossenlopp: Automated in what sense? Generating class documentation html pages or some other automation.

c-monkey: Lizzie's comment started off about the pretty printing of your comment itself (specifically a line of whitespace between the comment and the function declaration). That's just a formatting of the source code, not a formatting of the generated documentation! Oops, I can see why my feedback might be ambiguous. I don't want to configure RDoc, I assume the default format for that is preferable. I was only mentioning that user error formatting the source code can be fixed automatically with a code pretty printer.

hossenlopp: I don't know if it can. I don't even have a code pretty printer setup. 😕

edeyoung (approved):

app/controllers/archived_measures_controller.rb
c-monkey: Other more established languages do have easy pretty printing, but I don't see anything for Ruby with a cursory Google session. That sucks!

One could consider https://github.com/bbatsov/rubocop it seems okay. There is an Atom plugin (rubocop-auto-correct) for applying Rubocop changes automatically. It does some whitespace revisions, but doesn't care either way about spaces between RDoc blocks and the code. The Ruby style guide actually says to *not* have a space between them, but Rubocop doesn't seem to have support for enforcing that.

Certainly we need to table this issue for now, but if there were tension in the team over pretty printing, perhaps we should consider checking out Rubocop or something else.

PR #616, opened 2016-12-12
fixed bug where diffy view did not reflect the selected population. I…

edeyoung:
…ssue was that the diffy view was instatiated in two views and one view was being changed and the other was being shown.


hossenlopp (approved):
Looks good!

PR #620, opened 2016-12-13
Measure history upload summary revamp + misc CSS fixes

thecristen:
# Screenshots

## Measure with multiple populations

Before | After
--- | --- 
image | image
 
## Measure with one population

Before | After
--- | --- 
image | image

# Notes
Addresses BONNIE-239 and several PR comments

### DIFF
--- a/app/assets/javascripts/helpers.js.coffee
+++ b/app/assets/javascripts/helpers.js.coffee
@@ -73,7 +73,7 @@ Handlebars.registerHelper 'populationName', (population) =>
   return new Handlebars.SafeString Thorax.Models.Measure.PopulationMap[population]
 
 ###
-Takes a phrase string and returns a slugified version
+Takes a string and returns a slugified version by replacing any spaces with dashes and making all characters lowercase. This is useful for transforming phrases into strings formatted appropriately for HTML markup. e.g. "Population 2" to "population-2"
 ###
 Handlebars.registerHelper 'slugify', (str) ->
   slug = str.replace(/[^\w\s]+/gi, '').replace(/ +/gi, '-')
--- a/app/assets/javascripts/helpers.js.coffee
+++ b/app/assets/javascripts/helpers.js.coffee
@@ -71,3 +71,10 @@ that screen readers will read out the complete name
 Handlebars.registerHelper 'populationName', (population) =>
   return '' unless population?
   return new Handlebars.SafeString Thorax.Models.Measure.PopulationMap[population]
+
+###
+Takes a phrase string and returns a slugified version
+###
+Handlebars.registerHelper 'slugify', (str) ->
+  slug = str.replace(/[^\w\s]+/gi, '').replace(/ +/gi, '-')
+  return slug.toLowerCase()
--- a/app/assets/javascripts/templates/measure_upload_summary_dialog.hbs
+++ b/app/assets/javascripts/templates/measure_upload_summary_dialog.hbs
@@ -3,10 +3,9 @@
 
     <div class="modal-content">
       <div class="modal-header">
-        <h1 id="uploadSummaryTitle">Measure Upload Completed</h1>
+        <h1 class="modal-title">Measure Upload Completed</h1>
       </div>
       <div class="modal-body">
-        <h2>Your measure has uploaded successfully!</h2>
         {{view uploadSummaryView}}
       </div>
       <div class="modal-footer">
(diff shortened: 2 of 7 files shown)


c-monkey:
I really like how the "failing" was changed to "FAIL", etc. The consistency with the rest of Bonnie reduces confusion.

thecristen:
I pushed a few more commits; this should fix the remaining CSS-related comments in the Measure History pull request!

hossenlopp:
Comparing the color differences in the bubbles.

Before | After
--- | --- 
screen shot 2016-12-13 at 12 52 03 pm | screen shot 2016-12-13 at 12 52 21 pm

app/assets/javascripts/helpers.js.coffee
```diff
@@ -71,3 +71,10 @@ that screen readers will read out the complete name
 Handlebars.registerHelper 'populationName', (population) =>
   return '' unless population?
   return new Handlebars.SafeString Thorax.Models.Measure.PopulationMap[population]
+
+###
+Takes a phrase string and returns a slugified version
```
hossenlopp: What does slugified mean?

edeyoung: agreed - please clarify in comment.

app/assets/javascripts/templates/measure_upload_summary_dialog.hbs
```diff
@@ -3,10 +3,9 @@
 
     <div class="modal-content">
       <div class="modal-header">
-        <h1 id="uploadSummaryTitle">Measure Upload Completed</h1>
+        <h1 class="modal-title">Measure Upload Completed</h1>
```
hossenlopp: Was the modal-title class supposed to be removed? Does this need an id?

thecristen: Good catch, it turns out it needs an ID for the modal's `aria-labelledby` attribute, so I'm going to put the ID back in.

thecristen:
Thanks, @hossenlopp  I meant to add a screenshot with the color change. I did like the previous colors, but using the existing colors seems simpler to me. As a bonus, the foreground and background have better contrast as well.

app/assets/javascripts/templates/measure_upload_summary_dialog.hbs
hossenlopp: Whoops. I totally read that diff wrong... but yeah, I caught the wrong thing, that made you catch the right thing. Cool!

thecristen: woohoo!

app/assets/javascripts/helpers.js.coffee
thecristen: done

edeyoung:
I'm showing this weird green (X)

image


I had a patient passing in v5. I edited the patient so it would fail in V5 but pass in V3. I updated to V3. The summary view displayed as expected but the coloring in the measure history is weird.

edeyoung:
(related to previous comment)

If I go back to `measure_history_rebase` branch, this displays as expected:

image

edeyoung:
Here's the HTML for the cell:

```
                    <td headers="version-before-58506872b0ede66c45000921 upload-58506872b0ede66c45000921 patient-585067f3b0ede66c4500049a">
                      <a class="history-circle text-success bg-success" href="#measures/A4B9763C-847E-4E02-BB7E-ACC596E90E2C/patients/585067f3b0ede66c4500049a/compare/at_upload/58506872b0ede66c45000921" data-upload-id="58506872b0ede66c45000921" data-patient-id="585067f3b0ede66c4500049a">
                        <span class="sr-only">Patient fail before measure upload</span>
                        <i class="fa fa-times" aria-hidden="true"></i>
                      </a>
                    </td>
```

It looks like the issue might be caused by a copy/paste error on line 61 in file `measure_history_timeline.hbs`.

```
                    <a class="history-circle {{#ifCond post_upload_status '==' 'pass'}}text-success bg-success{{else}}text-danger bg-danger{{/ifCond}}" href="#measures/{{../../../../../../hqmf_set_id}}/patients/{{../../id}}/compare/at_upload/{{../../../../../attributes._id}}" data-upload-id="{{../../../../../attributes._id}}" data-patient-id="{{../../id}}">
```

`post_upload_status` is being used rather than `pre_upload_status`.

thecristen:
@edeyoung thanks for the help! updated that and rebased.

PR #622, opened 2016-12-13
Fixing jasmine side effects

c-monkey:
The basic change of this pull request is the measure used for testing is a global variable in 'bonnie' `aka 'window.bonnie = new BonnieRouter()'` in application.js.coffee.

The measures are loaded one time by measures.js.erb. This is a complicated task that we probably don't want to repeat.

Perhaps we could do a deep clone? If that is the correct solution it is odd to me we haven't done it already, but it seems cleaner than having the various unit tests be responsible for removing the changes done to the measure.

### DIFF
--- a/app/assets/javascripts/models/patient.js.coffee
+++ b/app/assets/javascripts/models/patient.js.coffee
@@ -114,9 +114,6 @@ class Thorax.Models.Patient extends Thorax.Model
       defaults = {}
       defaults[section] = [] for section in Thorax.Models.Patient.sections
       @set _(data).chain().pick(_(defaults).keys()).defaults(defaults).value(), silent: true
-      # maybe this error only occurs in unit tests?
-      if @get('source_data_criteria').models.length > data['source_data_criteria'].length
-        debugger
       for criterium, i in @get('source_data_criteria').models
         criterium.set 'coded_entry_id', data['source_data_criteria'][i]['coded_entry_id'], silent: true
         # if we already have codes, then we know we're up to date; no change is necessary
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/models/patient.js.coffee
```diff
@@ -114,6 +114,9 @@ class Thorax.Models.Patient extends Thorax.Model
       defaults = {}
       defaults[section] = [] for section in Thorax.Models.Patient.sections
       @set _(data).chain().pick(_(defaults).keys()).defaults(defaults).value(), silent: true
+      # maybe this error only occurs in unit tests?
+      if @get('source_data_criteria').models.length > data['source_data_criteria'].length
+        debugger
```
hossenlopp: Remove conditional debugger statement.

spec/javascripts/integration/population_state_spec.js.coffee
```diff
@@ -6,61 +6,65 @@ describe "Population state between routes", ->
...
+    # ensure that the second is currently selected
+    expect(active).toBe(populationNavs[1])
+    expect(active.text.trim()).toBe(@measureToTest.get('displayedPopulation').get('title'))
+    expect(@measureToTest.get('displayedPopulation').cid).toBe(@measureToTest.get('populations').at(1).cid)
 
-      @patientBuilder.remove()
+    @patientBuilder.remove()
 
-  it "resets when user goes to measures route", ->
+  xit "resets when user goes to measures route", ->
```
edeyoung, line 69: can you add a comment about why this test is being skipped?

app/assets/javascripts/models/patient.js.coffee
c-monkey: Okay, replaced this debug code with a JIRA ticket.

https://jira.mitre.org/browse/BONNIE-317

edeyoung:
Can you add documentation here: https://github.com/projecttacoma/bonnie/wiki/Testing
on the stateful state of the front-end testing and the need to clear out the patients?

Can you include some sample set-up code (the necessary changes that Matt had to make to make the teaspoon stuff work) and tear-down code?

edeyoung:
(I did not mean to close this pull request)

mayerm94 (approved):

c-monkey:
I am going to update test wiki directions after Alex and I revamp how one does this.

PR #623, opened 2016-12-14
Bonnie 264 test based

alexanderelliott121:
Adds back end fixtures for tested


edeyoung (changes requested):
Can you address the following:

* some sub-directories say `measure_history` and some say `measure_history_set`. Can you make these consistent? Part of me likes `measure_history_set` but I don't have a strong preference if you think otherwise.
* the `draft_measures` directory has both `measure_history` (your fixtures) and `measure_history_set` (kristian's fixtures). Can you merge/deconflict these folders?
* previously in `draft_measures`, the paradigm has been to name the json file with the measure name. Can you change the fixtures your added to do this?
* records/measure_history/multi-populations, can you rename patient.json to patient_1.json for consistency?
* can you add a `measure_history` sub-directory to `upload_summary` and update the corresponding tests?

edeyoung:
can you move the measure exports you added into parallel files where you have the same measure information in the other test folders? e.g. `base_example`.

edeyoung:
Looking at CMS68, it doesn't look like it's possible to have a patient pass in one version and fail in the other version without changing the patient. the versions look too static. I think this should be replaced with a measure where a patient might break between version uploads.

edeyoung:
This got pulled in from a different branch: https://github.com/projecttacoma/bonnie/pull/624

PR #626, opened 2016-12-15
made the measure history rake setup tasks use the methods that have b…

edeyoung:
…een added onto the patient for updating the expected values and adding calculation results.

### DIFF
--- a/lib/ext/measure_upload_summary.rb
+++ b/lib/ext/measure_upload_summary.rb
@@ -78,10 +78,12 @@ module UploadSummary
         patient.update_expected_value_structure!(current_measure)
       end
 
-      # recalculate the patient information for the newly uploaded measure
+      # recalculate the patient information for the newly uploaded measure and
+      # note that the patient has measure history
       calculator = BonnieBackendCalculator.new
       current_measure.populations.each_with_index do |population, population_set_index|
         patients.each do |patient|
+          patient.has_measure_history = true # update_calc_results does the patient save to persist this change
           patient.update_calc_results!(current_measure, population_set_index, calculator)
         end
       end
--- a/lib/tasks/setup_history_tracks.rake
+++ b/lib/tasks/setup_history_tracks.rake
@@ -29,18 +29,9 @@ namespace :bonnie do
           next
         end
         measure.populations.each_with_index do |population_set, population_index|
-          begin
-            calculator.set_measure_and_population(measure, population_index, clear_db_cache: true, rationale: true)
-          rescue => e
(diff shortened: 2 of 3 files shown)


lib/tasks/setup_history_tracks.rake
```diff
@@ -49,17 +40,12 @@ namespace :bonnie do
             processed_patients_array << patient
 
             begin
-              result = calculator.calculate(patient).slice(*patient_fields_to_process)
-            rescue => e
-              puts "\nError calculating for user #{measure.user.email} measure #{measure.cms_id} population set #{population_index} patient '#{patient.first} #{patient.last}' (_id: ObjectId('#{patient.id}')):"
-              puts "Measure calculation exception: #{e.message}"
+              patient.update_calc_results!(measure, population_index, calculator)
```
mulcahyk, line 43: When you use this method it sets the `has_measure_history` flag to true.  In the rake there is no measure history yet, it is just laying the groundwork for history.  We can still use this method, we just need to add something at the end of this task to clear the flag.

edeyoung: that's a really good point. The question I'm not asking is, should `update_calc_results!` set `has_measure_history` to true? It almost seems like it shouldn't...

Or does `has_measure_history` means something else for a patient? does it just mean that the patient has calculation results for a version of the measure? Like, the patient existed when the measure existed?

lib/ext/measure_upload_summary.rb
```diff
@@ -78,10 +78,12 @@ def get_population_set_summaries(current_measure, archived_measure)
         patient.update_expected_value_structure!(current_measure)
       end
 
-      # recalculate the patient information for the newly uploaded measure
+      # recalculate the patient information for the newly uploaded measure and
+      # note that the patient has measure history
       calculator = BonnieBackendCalculator.new
       current_measure.populations.each_with_index do |population, population_set_index|
         patients.each do |patient|
+          patient.has_measure_history = true # update_calc_results does the patient save to persist this change
```
mulcahyk, line 86: Minor but I would have this after the calc results have been set.  Just in case there is an error.

lib/ext/measure_upload_summary.rb
```diff
@@ -78,10 +78,12 @@ def get_population_set_summaries(current_measure, archived_measure)
...
           patient.update_calc_results!(current_measure, population_set_index, calculator)
```
mulcahyk, line 87: There is not error handling on this.  There is also no error handling in the method itself.

mulcahyk (approved):
I agree with these changes.  Did discuss the need to have better error handling for when there is a failure during the calculation.  We will come back to that issue later.

PR #630, opened 2016-12-19
Added selenium-webdriver gem and modified Teaspoon config.

jbradl11:
Added Selenium Webdriver support. In order to use selenium, 2 lines in teaspoon config need to be uncommented. (So we don't break our CI).

Please see the new [Selenium-Testing](https://github.com/projecttacoma/bonnie/wiki/Testing) section on the Testing wiki page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,6 @@
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile ~/.gitignore_global
 
-# Ignore teaspoon config file
-/spec/teaspoon_env.rb
-
 # Ignore bundler config
 /.bundle
 
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile ~/.gitignore_global
 
+# Ignore teaspoon config file
+/spec/teaspoon_env.rb
+
 # Ignore bundler config
 /.bundle
 
--- a/spec/teaspoon_env.rb
+++ b/spec/teaspoon_env.rb
@@ -65,7 +65,7 @@ Teaspoon.configure do |config|
     #suite.hook :fixtures, &proc{}
 
     # Determine whether specs loaded into the test harness should be embedded as individual script tags or concatenated
-    # into a single file. Similar to Rails' asset `debug: true` and `config.assets.debug = true` options. By default, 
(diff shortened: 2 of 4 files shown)


spec/teaspoon_env.rb
```diff
@@ -98,15 +98,15 @@
   # Selenium Webdriver: https://github.com/modeset/teaspoon/wiki/Using-Selenium-WebDriver
   # BrowserStack Webdriver: https://github.com/modeset/teaspoon/wiki/Using-BrowserStack-WebDriver
   # Capybara Webkit: https://github.com/modeset/teaspoon/wiki/Using-Capybara-Webkit
-  #config.driver = :phantomjs
```
mayerm94, line 101: Might want to keep this line in the config commented out so that it is easy to switch in and out

jbradl11: Correct me if I am wrong, but doesn't teaspoon run phantomjs without this line being present? If not, what is the default driver?

mayerm94:
We should add spec/teaspoon_env.rb to the .gitignore so that somebody doesn't accidentally commit the config changes which would break CI

mayerm94 (approved):

.gitignore
```diff
@@ -4,6 +4,9 @@
 # or operating system, you probably want to add a global ignore instead:
 #   git config --global core.excludesfile ~/.gitignore_global
 
+# Ignore teaspoon config file
+/spec/teaspoon_env.rb
```
c-monkey: does this work? even if it does, it seems sketchy, because we need this file, and required changes might not show up in your list of changes so you might forget them in your `git add`

also, see: https://cbanta.wordpress.com/2010/11/10/ignore-git-files-that-already-exist-in-the-repository/

mayerm94: My thought was that since most people are going to be modifying this file, we don't want the changes committed by accident which would break the travis build.

c-monkey: Does your change to the .gitignore file properly exclude these changes from showing up in a `git add` command?

mayerm94: You're right though, since the file was already in the repo, the changes will still show up in the "changes not staged for commit" list, meaning a git commit -a would still mistakenly commit the teaspoon config changes

mayerm94: Git will not mention changes made to the file if this is run (git rm --cached spec/teaspoon_env.rb). Though I will revert the change made to the git ignore

PR #632, opened 2016-12-20
there was a bug processing patients with multiple expected values for…

edeyoung:
… multiple measures associated iwth them. The code broke out of the loop instead of continuing to the next item in the array. there was also an issue where some expected_value items were empty dictionaries. This is now handled.


edeyoung:
this bug was found on the production database on bonnie-review. user account hmdmtjc@gmail.com measure CMS72

mulcahyk (approved):
Looks good.  I agree.

PR #633, opened 2016-12-20
fixed bug where update summary was not showing up if the measure need…

edeyoung:
…ed finalization.


mulcahyk:
Do we need line 347?  If we do do we need to add `flash[:uploaded_hqmf_set_id] = measure.hqmf_set_id`

mulcahyk (approved):
Looks good.

PR #634, opened 2016-12-21
Front End Testing: Allow bonnie state loading

alexanderelliott121:
Allows tests to load a specific state for the bonnie router, without issues resulting from changing global variables.


spec/javascripts/router/router_cache.js.coffee
```diff
@@ -0,0 +1,18 @@
+# duplicates the BonnieRouter for test isolation
+@BonnieRouterCache = class BonnieRouterCache
+  DEFAULT_STATE = 'default'
```
edeyoung: add comment

app/assets/javascripts/models/measure.js.coffee
```diff
@@ -194,3 +194,13 @@ class Thorax.Collections.Measures extends Thorax.Collection
     measureToOids = {} # measure hqmf_set_id : valueSet oid
     @each (m) => measureToOids[m.get('hqmf_set_id')] = m.valueSets().pluck('oid')
     measureToOids
+    
+  deepClone: ->
+    cloneMeasures = new Thorax.Collections.Measures (@.toJSON())
+    cloneMeasures.each (measure) ->
+      clonePopulations = new Thorax.Collections.Population
+      clonePopulations.parent = measure
+      measure.get('populations').each (p) -> clonePopulations.add(p.toJSON())
```
edeyoung: can this be done by just pointing the measure populations parent to the new measure? do you need to do double JSON serialization?

spec/javascripts/router/router_cache.js.coffee
```diff
@@ -0,0 +1,18 @@
...
+  constructor: () ->
+    @state = {}
+    #default state is as we find things now
+    this.save(DEFAULT_STATE)
+
+  load: (key) ->
+    throw "Key '#{key}' doesn't exist!" if not (key of @state)
```
edeyoung, line 11: should we add a comment with the assumption that `@state` has been populated already from calling the `save` function? I'm ambivalent on this comment, but reading this line made me wonder about how `@state` was being populated in time for `load`.

spec/javascripts/router/router_cache.js.coffee
```diff
@@ -0,0 +1,18 @@
...
+    # Clear the fixtures cache so that getJSONFixture does not return stale/modified fixtures
+    jasmine.getJSONFixtures().clearCache()
+    # deep copy
+    window.bonnie = $.extend(true, {}, @state[key])
+    window.bonnie.measures = @state[key].measures.deepClone()
+
+  save: (key) ->
+    @state[key] = window.bonnie
```
edeyoung, line 25: is this enough? do we need to clone bonnie when we save the state? If this is just a pointer, does this change when we update window.bonnie?

spec/javascripts/router/router_cache_spec.js.coffee
```diff
@@ -0,0 +1,16 @@
...
+    expect( -> window.bonnieRouterCache.load('non-existant')).toThrow()
+
+  it 'does not throw if state exists', ->
+    expect( -> window.bonnieRouterCache.load('default')).not.toThrow()
+
+  it 'load clears JSON Fixtures', ->
+    myfixture = getJSONFixture('records/base_set/patients.json')
+    expect($.isEmptyObject(jasmine.getJSONFixtures().fixturesCache_)).toBe(false)
+    window.bonnieRouterCache.load('default')
+    expect($.isEmptyObject(jasmine.getJSONFixtures().fixturesCache_)).toBe(true)
```
edeyoung, line 16: maybe add a test ensuring that two different states are actually different? Like, `@state['base_set']` is actually different than `@state['measure_history_set']` (this is based off my concern of how `@state[key]` is currently being set. See comment above).

spec/javascripts/router/router_spec.js.coffee
```diff
@@ -0,0 +1,5 @@
+# naming convention of this spec file matches that of the BonnieRouter (i.e. router.js instead of bonnie_router.js)
+describe 'BonnieRouter', ->
+  it 'no measures in router by default', ->
```
edeyoung: nitpick, but consider rephrasing name to match expected syntax:

`it 'has no measures by default'`

so when the test displays, it will show something like

`bonnie router has no measures by default`

spec/javascripts/setup/measures.js.erb
```diff
@@ -4,34 +4,39 @@
 <%# We want Jasmine to be independent of content in any particular MongoDB instance %>
 <% Mongoid.configure { |config| config.sessions = { default: { database: 'jasmine', hosts: ['localhost:27017'] } } } %>
 
-<%# Load the JSON for our fixture measures %>
-<% measures_fixture_file = File.join File.dirname(__FILE__), '../fixtures/json/measure_data/base_set/measures.json' %>
-<% measures_json = File.read(measures_fixture_file) %>
-
 <%# Set up the value sets %>
+<%# TODO: this value_sets_json is duplicated per state into bonnie.valueSetsByOid, but should be loaded separately?  %>
```
edeyoung, line 8: what's the status on the TODO?

If I'm understanding it correctly, shouldn't this be loaded separately? Are we going to have value set issues if we load the `base_set`'s value sets for the `measure_history_set` measures?

c-monkey: @mayerm94 and I discussed and today we don't plan on having more value sets, but it's kinda lame to just use one giant set. @alexanderelliott121 thinks we should just suck it up today and duplicate the existing so we already support multiple. I'm ambivalent, except if we never make any more value sets, then it is a waste of memory and adds complexity.

edeyoung (approved):

c-monkey:
@alexanderelliott121 and I added documentation to the wiki:
https://github.com/projecttacoma/bonnie/wiki/Testing

PR #636, opened 2016-12-21
Patient Save Refactor

hossenlopp:
Added a new function to the patient model for both the patient builder and patient dashboard to use. This function runs the calculations before saving.

Other changes needed for this:
 - Calculation result deferred now to resolves with a reference to the result.
 - Patient materialize now has an optional callback for when it completes.

### DIFF
--- a/app/assets/javascripts/models/patient.js.coffee
+++ b/app/assets/javascripts/models/patient.js.coffee
@@ -260,15 +260,10 @@ class Thorax.Models.Patient extends Thorax.Model
       
       # wait for all calculation deferreds to complete
       $.when.apply(@, allCalculations)
-        .done((results...) =>
+        .done( (results...) =>
           # Pull out only the result parts we need to save and replace them on the patient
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/models/patient.js.coffee
```diff
@@ -203,6 +206,71 @@ class Thorax.Models.Patient extends Thorax.Model
...
+      # wait for all calculation deferreds to complete
+      $.when.apply(@, allCalculations)
+        .done((results...) =>
+          # Pull out only the result parts we need to save and replace them on the patient
+          @set({ calc_results: _.map(results, @_filterResult) }, { silent: true })
+          @save(null, options)
+          )
+        .fail( ->
+          # TODO: deal with failure situation
+          debugger
```
edeyoung: can you do something here like a log message? there is also examples of a "flash" pop up when there's a calculation failure on the dashboard.

edeyoung (changes requested):
it looks like a debugger statement is still there. error handling needed.

edeyoung (approved):

PR #639, opened 2016-12-22
updated rake task to use 'no_timeout' so that the mongo iterator woul…

edeyoung:
…dn't timeout for long processing times.


edeyoung:
also made there be a check on the email address as there is a measure in the staging server with a null user and this causes an error.

mulcahyk (approved):

PR #648, opened 2016-12-27
Bonnie 307 timeline text

edeyoung:
initially, this was just going to add a new test to check that the timeline view was correct. However, it now has a generalized method for loading history measures and there's cleanup to use this method and remove old fixtures.


spec/javascripts/setup/measures.js.erb
```diff
@@ -9,6 +9,8 @@
 <% value_sets_fixture_file = File.join File.dirname(__FILE__), '../fixtures/json/measure_data/base_set/value_sets.json' %>
 <% value_sets_json = File.read(value_sets_fixture_file) %>
 
+window.measureHistorySpecLoader = new MeasureHistorySpecLoader();
```
c-monkey: Since measureHistorySpecLoader would fail if called prior to the next line (setting the global bonnieRouterCache) perhaps their order should be reversed?

This obviously ins't a problem today, but if through time this file were modified for other changes and the spec loader called before the route cache was created it would be a bug.

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -1,20 +1,18 @@
 describe 'MeasureHistoryView', ->
   
   beforeEach (done) ->
```
c-monkey, line 3: gorgeous changes, I really like the feel of this setup code compared to the previous version

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -1,20 +1,18 @@
...
-    @patients = new Thorax.Collections.Patients()
-    @patients.add(new Thorax.Models.Patient patientFixture) for patientFixture in getJSONFixture('records/measure_history_set/patients.json')
-    @measure.attributes.upload_summaries = @uploadSummaries
+
     @measure_history_view = new Thorax.Views.MeasureHistoryView model: @measure, patients: @patients, upload_summaries: @uploadSummaries
     @measure_history_view.on "rendered", ->
       done()
     @measure_history_view.render()
     
   it 'Ensure that version is displayed properly', ->
```
c-monkey: This test is really a characterization test for the fixture tuple 'single_population_set' 'CMS68v6' and should now be renamed thus. Also, test name is wrong grammar, shouldn't start with a capital. How about 'contains version strings from single_population_set CMS68v6 fixture'?

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -1,20 +1,18 @@
...
-    expect(@uploadSummaries.length).toEqual 2
-    expect(@measure_history_view.$('tr[data-upload-id="584187bae76e94d175000227"]')[0]).toContainText "v5.4.000"
-    expect(@measure_history_view.$('tr[data-upload-id="584187bae76e94d175000227"]')[1]).toContainText "v5.3.000"
+    expect(@uploadSummaries.length).toEqual 3
```
c-monkey: This number '3' is brittle. It's serving as an array length assertion here so we don't index out of bounds. But then, why are we not testing three version strings?

spec/javascripts/views/patient_builder_spec.js.coffee
```diff
@@ -330,13 +330,24 @@ describe 'PatientBuilderView', ->
...
   
-  beforeEach -> 
-    history_patient = new Thorax.Models.Patient getJSONFixture('records/measure_history_set/patient_history_set/CMS104v2PatientWithHistory.json'), parse: true
-    cms104v2 = new Thorax.Models.Measure getJSONFixture('measure_data/measure_history_set/patient_history_set/CMS104v2.json'), parse: true
-    patients = new Thorax.Collections.Patients()
-    patients.add(history_patient)
-    @patientBuilder = new Thorax.Views.PatientBuilder(model: history_patient, measure: cms104v2, patients: patients)
+  beforeEach ->
+    results = window.measureHistorySpecLoader.loadWithHistory('single_population_set', 'CMS68v6')
+
```
c-monkey: This line of whitespace is missing from the `beforeEach` method in MeasureHistoryView. I think we should be consistent, although I can't think of any trivial way to enforce that code style. Today one could pass in the suite and have measure, patients, and uploadSummaries populated, but that reduces our flexibility - or do we always intend to set those 3 fields, in which case we can remove this redundant code?

spec/javascripts/views/measure_history_display_spec.js.coffee
c-monkey: Since we're going to rely on these tests all over the place, shouldn't the same characterization tests exist for the other 2 measure historical contexts? Are we testing enough of the characteristics? Is there documentation outside of the code for what these measure fixtures/historical contexts contain? Should there be?

spec/javascripts/helpers/measure_history_spec_loader.js.coffee
```diff
@@ -0,0 +1,63 @@
...
+    patients = new Thorax.Collections.Patients()
+    patients.add(new Thorax.Models.Patient patientFixture) for patientFixture in getJSONFixture('records/' + path + '/patients.json')
+
+    patients.each (patient) ->
+      patient.set('has_measure_history', true)
+
+    uploadSummaries= new Thorax.Collections.UploadSummaries([], {measure_id: measure.id, _fetched: true})
+    initialLoad = new Thorax.Models.UploadSummary getJSONFixture('upload_summaries/' + path + '/initialLoad.json'), parse: true
+    initialLoad._fetched = true
+    update1 = new Thorax.Models.UploadSummary getJSONFixture('upload_summaries/' + path + '/update1.json'), parse: true
```
c-monkey: Did you consider factoring the shared code between `loadWithHistory` before this line and most of `loadWithoutHistory` into a shared method since they are identical?

I googled how to make a method private, if you're worried about confusing the interface:
https://crimefighter.svbtle.com/using-private-methods-in-coffeescript

c-monkey (changes requested):
I really like this change!

c-monkey (approved):

PR #649, opened 2016-12-27
Measure history r cv fix

edeyoung:
Test with CMS111v5

measure history was not properly handling continuous variable measures (measures with OBSERV values). this code makes it so measure history does properly handle those measure types.

Note: there are pre-existing issues with the saved expected OBSERV values populating correctly in the patient builder. This seems to be unrelated to the measure history changes. I've added an internal bug: BONNIE-353.


c-monkey (approved):
I don't know if there are any automated tests for this change. I did a manual test of before and after and observed does appear to properly serialize (while it didn't before).

edeyoung:
there are tests for this in the queue of tests for this sprint, but none currently in existence.

PR #650, opened 2016-12-29
Upload Summary Dialog Ignore Uploads With No Previous Version

hossenlopp:


mulcahyk (commented):
I think the change itself is fine.  
Does it make sense though to have the message be: "Patient calculations are unchanged"?

hossenlopp:
@mulcahyk That message doesn't show in the case there is no previous copy of the measure.

edeyoung (approved):

PR #651, opened 2016-12-29
Export for measure history

alexanderelliott121:
Creates fixtures, adds fixture export tool.


edeyoung:
I saw some issues with the backend fixtures that were added (naming and structure). I wasn't comfortable with these additions so I created a new pull request removing these particular additions. I'm closing this branch but not deleting it so we can easily add these fixtures back in at another time.

See new pull request here: https://github.com/projecttacoma/bonnie/pull/653

PR #654, opened 2016-12-30
Measure history r upload summaries 500 error

hossenlopp:
- Added dialog boxes that show up if the loading fails
- Added index for descending order `created_at` on the upload summary model. This makes it so mongo can skip sorting when fetching the upload_summaries

### DIFF
--- a/app/assets/javascripts/router.js.coffee
+++ b/app/assets/javascripts/router.js.coffee
@@ -114,8 +114,8 @@
         @navigationSetup "Measure Upload History - #{measure.get('cms_id')}", 'test-case-history'
         # @collection = new Thorax.Collections.Patients
         @mainView.setView new Thorax.Views.MeasureHistoryView model: measure, patients: measure.get('patients'), upload_summaries: uploadSummaries
-        @breadcrumb.viewMeasureHistory(measure)
-        )
+        @breadcrumb.viewMeasureHistory(measure) )
+      .fail( => @showError title: "Measure History Load Failed", summary: "Historic data failed to load due to a server error." )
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/router.js.coffee
```diff
@@ -114,8 +114,8 @@
         @navigationSetup "Measure Upload History - #{measure.get('cms_id')}", 'test-case-history'
         # @collection = new Thorax.Collections.Patients
         @mainView.setView new Thorax.Views.MeasureHistoryView model: measure, patients: measure.get('patients'), upload_summaries: uploadSummaries
-        @breadcrumb.viewMeasureHistory(measure)
-        )
+        @breadcrumb.viewMeasureHistory(measure) )
+      .fail( => @showError title: "Measure History Load Failed", summary: "Historic data failed to load due to a server error." )
```
edeyoung, line 118: Make all the error messages refer the user to the bonnie-feedback list.

edeyoung: NM!  The error pop up already does this.

edeyoung (approved):
I think this works well for now. However, for longer term it would be nice if this navigated you back to the measure dashboard view.

app/assets/javascripts/router.js.coffee
hossenlopp: It doesn't refer them to the bonnie feedback list. Just the JIRA. 
screen shot 2016-12-30 at 2 46 33 pm

edeyoung: yeah, that's fine

hossenlopp: Should I have it navigate back to the measure in this PR?

PR #667, opened 2017-01-17
Bonnie 410 update fixtures tool

alexanderelliott121:
Updates the fixture generation tool to create frontend fixtures, updates backend fixture generation signature to be more streamlined/generalized.


lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,106 @@
...
+    #   values: cms, hqmf
+    # path: Fixture path.
+    # user_email: email of user to export
+    # measure_id: id of measuer to export
+    # patients: which patients should be exported.
+    #   values: each element should have three fields, firstname, lastname, displayname
+    #   ex: [[first1, last1, patient1], [first2, last2, patient2]]
+    ###
+    task :generate_frontend_fixture, [:cms_hqmf, :path, :test_name, :user_email, :measure_id] => [:environment] do |t, args|
+      user = User.find_by email: args[:user_email]
```
mayerm94: Should probably end if user is not found

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,106 @@
...
+      if (args[:cms_hqmf].downcase  == 'cms')  
+        measure = user.measures.find_by(cms_id: args[:measure_id])
+      elsif (args[:cms_hqmf] == 'hqmf')
+        measure = user.measures.find_by(hqmf_id: args[:measure_id])
+      else
+      end
```
mayerm94: Add content to else for error check

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,106 @@
...
+      
+      oid_to_vs_map = {}
+      
+      value_sets = measure.value_sets.each do |vs|
+        oid_to_vs_map[vs.oid] = vs
+      end
+      value_sets_file = File.join(fixtures_path, 'measure_data', args[:path], cms_dir, 'value_sets.json')
+      create_fixture_file(value_sets_file, JSON.pretty_generate(JSON.parse(oid_to_vs_map.to_json)))
+    
+      us_filename = "#{args[:test_name]}.json"
```
mayerm94: Why is the upload_summaries filename named after an argument but the archived_measures, value_sets and measures are named after what they contain, ie: measures.json, archived_measures.json and value_sets.json

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,106 @@
...
-        test_name = "#{tmp_email}_#{args[:cms_id]}_#{args[:patient_first_name]}_#{args[:patient_last_name]}"
-        test_name["\."] = "_"
-      end
-      
+    task :generate_fixtures, [:cms_hqmf, :path, :user_email, :measure_id] => [:environment] do |t, args|
       user = User.find_by email: args[:user_email]
-      if (args[:use_hqmf] == true)
-        measure = user.measures.find_by(hqmf_id: args[:cms_id])
+      if (args[:cms_hqmf].downcase  == 'cms')  
+        measure = user.measures.find_by(cms_id: args[:measure_id])
```
mayerm94: Getting the user and determining if the cms_hqmf arg is a cms_id or hqmf_id seems to be repeated code between rake tasks. Perhaps it could be moved to it's own function

mayerm94 (changes requested):

lib/tasks/generate_fixtures.rake
alexanderelliott121: This is a weird one.
At the moment, the infrastructure we have looks upload_summaries in a specific way: it expects three files called initialLoad, update1 and update2.
I'm not adverse to changing this or updating the load tool, but I ultimately decided not to change the existing structures in this branch.

lib/tasks/generate_fixtures.rake
alexanderelliott121: It does, mongoid error.

mayerm94 (approved):

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
-      end
-      
-      user = User.find_by email: args[:user_email]
-      if (args[:use_hqmf] == true)
-        measure = user.measures.find_by(hqmf_id: args[:cms_id])
+
+    def get_measure(user, cms_hqmf, measure_id)
+      if (cms_hqmf.downcase  == 'cms')  
+        measure = user.measures.find_by(cms_id: measure_id)
+      elsif (cms_hqmf == 'hqmf')
```
edeyoung: should this also be `downcase`?

edeyoung: or maybe these should be exact matches? you have a nice error message if they're not. then we don't downcase anything.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
-      measure_file = args[:gen_test] ? File.join(fixtures_directory, "draft_measures", args[:test_set], test_name, "measures.json") : File.join(output_directory, "measure.json")
-      value_sets_dir = args[:gen_test] ? File.join(fixtures_directory, "health_data_standards_svs_value_sets", args[:test_set], test_name) : File.join(output_directory, "value_sets")
+    def create_fixture_file(file_path, fixture_json)
+      FileUtils.mkdir_p(File.dirname(file_path)) unless Dir.exists? File.dirname(file_path)
+      File.new(file_path, "w+")
+      File.write(file_path, fixture_json)
+    end
 
+    ###
+    # cms_hqmf: Use cms or hqmf id.
```
edeyoung: Change comment to "indicates if a CMS id or an HQMF id is used"

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+    #   values: cms, hqmf
+    # path: Fixture path.
```
edeyoung: Change to "path to output fixture files" or something similarly more explanatory

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+    # test_name: Name of the test the fixture is for.
```
edeyoung: What about `test_type`? That makes this seem more generalizable. I'd like to use the same set of fixtures across multiple tests as long as I want similar sort of behavior (e.g. proportion measure with multiple populations).

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+    # user_email: email of user to export
+    # measure_id: id of measuer to export
```
edeyoung: should we mention that the measure needs to be contained in user_email's account?

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+      File.new(file_path, "w+")
+      File.write(file_path, fixture_json)
+    end
 
+    ###
+    # cms_hqmf: Use cms or hqmf id.
+    #   values: cms, hqmf
+    # path: Fixture path.
+    # test_name: Name of the test the fixture is for.
+    # user_email: email of user to export
```
edeyoung: should we also add something about this being the account that will be used for patients/measure/etc.?

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
-      patient_file = args[:gen_test] ? File.join(fixtures_directory, "records", args[:test_set], test_name, "patient.json") : File.join(output_directory, "patient.json")
-      measure_file = args[:gen_test] ? File.join(fixtures_directory, "draft_measures", args[:test_set], test_name, "measures.json") : File.join(output_directory, "measure.json")
-      value_sets_dir = args[:gen_test] ? File.join(fixtures_directory, "health_data_standards_svs_value_sets", args[:test_set], test_name) : File.join(output_directory, "value_sets")
+    def create_fixture_file(file_path, fixture_json)
+      FileUtils.mkdir_p(File.dirname(file_path)) unless Dir.exists? File.dirname(file_path)
+      File.new(file_path, "w+")
+      File.write(file_path, fixture_json)
+    end
 
+    ###
```
edeyoung: Can you confirm that this comment follows ruby comment guidance? Talk with Hoss for where to find the guidance. Most of our comments don't follow this, but since this is so comprehensive and for an important rake task, I think it would be good to follow the guidance here.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+    # cms_hqmf: Use cms or hqmf id.
+    #   values: cms, hqmf
+    # path: Fixture path.
+    # test_name: Name of the test the fixture is for.
+    # user_email: email of user to export
+    # measure_id: id of measuer to export
+    # bundle exec rake HDS:test:generate_fixtures[cms,test/CMSFakevFake,initialLoad,fake@fake,CMSFakevFake]
```
edeyoung: Can you note that this is an example usage?
Can you use real-ish data? the "FakevFake" is hard to parse.
To ensure there are no typos, it would be good to use an example that you were actually able to execute on your machine.

edeyoung: Also, please add spaces after the commas for readability.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+    ###
+    task :generate_frontend_fixture, [:cms_hqmf, :path, :test_name, :user_email, :measure_id] => [:environment] do |t, args|
+      user = User.find_by email: args[:user_email]
+      measure = get_measure(user, args[:cms_hqmf], args[:measure_id])
+      
+      cms_dir = File.join(args[:test_name], args[:measure_id])
```
edeyoung: why is this called `cms_dir`? Should this be `output_dir` or something similar?

edeyoung: for `test_name` may be worth adding to the comment above that this will be used to build the path to the tests along with the `path` argument.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+      
+      measure_summaries = UploadSummary::MeasureSummary.by_user_and_hqmf_set_id(user, measure.hqmf_set_id).first
+      archived_measures = ArchivedMeasure.by_user_and_hqmf_set_id(user, measure.hqmf_set_id)
+      
+      fixtures_path = File.join('spec', 'javascripts', 'fixtures', 'json')
+      
+      measure_file = File.join(fixtures_path, 'measure_data', args[:path], cms_dir, 'measures.json')
```
edeyoung: add a comment above this along the lines of "generating the fixtures for the measure. these go in a `measure_data` parent directory. the measure file is called `measures.json`. The accompanying value sets file is called `value_sets.json`"

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
-        FileUtils.mkdir_p(File.dirname(patient_file)) unless Dir.exists? File.dirname(patient_file)
-        FileUtils.mkdir_p(File.dirname(measure_file)) unless Dir.exists? File.dirname(measure_file)
-        FileUtils.mkdir_p(value_sets_dir) unless Dir.exists? value_sets_dir
+      records = Record.by_user_and_hqmf_set_id(user, measure.hqmf_set_id)
+      record_file = File.join(fixtures_path, 'records', args[:path], cms_dir, "patients.json")
+      create_fixture_file(record_file, JSON.pretty_generate(JSON.parse(records.to_json)))
+      
+      oid_to_vs_map = {}
+      
+      value_sets = measure.value_sets.each do |vs|
```
edeyoung: move this value set stuff up to the measure stuff to create cohesion in where things are being generated.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+      
+      measure_file = File.join(fixtures_path, 'measure_data', args[:path], cms_dir, 'measures.json')
+      create_fixture_file(measure_file, JSON.pretty_generate(JSON.parse([measure].to_json)))
 
-      if args[:gen_test]
-        FileUtils.mkdir_p(File.dirname(user_file)) unless Dir.exists? File.dirname(user_file)
-        FileUtils.mkdir_p(File.dirname(patient_file)) unless Dir.exists? File.dirname(patient_file)
-        FileUtils.mkdir_p(File.dirname(measure_file)) unless Dir.exists? File.dirname(measure_file)
-        FileUtils.mkdir_p(value_sets_dir) unless Dir.exists? value_sets_dir
+      records = Record.by_user_and_hqmf_set_id(user, measure.hqmf_set_id)
```
edeyoung: isn't the fixture for a record a bit different on the front end than the back end?

edeyoung: add a similar high level comment as described above to this patient section.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+      
+      oid_to_vs_map = {}
+      
+      value_sets = measure.value_sets.each do |vs|
+        oid_to_vs_map[vs.oid] = vs
       end
+      value_sets_file = File.join(fixtures_path, 'measure_data', args[:path], cms_dir, 'value_sets.json')
+      create_fixture_file(value_sets_file, JSON.pretty_generate(JSON.parse(oid_to_vs_map.to_json)))
+    
+      us_filename = "#{args[:test_name]}.json"
```
edeyoung: rename to upload_summaries_filename or something similar. `us` is confusing.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+    # measure_id: id of measuer to export
+    # bundle exec rake HDS:test:generate_fixtures[cms,test/CMSFakevFake,initialLoad,fake@fake,CMSFakevFake]
+    ###
+    task :generate_frontend_fixture, [:cms_hqmf, :path, :test_name, :user_email, :measure_id] => [:environment] do |t, args|
+      user = User.find_by email: args[:user_email]
+      measure = get_measure(user, args[:cms_hqmf], args[:measure_id])
+      
+      cms_dir = File.join(args[:test_name], args[:measure_id])
+      
+      measure_summaries = UploadSummary::MeasureSummary.by_user_and_hqmf_set_id(user, measure.hqmf_set_id).first
```
edeyoung: consider moving these declarations down to where the fixtures get generated for them.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
 
+    ###
+    # cms_hqmf: Use cms or hqmf id.
+    #   values: cms, hqmf
+    # path: Fixture path.
+    # test_name: Name of the test the fixture is for.
+    # user_email: email of user to export
+    # measure_id: id of measuer to export
+    # bundle exec rake HDS:test:generate_fixtures[cms,test/CMSFakevFake,initialLoad,fake@fake,CMSFakevFake]
+    ###
```
edeyoung: consider adding to the comments a comment outlining what the file structure will look like after this is run.

edeyoung: also consider adding an explanation on where each of these files should go in the actual spec file structure.

what happened to making these be in one collective space so you wouldn't have to move individual files?

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
-      value_sets = measure.value_sets
+    #Usage
+    # cms_hqmf: Use cms or hqmf id.
+    #   values: cms, hqmf
+    # path: Fixture path.
+    # user_email: email of user to export
+    # measure_id: id of measuer to export
+    # bundle exec rake HDS:test:generate_fixtures[cms,test/CMSFakevFake,fake@fake,CMSFakevFake]
+    desc "Exports a set of fixtures that can be loaded for testing purposes"
+    task :generate_fixtures, [:cms_hqmf, :path, :user_email, :measure_id] => [:environment] do |t, args|
```
edeyoung: should this be renamed to "generate_backend_fixtures"?

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
-      File.new(measure_file, "w+")
-      File.write(measure_file, JSON.pretty_generate(JSON.parse(measure.to_json(except: [:map_fns, :record_ids], methods: [:value_sets]))))
-
-      value_sets = measure.value_sets
+    #Usage
+    # cms_hqmf: Use cms or hqmf id.
+    #   values: cms, hqmf
+    # path: Fixture path.
+    # user_email: email of user to export
+    # measure_id: id of measuer to export
```
edeyoung: see all comments above re making comments more explanatory.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
+      measure_file = File.join(fixtures_path, 'draft_measures', args[:path], measure_name)
+      create_fixture_file(measure_file, JSON.pretty_generate(JSON.parse(measure.to_json)))
 
-      value_sets.each do |vs|
-        vs_file_name = vs.oid + ".json"
-        value_sets_file = File.join(value_sets_dir, vs_file_name)
-        File.new(value_sets_file, "w+")
-        File.write(value_sets_file, JSON.pretty_generate(JSON.parse(vs.to_json)))
+      records = Record.by_user_and_hqmf_set_id(user, measure.hqmf_set_id).each do |rec|
+        rec_name = rec.first + "_" + rec.last + ".json"
```
edeyoung: the first and last names often have spaces in them. does it make sense (as painful as it is) to use a space here instead of an underscore?

lib/tasks/generate_fixtures.rake
```diff
@@ -1,64 +1,109 @@
...
-
-      value_sets = measure.value_sets
+    #Usage
+    # cms_hqmf: Use cms or hqmf id.
+    #   values: cms, hqmf
+    # path: Fixture path.
+    # user_email: email of user to export
+    # measure_id: id of measuer to export
+    # bundle exec rake HDS:test:generate_fixtures[cms,test/CMSFakevFake,fake@fake,CMSFakevFake]
+    desc "Exports a set of fixtures that can be loaded for testing purposes"
```
edeyoung: it seems like there is some repeated code in here. Is it worth refactoring to reduce repeated code?

edeyoung (changes requested):
Please excuse my brevity. rushed the comments. i'm curious about the resulting file structure. let's discuss. i'm also curious because the fixtures for front and back end seem the same other than being in an array or separate. i thought there were differences. does it turn out there aren't?

lib/tasks/generate_fixtures.rake
```diff
@@ -1,83 +1,127 @@
...
+    def set_mongoid_ids(json)
+      if json.kind_of?( Hash)
+        json.each_pair do |k,v|
+          if v && v.kind_of?( Hash )
+            if v["$oid"]
+              json[k] = BSON::ObjectId.from_string(v["$oid"])
+            else
+              set_mongoid_ids(v)
+            end
+          elsif k == '_id' || k == 'bundle_id' || k == 'user_id'
```
edeyoung: does this cause an issue with `cms_id`?

lib/tasks/generate_fixtures.rake
```diff
@@ -1,83 +1,127 @@
...
+            puts "Converting #{k} : value #{v}"
+            json[k] = BSON::ObjectId.from_string(v)   
+            puts "Converted: #{k} to #{json[k]}"   
+          end
+        end
+      end
+    end
+    
+    def collection_fixtures(*collection_names)
```
edeyoung: are these duplicated from elsewhere in the code? if so, can these just call that code? if not, please add a comment explaining why the duplication was necessary.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,83 +1,127 @@
...
-    # cms_hqmf: Use cms or hqmf id.
+    # cms_hqmf: indicates if a CMS id or an HQMF id is used.
     #   values: cms, hqmf
-    # path: Fixture path.
-    # test_name: Name of the test the fixture is for.
+    # path: Path to fixture files, derived from the data-type directories (EX: measure_data/${path}).
     # user_email: email of user to export
     # measure_id: id of measuer to export
-    # bundle exec rake HDS:test:generate_fixtures[cms,test/CMSFakevFake,initialLoad,fake@fake,CMSFakevFake]
+    # bundle exec rake HDS:test:generate_frontend_fixtures[cms,test/CMSFakevFake,initialLoad,fake@fake,CMSFakevFake]
```
edeyoung: `HDS:test:` got changed to `bonnie:fixtures` above

edeyoung: try using real data for the example. the fake data is hard to read.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,83 +1,127 @@
...
     end
 
     ###
-    # cms_hqmf: Use cms or hqmf id.
+    # cms_hqmf: indicates if a CMS id or an HQMF id is used.
     #   values: cms, hqmf
-    # path: Fixture path.
-    # test_name: Name of the test the fixture is for.
+    # path: Path to fixture files, derived from the data-type directories (EX: measure_data/${path}).
     # user_email: email of user to export
```
edeyoung: from previous review: "should we also add something about this being the account that will be used for patients/measure/etc.?"

lib/tasks/generate_fixtures.rake
```diff
@@ -1,83 +1,127 @@
...
     # measure_id: id of measuer to export
```
edeyoung: from previous review: "should we mention that the measure needs to be contained in user_email's account?
 Show outdated"

lib/tasks/generate_fixtures.rake
edeyoung: from the previous review: 
Can you note that this is an example usage?
Can you use real-ish data? the "FakevFake" is hard to parse.
To ensure there are no typos, it would be good to use an example that you were actually able to execute on your machine.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,83 +1,127 @@
...
-    task :generate_frontend_fixture, [:cms_hqmf, :path, :test_name, :user_email, :measure_id] => [:environment] do |t, args|
+    task :generate_frontend_fixtures, [:cms_hqmf, :path, :user_email, :measure_id] => [:environment] do |t, args|
       user = User.find_by email: args[:user_email]
       measure = get_measure(user, args[:cms_hqmf], args[:measure_id])
       
-      cms_dir = File.join(args[:test_name], args[:measure_id])
       
-      measure_summaries = UploadSummary::MeasureSummary.by_user_and_hqmf_set_id(user, measure.hqmf_set_id).first
+      measure_summaries = UploadSummary::MeasureSummary.by_user_and_hqmf_set_id(user, measure.hqmf_set_id).desc(:created_at)
       archived_measures = ArchivedMeasure.by_user_and_hqmf_set_id(user, measure.hqmf_set_id)
```
edeyoung: from previous review: consider moving these declarations down to where the fixtures get generated for them.
 Hide outdated

lib/tasks/generate_fixtures.rake
```diff
@@ -1,83 +1,127 @@
...
       
       fixtures_path = File.join('spec', 'javascripts', 'fixtures', 'json')
       
-      measure_file = File.join(fixtures_path, 'measure_data', args[:path], cms_dir, 'measures.json')
+      measure_file = File.join(fixtures_path, 'measure_data', args[:path], 'measures.json')
```
edeyoung: from previous review: add a comment above this along the lines of "generating the fixtures for the measure. these go in a measure_data parent directory. the measure file is called measures.json. The accompanying value sets file is called value_sets.json"

lib/tasks/generate_fixtures.rake
```diff
@@ -1,83 +1,127 @@
...
-      create_fixture_file(record_file, JSON.pretty_generate(JSON.parse(records.to_json)))
-      
       oid_to_vs_map = {}
       
       value_sets = measure.value_sets.each do |vs|
         oid_to_vs_map[vs.oid] = vs
       end
-      value_sets_file = File.join(fixtures_path, 'measure_data', args[:path], cms_dir, 'value_sets.json')
+
+      records = Record.by_user_and_hqmf_set_id(user, measure.hqmf_set_id)
```
edeyoung: from previous review: add a similar high level comment as described above to this patient section.
 Show outdated

edeyoung (changes requested):
Hey, it looks like several of my comments were missed. github is doing weird stuff which is making  me wonder if maybe they didn't show up. If they did show up and you disagreed with them so didn't implement them, can you make sure to respond to the comment noting that? Thanks!

lib/tasks/generate_fixtures.rake
alexanderelliott121: Nope

edeyoung: why not?

alexanderelliott121: k == '_id' isn't a regex, so it doesn't recognize/convert the cms_id.
cms_id isn't a bson object, so no conversion is required.

edeyoung:
@alexanderelliott121 

I do not seem to be able to login to the bonnie user after I wipe my db.

please check out my very minor changes.

edeyoung (approved):

PR #668, opened 2017-01-17
Bonnie 273 patient history test

alexanderelliott121:
Adds a test to ensure that newly created patients have no history.

### DIFF
--- a/test/functional/patients_controller_test.rb
+++ b/test/functional/patients_controller_test.rb
@@ -55,8 +55,11 @@ include Devise::TestHelpers
     assert_equal 2, r.source_data_criteria.length
     assert_equal 'EncounterPerformedPsychVisitDiagnosticEvaluation', r.source_data_criteria[0]["id"]
     assert_equal 1, r.encounters.length
-    assert_equal "pass", r.calc_results[0]['status'], "Checking that calc_status worked."
+    assert_equal [{"measure_id"=>"E35791DF-5B25-41BB-B260-673337BC44A8", "population_index"=>"0", "IPP"=>"1", "DENOM"=>"0", "NUMER"=>"0", "status"=>"pass"}], r.calc_results, "Checking that calc_status worked."
     assert_equal false, r.has_measure_history
+    assert_equal false, r.results_exceed_storage
+    assert_equal nil, r.condensed_calc_results
+    assert_equal 128, r.results_size
     assert_equal nil, UploadSummary::PopulationSetSummary.first
     json = JSON.parse(response.body)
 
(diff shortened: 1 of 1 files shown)


test/functional/patients_controller_test.rb
```diff
@@ -56,6 +56,8 @@ class PatientsControllerTest  < ActionController::TestCase
     assert_equal 'EncounterPerformedPsychVisitDiagnosticEvaluation', r.source_data_criteria[0]["id"]
     assert_equal 1, r.encounters.length
     assert_equal "pass", r.calc_results[0]['status'], "Checking that calc_status worked."
```
edeyoung: this line only looks at the `calc_results` status. Can you also check the values for each population?

test/functional/patients_controller_test.rb
```diff
@@ -56,6 +56,8 @@ class PatientsControllerTest  < ActionController::TestCase
...
+    assert_equal false, r.has_measure_history
+    assert_equal nil, UploadSummary::PopulationSetSummary.first
```
edeyoung: I'm unclear on why this assertion belongs here. Can you explain?

edeyoung: These are all the new fields added for measure history:

```
  field :calc_results, type: Array, default: []
  field :has_measure_history, type: Boolean, default: false # has the record gone through an update to the measure
  field :results_exceed_storage, type: Boolean, default: false # True when the size of calc_results > 12000000
  field :condensed_calc_results, type: Array
  field :results_size, type: Integer
```

Can you check that each exists and has the appropriate value?

edeyoung (changes requested):

test/functional/patients_controller_test.rb
alexanderelliott121: Fields have been added

alexanderelliott121: My original thought was that the PopulationSetSummary contained patient upload history.
Its inclusion is an error.

test/functional/patients_controller_test.rb
```diff
@@ -55,7 +55,12 @@ class PatientsControllerTest  < ActionController::TestCase
     assert_equal 2, r.source_data_criteria.length
     assert_equal 'EncounterPerformedPsychVisitDiagnosticEvaluation', r.source_data_criteria[0]["id"]
     assert_equal 1, r.encounters.length
-    assert_equal "pass", r.calc_results[0]['status'], "Checking that calc_status worked."
+    assert_equal [{"measure_id"=>"E35791DF-5B25-41BB-B260-673337BC44A8", "population_index"=>"0", "IPP"=>"1", "DENOM"=>"0", "NUMER"=>"0", "status"=>"pass"}], r.calc_results, "Checking that calc_status worked."
+    assert_equal false, r.has_measure_history
+    assert_equal false, r.results_exceed_storage
+    assert_equal nil, r.condensed_calc_results
+    assert_equal 128, r.results_size
+    assert_equal nil, UploadSummary::PopulationSetSummary.first
```
edeyoung: should this be more like `UploadSummary::PopulationSetSummary.where(...)` and we give it the appropriate info to find the patient?

Or, if you're comfortable that we know that state (that it's basically an empty DB with no measure uploaded/updated that have patients associated with them), can you add a comment noting that?

Other than this, this PR looks good! once this gets updated I'll pull it in!

edeyoung (approved):

PR #669, opened 2017-01-18
add tests for measure history timeline patient display

thecristen:


spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -44,3 +44,40 @@ describe 'MeasureHistoryView', ->
...
+
+            # Check pass/fail text
+            expect(afterCell).toContainText summary.post_upload_status
+            expect(beforeCell).toContainText summary.pre_upload_status
+
+            # Check pass/fail icon
+            expect($(afterCell).find("i.fa")[0].className).toMatch (if summary.post_upload_status is "pass" then "fa-check" else "fa-times")
+            expect($(beforeCell).find("i.fa")[0].className).toMatch (if summary.pre_upload_status is "pass" then "fa-check" else "fa-times")
+
+    it 'does *not* show new patients', ->
```
edeyoung: I don't think this method of adding new patients exercises the functionality being tested. I believe that this view checks if `has_measure_history` is true on the patient. If so, it displays it. Here, `new_clone` actually does have `has_measure_history` as true. I think you may need to create the new patient (I think as you did here is ok for `new_patient`), associate it with the measure, and then create a new instance of measure history to check that it's not showing the new patient. It may also be worth adding `new_clone` as you have here, but expecting that to show up (because `has_measure_history` is true).

@hossenlopp can you add your thoughts on how to appropriately test this?

edeyoung (changes requested):
i like the `does show current patients` test. I think the `does *not* show new patients` may need to be reworked.

spec/javascripts/views/measure_history_display_spec.js.coffee
hossenlopp: `has_measure_history` is not used by the measure history timeline view or template. It looks through all of the upload summaries to collect the list of patients. If a patient (found by id) is not in an upload summary, then it will not be in the `@patientIndex` that is used to populate the template.

I think this test is very close. I have two suggestions:
- The two added patients should have specific `_id`s that are not in any of the upload summaries and unique so they will not be included in `@patientIndex`.
- Create the MeasureHistoryView from scratch to ensure it is redoing all of its initialization.
  - The correct result you have now is happening simply because the `loadHistory` method in MeasureHistoryTimelineView is not being called again. Its code is what needs to be tested.

edeyoung: I think the patients may also need to be added to the measure and have the measure hqmf_set_id referenced in `measure_ids` on the patient

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -66,18 +66,35 @@ describe 'MeasureHistoryView', ->
...
       @new_clone = @measure_history_view.patients.at(0).deepClone({omit_id: true})
+      
+      @patients.add [@new_patient, @new_clone]
+      expect(@patients.length).toEqual 4
 
-      @measure_history_view.patients.add [@new_patient, @new_clone]
-      @measure_history_view.measureTimelineView.render()
+      # Add all the patients as patients on the measure
+      for patient in @patients
+        @measure.get('patients').push [patient]
```
thecristen: You shouldn't need to loop through the patients; you can add them the same way as in line 80 (adding models to a Thorax.Collection):

```coffeescript
@measure.get('patients').add [@new_patient, @new_clone]
```

edeyoung: I second cristen's comment. additionally, looping through this way, you're re-adding patients that are already on the measure (if the measure history fixture loader does things correctly).

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -44,3 +44,57 @@ describe 'MeasureHistoryView', ->
...
+      expect(@measure.get('patients').length).toEqual 4
+
+      # create new measure history view with the measure
+      renderView @, =>  
+       
+        expect(@measure_history_view.patients.length).toEqual 4
+        expect(@measure_history_view.measureTimelineView.patients.length).toEqual 4
+
+        expect(@measure_history_view.$('th[class="history-patient-header"]').length).not.toEqual 4
```
edeyoung: you can just check that it equals 2

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -44,3 +44,57 @@ describe 'MeasureHistoryView', ->
...
+      # Create and add new patients to this measure history view
+      @new_patient = new Thorax.Models.Patient getJSONFixture('records/base_set/patients.json')[0], parse: true
+      @new_clone = @measure_history_view.patients.at(0).deepClone({omit_id: true})
+      
+      @patients.add [@new_patient, @new_clone]
+      expect(@patients.length).toEqual 4
+
+      # Add all the patients as patients on the measure
+      for patient in @patients
+        @measure.get('patients').push [patient]
```
edeyoung: we talked about you editing the spec loader so that it properly adds patients to the measure. that code is located here: spec/javascripts/helpers/measure_history_spec_loader. after line 52, you can add something along the lines of:

```
measure.set('patients', patients)
```

edeyoung (changes requested):

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -42,3 +42,54 @@ describe 'MeasureHistoryView', ->
...
+
+      # Create and add new patients to this measure history view
+      @new_patient = new Thorax.Models.Patient getJSONFixture('records/base_set/patients.json')[0], parse: true
+      @new_clone = @measure_history_view.patients.at(0).deepClone({omit_id: true})
+
+      @patients.add [@new_patient, @new_clone]
+      expect(@patients.length).toEqual 4
+
+      # Add all the patients as patients on the measure
+      @measure.get('patients').add [@new_patient, @new_clone]
```
edeyoung: the patient's measure ids also need to be reset (didn't think of this when we were talking).

For example (untested):

```
@new_patient.set('measure_ids', [@measure.get('hqmf_set_id'), null])
@new_clone.set('measure_ids', [@measure.get('hqmf_set_id'), null])
```

mulcahyk: The patients already have their `measures_ids` set.  In fact it looks like the patients in this test are portfolio patients in that they have multiple `measure_ids`.

edeyoung: i don't think they're necessarily set to the current measure. you can clear it out and reset it to the current measure (which is what the example code i gave you would do).

edeyoung (approved):

mulcahyk (approved):
I'm approving the work that Cristen and Chris H. did.

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -42,3 +42,57 @@ describe 'MeasureHistoryView', ->
...
+
+      # Confirm that there are only 2 patients in the most resent uploadSummaries (assumes that uploadSummaries are sorted in desc order based on date)
+      expect(Object.keys(this.uploadSummaries.at(0).get('population_set_summaries')[0].patients).length).toEqual 2
+
+      # Create and add new patients to this measure history view
+      @new_patient = new Thorax.Models.Patient getJSONFixture('records/base_set/patients.json')[0], parse: true
+      @new_clone = @measure_history_view.patients.at(0).deepClone({omit_id: true})
+      
+      # Make sure that the patients are associated to the current measusre
+      @new_patient.set('measure_ids', [@measure.get('hqmf_set_id'), null])
```
hossenlopp: The `null` at the end of the array shouldn't be needed. I don't why bonnie adds that to all `measure_ids` arrays.

spec/javascripts/views/measure_history_display_spec.js.coffee
hossenlopp: That `null` at the end of the array in @edeyoung's example shouldn't be needed. I don't know why bonnie adds that.

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -42,3 +42,57 @@ describe 'MeasureHistoryView', ->
...
+
+      # create new measure history view with the measure
+      renderView @, =>
+
+        expect(@measure_history_view.patients.length).toEqual 4
+        expect(@measure_history_view.measureTimelineView.patients.length).toEqual 4
+
+        expect(@measure_history_view.$('th[class="history-patient-header"]').length).toEqual 2
+
+        # Confirm that the number of uploadSummaries hasn't changed
```
hossenlopp: This comment doesn't match the check that is being done below it. It is checking the first (and only?) upload summary has only two patients in its record.

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -42,3 +42,57 @@ describe 'MeasureHistoryView', ->
...
+      @measure.get('patients').add [@new_patient, @new_clone]
+      expect(@measure.get('patients').length).toEqual 4
+
+      # create new measure history view with the measure
+      renderView @, =>
+
+        expect(@measure_history_view.patients.length).toEqual 4
+        expect(@measure_history_view.measureTimelineView.patients.length).toEqual 4
+
+        expect(@measure_history_view.$('th[class="history-patient-header"]').length).toEqual 2
```
hossenlopp: Same as above. This could look at the `patientIndex` of the MeasureHistoryTimelineView.

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -1,11 +1,11 @@
 describe 'MeasureHistoryView', ->
 
-  renderView = (suite, done) ->
+  renderView = (suite, callback) ->
     suite.mainView = new Thorax.LayoutView(el: '#bonnie')
     suite.measure_history_view = new Thorax.Views.MeasureHistoryView
```
hossenlopp: `measure_history_view` should be camelcase.

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -42,3 +42,57 @@ describe 'MeasureHistoryView', ->
...
+            expect($(beforeCell).find("i.fa")[0].className).toMatch (if summary.pre_upload_status is "pass" then "fa-check" else "fa-times")
+
+    it 'does *not* show new patients', (done) ->
+      expect(@measure_history_view.$('th[class="history-patient-header"]').length).toEqual 2
+
+      # Confirm that there are only 2 patients in the most resent uploadSummaries (assumes that uploadSummaries are sorted in desc order based on date)
+      expect(Object.keys(this.uploadSummaries.at(0).get('population_set_summaries')[0].patients).length).toEqual 2
+
+      # Create and add new patients to this measure history view
+      @new_patient = new Thorax.Models.Patient getJSONFixture('records/base_set/patients.json')[0], parse: true
```
hossenlopp: `@new_patient` and `@new_clone` should both be camelcase.

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -42,3 +42,57 @@ describe 'MeasureHistoryView', ->
       spyOn(bonnie,'showPageNotFound')
       bonnie.renderMeasureUploadHistory('non_existant_hqmf_set_id')
       expect(bonnie.showPageNotFound).toHaveBeenCalled()
+
+    it 'does show current patients', ->
+      expect(@measure_history_view.patients.length).toEqual 2
+      expect(@measure_history_view.$('th[class="history-patient-header"]').length).toEqual 2
```
hossenlopp: To check the number of patients, it would be easier to look at the data that the template is using than to look at the resultant DOM. The MeasureHistoryTimelineView has a member `patientIndex` that could be checked. ex:

```coffeescript
expect(@measure_history_view.measureTimelineView.patientIndex.length).toEqual 2
```

Unless the point of this suite is to test the template too, then ignore me.

spec/javascripts/views/measure_history_display_spec.js.coffee
```diff
@@ -42,3 +42,57 @@ describe 'MeasureHistoryView', ->
...
+
+      @uploadSummaries.each (us) =>
+        summary_id = us.get('_id')
+        for population_set_summary in us.get('population_set_summaries')
```
hossenlopp: This test would fail if there is more than one population set because only one is rendered at a time. I think it should specifically look at the 0th population set instead of all of them.

hossenlopp (changes requested):
A few small changes, mainly casing of variables. This test is a bit odd since it does a lot of things that test the template instead of the view. Not sure if that is normal/expected or not.

spec/javascripts/views/measure_history_display_spec.js.coffee
mulcahyk: I didn't write the initial version of this test.  I'm going to err on the of caution here and say we leave it as is given that this test is more extensive.  I can also be easily convinced that it should be changed.

hossenlopp (approved):

PR #670, opened 2017-01-19
Bonnie 335

mulcahyk:
Test baseline functionality for simple measure (proportion with a single population set).

Created a test for measure history related measures where the tests themselves are defined as methods and then those methods are "applied" to the measure.


mayerm94:
The CI tests for this pull request should pass once https://github.com/projecttacoma/bonnie/pull/671 is merged into master and this branch is rebased onto master. OR you could rebase onto  BONNIE-440-Remove-Console-Errors

edeyoung:
#671 has been merged into master.
@mulcahyk can you do the rebase so these pass?

test/fixtures/records/measure_history_set/base_example/patient_1.json
```diff
@@ -1,5 +1,5 @@
 {
-  "_id": "58514916e76e941a3d000489",
+  "_id": { "$oid": "58514916e76e941a3d000489"},
```
edeyoung: @alexanderelliott121 it looks like kristian added this here so the id would parse nicely. is this something that you can ensure in the fixture generation code?

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
+require 'test_helper'
+
+class MeasuresControllerMeasureHistoryTest < ActionController::TestCase
+include Devise::TestHelpers
+
+  tests MeasuresController
+
+  setup do
+    @error_dir = File.join('log', 'load_errors')
+    FileUtils.rm_r @error_dir if File.directory?(@error_dir)
```
edeyoung: is `@error_dir` used anywhere? can these lines be deleted?

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    dump_database
```
edeyoung: It looks like the upper code is temporary code as a place holder until the fixtures get re-generated (makes sense). Can you add a comment explaining that? it will help remind us to do the update.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
+require 'test_helper'
```
edeyoung, line 1: thoughts on moving this into `meaures_controller_measure_history_test`?

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    measure = Measure.where(hqmf_set_id: '9A032D9C-3D9B-11E1-8634-00237D5BF174', user_id: @user.id).first
+    assert_nil measure
+    # Assert that there is nothing in the archive for this measure
+    archive = ArchivedMeasure.where(hqmf_set_id: '9A032D9C-3D9B-11E1-8634-00237D5BF174', user_id: @user.id)
+    assert_equal 0, archive.count
+
+    # Load the measure
+    post :create, measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional'
+    measure = Measure.where(cms_id: 'CMS68v4', user_id: @user.id).first
+    assert_equal '40280381-4555-E1C1-0145-DC7DC26A44BF', measure.hqmf_id
```
edeyoung: what about adding an `hqmf_set_id` assert here? the above assertion that the measure is not yet loaded uses the `hqmf_set_id`. Using it here would directly demonstrate the counter to the previous test.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    assert_equal 0, archive.count
+
+    # Get the ObjectId of the current version of the measure
+    measure_first_load_db_id = measure.id
+
+    # Go the next version
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'base_example', 'CMS68v6.zip'), 'application/zip')
+    post :create, measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional', hqmf_set_id: '9A032D9C-3D9B-11E1-8634-00237D5BF174'
+    measure = Measure.where(hqmf_id: '40280381-51F0-825B-0152-227DFBAC15AA', user_id: @user.id).first
+    assert_equal '40280381-51F0-825B-0152-227DFBAC15AA', measure[:hqmf_id]
```
edeyoung: same comment here about checking `hqmf_set_id`.

also, i like things parallel. above, there's `measure.hqmf_id`. here, there's `measure[:hqmf_id]`. these should be the same.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    # associate_user_with_patients(@user, @patients)
+    # Record.each do |patient|
+    #   if patient['user_id'].is_a?(String)
+    #     patient['user_id'] = BSON::ObjectId.from_string(patient['user_id'])
+    #     patient.save!
+    #   end
+    # end
+  end
+
+  test 'measure archiving' do
```
edeyoung: is there any point to doing this for each measure type? i'm leaning toward no. curious in what you think.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    # Load version 1 of the measure. Starts with two population sets
```
edeyoung: i don't think this comment is necessary

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    # Assert that there is now an entry in the archive for this measure
+    archive = ArchivedMeasure.where(hqmf_set_id: '9A032D9C-3D9B-11E1-8634-00237D5BF174', user_id: @user.id)
+    assert_equal 1, archive.count
+    assert_equal measure_first_load_db_id, archive[0][:measure_db_id]
+    assert_equal 'CMS68v4', archive[0][:measure_content][:cms_id]
+
+  end # measure archiving
+
+  test 'measure upload summary' do
+    # Load version 1 of the measure. Starts with two population sets
```
edeyoung: I don't think this comment is necesary

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    measure = Measure.where(hqmf_set_id: '9A032D9C-3D9B-11E1-8634-00237D5BF174', user_id: @user.id).first
+    assert_nil measure
+    # Assert that there is nothing in the upload summary for this measure
+    upload_summary = UploadSummary::MeasureSummary.where(hqmf_set_id: '9A032D9C-3D9B-11E1-8634-00237D5BF174', user_id: @user.id)
+    assert_equal 0, upload_summary.count
+
+    # Load the measure
+    post :create, measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional'
+    measure = Measure.where(cms_id: 'CMS68v4', user_id: @user.id).first
+    assert_equal '40280381-4555-E1C1-0145-DC7DC26A44BF', measure.hqmf_id
```
edeyoung: same thought as above with `hqmf_set_id`

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    # Assert that there is information for the "post" upload portion
```
edeyoung: can you add to the comment something like, "and there is no "pre" upload portion."

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    @patients.each do |patient|
+      assert_equal measure.hqmf_set_id, patient[:measure_ids][0]
+      patient['user_id'] = BSON::ObjectId.from_string(patient['user_id']) if patient['user_id'].is_a?(String)
+      patient.unset(:calc_results)
+      patient.save!
+    end
+
+    calculator = BonnieBackendCalculator.new
+    patients = Record.where(user_id: measure.user_id, measure_ids: measure.hqmf_set_id)
+    # Move on if there are no patients associated with this measure for this user
```
edeyoung: didn't we just load patients? if there are no patients, should we error out?

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    # Load the patients for the measure
+    records_set = File.join('records','measure_history_set', 'base_example')
+    collection_fixtures(records_set)
+    @patients = Record.where(measure_ids: '9A032D9C-3D9B-11E1-8634-00237D5BF174')
+    associate_user_with_patients(@user, @patients)
+    @patients.each do |patient|
+      assert_equal measure.hqmf_set_id, patient[:measure_ids][0]
+      patient['user_id'] = BSON::ObjectId.from_string(patient['user_id']) if patient['user_id'].is_a?(String)
+      patient.unset(:calc_results)
+      patient.save!
```
edeyoung: is this being unset and saved to ensure that the latest calc results are used? if so, can you add that in a comment. If not, can you comment on why calc results are being unset?

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    end
+
+    calculator = BonnieBackendCalculator.new
+    patients = Record.where(user_id: measure.user_id, measure_ids: measure.hqmf_set_id)
+    # Move on if there are no patients associated with this measure for this user
+    if patients.count == 0
+      next
```
edeyoung: what is being `next`ed out of?

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    end
+    measure.populations.each_with_index do |population_set, population_index|
+      begin
```
edeyoung: we don't want a rescue here. we want to explicitly fail in our test if something unexpected happens.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+        calculator.set_measure_and_population(measure, population_index, clear_db_cache: true, rationale: true)
+      rescue => e
+        puts "\nError for #{measure.user.email if measure.user?} measure #{measure.cms_id} population set #{population_index}:"
+        puts "Calculator setup exception: #{e.message}."
+        next
+      end
+      processed_patients_array = []
+      patients.no_timeout.each_with_index do |patient, patient_index|
```
edeyoung: `no_timeout` shouldn't be necessary here. we have a controlled small set of patients.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+
+        # For some reason, patients show up multiple times during this iteration.
```
edeyoung: we don't need to do this check here.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+        # This checks for that and skips those patients to reduce the number of calculations.
+        if processed_patients_array.include?(patient)
+          next
+        end
+        processed_patients_array << patient
+
+        begin
```
edeyoung: no rescues for the test.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    measure_first_load_cms_id = measure.cms_id
+    measure_first_load_hqmf_version_number_pre_upload = measure.hqmf_version_number
+
+    # In order for the snapshot to persist it needs to be saved to an array
+    measure_patients_pre_upload = Record.where(measure_ids: '9A032D9C-3D9B-11E1-8634-00237D5BF174', user_id: @user.id).to_a
+
+    # Load the next version of the measure
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'base_example', 'CMS68v6.zip'), 'application/zip')
+    post :create, measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional', hqmf_set_id: '9A032D9C-3D9B-11E1-8634-00237D5BF174'
+    measure = Measure.where(hqmf_id: '40280381-51F0-825B-0152-227DFBAC15AA', user_id: @user.id).first
```
edeyoung: should this be accessed with `cms_id` like it is earlier?

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    assert_equal '40280381-51F0-825B-0152-227DFBAC15AA', measure.hqmf_id
```
edeyoung: see above re checking `hqmf_set_id`

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    assert_equal upload_summary.cms_id_pre_upload, measure_first_load_cms_id
+    assert_equal upload_summary.cms_id_post_upload, measure.cms_id
+    assert_equal upload_summary.hqmf_version_number_pre_upload, measure_first_load_hqmf_version_number_pre_upload
+    assert_equal upload_summary.hqmf_version_number_post_upload, measure.hqmf_version_number
+    assert_equal 1, upload_summary.population_set_summaries.count
+    assert_equal 1, upload_summary.measure_population_set_count[:pre_upload]
+    assert_equal 1, upload_summary.measure_population_set_count[:post_upload]
+    # Test the summary fields
+    assert_not_nil upload_summary.population_set_summaries[0][:summary]
+    assert_equal 2, upload_summary.population_set_summaries[0][:summary][:pass_before]
```
edeyoung: in a perfect world, one fixture would pass before and fail after. although, this might be the measure that is so simple that there are no changes across versions.

edeyoung (changes requested):
I have more to review.

This is looking very good! I have some minor comments but nothing major.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    assert_equal 2, upload_summary.population_set_summaries[0][:summary][:pass_after]
+    assert_equal 0, upload_summary.population_set_summaries[0][:summary][:fail_before]
+    assert_equal 0, upload_summary.population_set_summaries[0][:summary][:fail_after]
+    # Test the patient fields
```
edeyoung: add whitespace here

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    measure_patients_post_upload = Record.where(measure_ids: '9A032D9C-3D9B-11E1-8634-00237D5BF174', user_id: @user.id)
```
edeyoung: is it worth making this an array too so it can be accessed in the same way as the`measure_patients_pre_upload`?

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+    assert_equal 2, measure_patients_post_upload.count
+    upload_summary.population_set_summaries[0][:patients].each do |patient_id, patient|
+      pre_load_patient = (measure_patients_pre_upload.select { |r| r.id.to_s == patient_id})[0]
+      post_load_patient = measure_patients_post_upload.where(id: patient_id).first
+      assert_not_nil pre_load_patient
+      assert_not_nil post_load_patient
+
+      assert_equal 8, patient.count
```
edeyoung: can you add a comment here? It seems odd to check the count of a hash. is the idea to ensure that you don't have any more fields than the fields that you are testing for?

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
+require 'test_helper'
+
+class PatientTest < ActionController::TestCase  #ActiveSupport::TestCase
+  include Devise::TestHelpers
+  
+  tests MeasuresController
+  # TODO: See if the can be set to tests Record
```
edeyoung: reword comment - doesn't make sense as is.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'multi_population', 'CMS160v3.zip'), 'application/zip')
+    # Load the measure
+    post :create, measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional'
+
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'update_change', 'CMS104v3.zip'), 'application/zip')
+    # Load the measure
+    post :create, measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional'
+
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'continuous_variable', 'EH_CMS32v4.zip'), 'application/zip')
+    # Load the measure
```
edeyoung: add whitespace above this line.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    @error_dir = File.join('log', 'load_errors')
+    FileUtils.rm_r @error_dir if File.directory?(@error_dir)
+    dump_database
+    users_set = File.join('users', 'base_set')
+    collection_fixtures(users_set)
+    @user = User.by_email('bonnie@example.com').first
+    sign_in @user
+
+    # TODO: When measure fixtures are available switchh to loading them
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'base_example', 'CMS68v6.zip'), 'application/zip')
```
edeyoung: This also needs to be changed in measures_controller_measure_history:

`calculation_type` should be `patient` or `episode`. You are currently using `proportional`.

For CMS68: should be `episode` with `Occurrence A of Encounter, Performed: Medications Encounter Code Set` (or closest if this is encompassed by a variable)
For CMS160: should be `patient`
For CMS104: should be `episode` with `Encounter, Performed: Non-Elective Inpatient Encounter` (or closest if this is encompassed by a variable). *104 is also an EH measure not an EP measure*
For CMS32: should be `episode` with `Encounter, Performed: Emergency Department Visit` (or closest if this is encompassed by a variable) (it looks like you've already done this).

You can see if it's `patient` vs `episode` from the .html file. If there's a "Measure Item Count" field, then it's `episode`. That field describes what the`episode` is. Otherwise, it's `patient`.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'update_change', 'CMS104v3.zip'), 'application/zip')
+    # Load the measure
+    post :create, measure_file: measure_file, measure_type: 'ep', calculation_type: 'proportional'
+
+    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'continuous_variable', 'EH_CMS32v4.zip'), 'application/zip')
+    # Load the measure
+    post :create, measure_file: measure_file, measure_type: 'eh', calculation_type: 'episode'
+    post :finalize, {"t551"=>{"hqmf_id"=>"40280381-43DB-D64C-0144-6A8B0A3B30C6", "episode_ids"=>["OccurrenceAEmergencyDepartmentVisit1"], "titles"=>{"0"=>"Population 1 k", "1"=>"Stratification 1 k", "2"=>"Stratification 2 k", "3"=>"Stratification 3 k", "4"=>"foo"}}}
+
+    @golden_results = {}
```
edeyoung: add comment explaining what the `@golden_results` are

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+
+    records_set = File.join('records', "measure_history_set", "base_example")
```
edeyoung: oy. this is making me realize that the patient fixtures may not be the best test patient fixtures. There are a lot of zeros down there.
when the fixtures get regenerated, can you create the test patients similar to how they're done for the hand tests?

Also, let's chat before those fixtures get generated. I think we should have a game plan on when we create the patients, etc.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    collection_fixtures(records_set)
+    @golden_results.store(BSON::ObjectId('5852fa58e76e94e528001e9d'), [ {'IPP'=> 0, 'DENOM'=> 0, 'DENEX'=> 0, 'NUMER'=> 0, 'NUMEX'=> 0, 'DENEXCEP'=> 0, 'status'=> 'fail'} ] )
+
+    records_set = File.join('records', "measure_history_set", "continuous_variable")
+    collection_fixtures(records_set)
+    # TODO
+    # Add the golden results for the continuous variable measure
+
+    associate_user_with_measures(@user, Measure.all)
+    # associate_measures_with_patients([@measure], Record.all)
```
edeyoung: should this be deleted?

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+  setup do
+    @error_dir = File.join('log', 'load_errors')
+    FileUtils.rm_r @error_dir if File.directory?(@error_dir)
+    dump_database
+    users_set = File.join('users', 'base_set')
+    collection_fixtures(users_set)
+    @user = User.by_email('bonnie@example.com').first
+    sign_in @user
+
+    # TODO: When measure fixtures are available switchh to loading them
```
edeyoung: consider breaking some of the functionalities below into their own methods. that will make the code easier to read. For example, you could have:

```
  setup do
    @error_dir = File.join('log', 'load_errors')
    FileUtils.rm_r @error_dir if File.directory?(@error_dir)
    dump_database
    users_set = File.join('users', 'base_set')
    collection_fixtures(users_set)
    sign_in @user

    @load_measures
    @load_patients
    @get_golden_results_for_patients
  end

  def load_measures
    measure_file = fixture_file_upload(File.join('test', 'fixtures', 'measure_exports', 'measure_history_set', 'base_example', 'CMS68v6.zip'), 'application/zip')
    ....
  end

  ....
```

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    post :finalize, {"t551"=>{"hqmf_id"=>"40280381-43DB-D64C-0144-6A8B0A3B30C6", "episode_ids"=>["OccurrenceAEmergencyDepartmentVisit1"], "titles"=>{"0"=>"Population 1 k", "1"=>"Stratification 1 k", "2"=>"Stratification 2 k", "3"=>"Stratification 3 k", "4"=>"foo"}}}
+
+    @golden_results = {}
+
+    records_set = File.join('records', "measure_history_set", "base_example")
+    collection_fixtures(records_set)
+    @golden_results.store(BSON::ObjectId('58514916e76e941a3d000489'), [{'IPP'=> 1, 'DENOM'=> 1, 'NUMER'=> 1, 'DENEX'=> 0, 'DENEXCEP'=> 0, 'NUMEX'=> 0, 'status'=> 'pass'}])
+    @golden_results.store(BSON::ObjectId('58514954e76e941a3d0004c4'), [{'IPP'=> 0, 'DENOM'=> 0, 'DENEX'=> 0, 'NUMER'=> 0, 'NUMEX'=> 0, 'DENEXCEP'=> 0, 'status'=> 'pass'}])
+
+    records_set = File.join('records', "measure_history_set", "multi_population")
```
edeyoung: consider adding a comment above each section saying, "loading patients for CMS104" etc.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+
+    # TODO: Make this a method
+    calculator = BonnieBackendCalculator.new
+    patients = Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id)
+    # Move on if there are no patients associated with this measure for this user
+    if patients.count == 0
+      flunk "No patients for measure #{measure.cms_id}"
+    end
+    measure.populations.each_with_index do |population_set, population_index|
+      begin
```
edeyoung: see notes in previous review about rescue statements. they shouldn't be used in tests. same for any other rescue statements.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+      assert_not patient[:results_exceed_storage], "Failed results_exceed_storage == false on patient name #{patient.first} #{patient.last} for measure #{measure.cms_id}"
+      assert_equal 4, patient[:results_size], "Failed results_size == 4 on patient name #{patient.first} #{patient.last} for measure #{measure.cms_id}"
+    end
+
+    # TODO: Make this a method
+    calculator = BonnieBackendCalculator.new
+    patients = Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id)
+    # Move on if there are no patients associated with this measure for this user
+    if patients.count == 0
+      flunk "No patients for measure #{measure.cms_id}"
```
edeyoung: is `flunk` what should be used here? do we want to fail if there are no patients? Is the idea that we assume there always are patients so if there aren't, that's a problem?

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    end
+    measure.populations.each_with_index do |population_set, population_index|
+      begin
+        calculator.set_measure_and_population(measure, population_index, clear_db_cache: true, rationale: true)
+      rescue => e
+        puts "\nError for #{measure.user.email if measure.user?} measure #{measure.cms_id} population set #{population_index}:"
+        puts "Calculator setup exception: #{e.message}."
+        next
+      end
+      processed_patients_array = []
```
edeyoung: don't need to track this for tests.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+        rescue => e
+          puts "\nError for #{measure.user.email if measure.user?} measure #{measure.cms_id} population set #{population_index} patient '#{patient.first} #{patient.last}' (_id: ObjectId('#{patient.id}')):"
+          puts "Calculation exception: #{e.message}."
+          next # Move onto the next patient
+        end
+      end
+    end
+
+    # TODO: This list needs to be maintained.
+    calc_results_keys = ['IPP', 'rationale', 'finalSpecifics', 'measure_id', 'population_index', 'status', 'values', 'STRAT']
```
edeyoung: it's the same either way, but my slight preference is to move `IPP` out of this list and put it in both the lists below. That makes the lists make more sense to me. The initial list is all the metadata items. The lists being added are all the population sets for each group.

what happens if you put all the calc_results_keys together? like, the `MSRPOPL` with the `DENOM` and don't have two different lists? would it still work? if so, that might be the easiest thing to do. Then would this just be the same as unioning that master populations list (which is used in the code a bunch) with the extras list (`rationale`, `finalSpecifics`, etc.)?

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    unless measure.continuous_variable
+      calc_results_keys = calc_results_keys + ['DENOM', 'DENEX', 'NUMER', 'NUMEX', 'DENEXCEP']
+    else
+      calc_results_keys = calc_results_keys + ['MSRPOPL', 'MSRPOPLEX']
+    end
+
+    Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id).each do |patient|
+      assert_not_nil @golden_results[patient.id], "Missing golden result for patient name #{patient.first} #{patient.last} (patient.id: #{patient.id.to_s}) for measure #{measure.cms_id}."
+      assert_equal measure.populations.length, patient[:calc_results].count, "Failed calc_results count on patient name #{patient.first} #{patient.last} for measure #{measure.cms_id}"
```
edeyoung: I'm not sure this failure message adequately explains the failure condition. This is similar for the other messages as well.

Consider changing to something like, "patient #{patient.first} #{patient.last} for measure #{measure.cms_id} did not have same number of calc results as there are populations. Number of calc_results: #{...}. Number of populations: #{...}."

Take a look at the other messages to see if they should be changed. If there is a fail message (which I think is awesome that you added), then have it be explanatory. For example, I don't known what "failed measure_id
 means, but I do know what "mismatched measure_id for patient calculation results and given measure" (or something along those lines.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+      assert_operator 4, :<, patient[:results_size]
```
edeyoung: add comment explaining this. it's a pretty arbitrary number. people should understand where it comes from.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+      assert patient[:results_exceed_storage] if patient[:results_size] > APP_CONFIG['record']['max_size_in_bytes']
+      assert_not patient[:results_exceed_storage] if patient[:results_size] <= APP_CONFIG['record']['max_size_in_bytes']
+      assert_not_nil patient[:calc_results], "Failed not_nil calc_results exists on patient name #{patient.first} #{patient.last} for measure #{measure.cms_id}"
```
edeyoung: you check `not_nil` here but earlier you access `.count` for `calc_results`. The `not_nil` check should precede accessing attributes.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+
+    # TODO: This list needs to be maintained.
+    calc_results_keys = ['IPP', 'rationale', 'finalSpecifics', 'measure_id', 'population_index', 'status', 'values', 'STRAT']
+    unless measure.continuous_variable
+      calc_results_keys = calc_results_keys + ['DENOM', 'DENEX', 'NUMER', 'NUMEX', 'DENEXCEP']
+    else
+      calc_results_keys = calc_results_keys + ['MSRPOPL', 'MSRPOPLEX']
+    end
+
+    Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id).each do |patient|
```
edeyoung: white space is your friend with this type of thing. Use it liberally to separate concepts to make this more readable.

Consider something along the lines of:

```
// golden results - assertion and getting the golden results

// all things related to results size

// all things related to calc_results

// loop over populations
  // calc_results assertions

  // golden results assertions
```

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    # TODO
+    # Add the golden results for the continuous variable measure
+
+    associate_user_with_measures(@user, Measure.all)
+    # associate_measures_with_patients([@measure], Record.all)
+    associate_user_with_patients(@user, Record.all)
+
+  end
+
+  def calculation_results (measure)
```
edeyoung: add comments to the top of each of these tests explaining at a high level what they'll be testing and how.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+        assert_not_nil @golden_results[patient.id][population_index], "Missing golden result for patient name #{patient.first} #{patient.last} (patient.id: #{patient.id.to_s}) for measure #{measure.cms_id} for population_set: [#{population_index}]."
+        @golden_results[patient.id][population_index].each do |result_population, result_value|
+          assert_equal result_value, patient.calc_results[population_index][result_population], "Failed correct calc_result on patient name #{patient.first} #{patient.last} (patient.id: #{patient.id.to_s}) for measure #{measure.cms_id} for population_set: [#{population_index}], population: [#{result_population}].\n#{patient.calc_results[population_index]}"
+        end
+      end
+    end
+
+  end
+
+  def expected_values_empty_properly_populated (measure)
```
edeyoung: I don't understand the name for this function. can you explain and possibly reword?

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    # Add the golden results for the continuous variable measure
+
+    associate_user_with_measures(@user, Measure.all)
+    # associate_measures_with_patients([@measure], Record.all)
+    associate_user_with_patients(@user, Record.all)
+
+  end
+
+  def calculation_results (measure) 
+    # Remove any calc results and related fields that might have been loaded
```
edeyoung: this explains the what but not the why. can you add a note on why you're removing the calc results?

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+          assert_equal result_value, patient.calc_results[population_index][result_population], "Failed correct calc_result on patient name #{patient.first} #{patient.last} (patient.id: #{patient.id.to_s}) for measure #{measure.cms_id} for population_set: [#{population_index}], population: [#{result_population}].\n#{patient.calc_results[population_index]}"
+        end
+      end
+    end
+
+  end
+
+  def expected_values_empty_properly_populated (measure)
+
+    # Set the expected values to be empty
```
edeyoung: can you explain the why as well as the what here?

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+    Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id).each do |patient|
+      patient.expected_values = []
+      patient.save!
+    end
+
+    Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id).each do |patient|
+      assert patient[:expected_values].empty?, "Failed empty expected_values on patient name #{patient.first} #{patient.last} for measure #{measure.cms_id}"
+      patient.update_expected_value_structure!(measure) 
+      assert_not patient[:expected_values].nil?, "Failed not nil expected_values on patient name #{patient.first} #{patient.last} for measure #{measure.cms_id}"
+      # Test that the number of populations match
```
edeyoung: every time you have a comment, add a newline ahead of it. it helps visually separate the logic and makes the code way easier to read.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+      assert_equal measure.populations.length, patient.expected_values.length
+      measure.populations.each_with_index do |population_set, population_index|
```
edeyoung: when you go into loops, unless the line above is directly related to the loop (like initiates a variable used in the loop), add a newline ahead of it.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+        # Test that the population set is filled with the correct populations
+        assert_equal [:measure_id, :population_index], (patient[:expected_values][population_index].keys - population_set.keys), "Failed population set is filled with the correct populations on patient name #{patient.first} #{patient.last} for measure #{measure.cms_id}"
```
edeyoung: may be worth noting that we expect `population_set` to have keys that `expected_values` doesn't have, but not the other way around.

edeyoung: isn't `expected_values` the structure where the population sets aren't necessarily in order? if that's the case, we can't rely on `population_index` for this array.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+
+    # Set the expected values to be empty
+    Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id).each do |patient|
+      patient.expected_values = []
+      patient.save!
+    end
+
+    Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id).each do |patient|
+      assert patient[:expected_values].empty?, "Failed empty expected_values on patient name #{patient.first} #{patient.last} for measure #{measure.cms_id}"
+      patient.update_expected_value_structure!(measure)
```
edeyoung: consider adding a comment that (a) this is the functionality that you are particularly testing with this code and (b) you are expecting this to add in all relevant populations and initialize them to 0 or [] as appropriate

edeyoung (changes requested):
I still have a bit more to review.

as stated earlier, looks good! the comments are pretty minor. My biggest thoughts: watch your copy/pasting and determine what actually should be included; watch your whitespace; make your comments explain the why as well as the what.

test/functional/measures_controller_measure_history_granular_test.rb
mulcahyk: I don't mind doing that.  I think we are going to make `meaures_controller_measure_history_test` rather long doing that.  To help keep it clear I will make a "section" for these tests with some comments to identity their purpose.

For the sake of tracking the comments, this will be the last change that I make.

test/functional/measures_controller_measure_history_granular_test.rb
mulcahyk: Just a bit confused about this comment.  Are you referring to:
`users_set = File.join('users', 'base_set')`
`collection_fixtures(users_set)`
If so, then yes, this is temporary until the final fixtures are ready.
I will add the TODO comment to replace this.

test/functional/measures_controller_measure_history_granular_test.rb
mulcahyk: I don't really think that it matters.  The archive process just copies the data from one collection to another in the database.  Add other measure types would be just to further test the copy functionality of Ruby, Rails, and MongoDB.

test/functional/measures_controller_measure_history_granular_test.rb
mulcahyk: I think that is the case here.
image

test/functional/measures_controller_measure_history_granular_test.rb
mulcahyk: The `measure_patients_pre_upload` is kludge way of accessing the data.  I would rather not propagate that.  I know that you want things to be the same everywhere but I don't think it works here.  If you think it would help with readability then I'm happy to change it.  Otherwise I think a comment would be enough.

test/functional/patient_expected_and_calc_values_test.rb
mulcahyk: This comment comes from a conversation we had about the testing of measure history.  I was using `tests MeasuresController` so that I could load the measures from zip files.  All of these tests though have more to do with what happens at the `record` level.  This comment was to change the controller we are using.  You wanted me to come back to it after I got this working.

test/functional/measures_controller_measure_history_granular_test.rb
edeyoung: Yes, that's what I was referring to. Thanks!

test/functional/measures_controller_measure_history_granular_test.rb
edeyoung: Agree - let's just do it for one type.

test/functional/measures_controller_measure_history_granular_test.rb
edeyoung: yes, it is. bummer.

test/functional/measures_controller_measure_history_granular_test.rb
```diff
@@ -0,0 +1,215 @@
...
+
+    # Assert that there is now an entry in the archive for this measure
+    archive = ArchivedMeasure.where(hqmf_set_id: '9A032D9C-3D9B-11E1-8634-00237D5BF174', user_id: @user.id)
+    assert_equal 1, archive.count
+    assert_equal measure_first_load_db_id, archive[0][:measure_db_id]
+    assert_equal 'CMS68v4', archive[0][:measure_content][:cms_id]
+
+  end # measure archiving
+
+  test 'measure upload summary' do
```
edeyoung, line 59: should the upload summary be done for multiple measure types?

test/functional/measures_controller_measure_history_granular_test.rb
edeyoung: leaving it makes sense. like you suggested, a comment might be nice.

test/functional/patient_expected_and_calc_values_test.rb
edeyoung: The comment doesn't reflect that very well (I think there are some typos in it). Can you reword the comment to reflect what you wrote above? It doesn't have to be that verbose - can still be a one-liner, but as it is, it doesn't make sense.

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+          end
+        end
+      end
+    end
+
+  end
+
+  def expected_values_values_retained (measure)
+
+    Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id).each do |patient|
```
edeyoung: Add a comment explaining what you're doing here and why

test/functional/patient_expected_and_calc_values_test.rb
```diff
@@ -0,0 +1,287 @@
...
+      measure.populations.each_with_index do |population_set, population_index|
+        (population_set.keys- ['title', 'id']).each_with_index do |population, index|
+          patient.expected_values[population_index][population] = 999000 + (population_index * 100) + index unless population == 'OBSERV' || population == 'OBSERV_UNIT'
+          patient.save!
+        end
+      end
+    end
+
+    Record.where(user_id: @user.id, measure_ids: measure.hqmf_set_id).each do |patient|
+      patient.update_expected_value_structure!(measure)
```
edeyoung, line 225: is this the main functionality being tested? If so, note that.

test/functional/patient_expected_and_calc_values_test.rb
mulcahyk: Good point.  I'll set it the same way as the code.

test/functional/measures_controller_measure_history_granular_test.rb
mulcahyk: Is it worth it to make a measure with the MAT to test this as opposed to sorting through all of the existing measures and hoping we hit on one?

test/functional/patient_expected_and_calc_values_test.rb
mulcahyk: Do you think that it is over to have the `golden_results` method verify that there is an entry for every patient record that has been loaded?

mulcahyk:
Please see PR #683 for the final version.

edeyoung:
This pull request was maintained for comments in support of https://github.com/projecttacoma/bonnie/pull/683. https://github.com/projecttacoma/bonnie/pull/683 has been accepted and will be merged so this pull request is no longer necessary.

PR #672, opened 2017-01-24
Tweaks to data criteria display

c-monkey:
- revised attribute criteria to use value set display name
- added negation rationale to data criteria

For example, in CMS100V5:
- in addition to `Medication, Administered: Aspirin` (oid 2.16.840.1.113883.3.666.5.626), added is: `Medication, Administered: Aspirin (Negation Rationale)` (oid: 2.16.840.1.113883.3.666.5.1145)
- the attribute value set `Acute Myocardial Infarction (AMI) Grouping Value Set` now shows up as `Acute Myocardial Infarction (AMI)`

### DIFF
--- a/app/assets/javascripts/views/measure_value_sets.js.coffee
+++ b/app/assets/javascripts/views/measure_value_sets.js.coffee
@@ -98,7 +98,7 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
           # might contain negation rationale
           if criteria.negation_code_list_id? and criteria.negation
             displayed_item_name = @model.valueSets().findWhere({oid: criteria.negation_code_list_id})?.get('display_name') ? criteria.description
-            displayed_item_name = displayed_item_name + " (Negation Rationale)"
+            displayed_item_name = valueSet.name + " (Not Done: " + displayed_item_name + ")"
             # added another property to hold the negation criteria cid because criteria already has cid
             criteria.negationCidHolder = cid: undefined
             dataCriteria.push(@._createValueSet(criteria.negationCidHolder, criteria.negation_code_list_id, displayed_item_name))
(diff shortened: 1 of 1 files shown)


edeyoung (changes requested):
As it is now, the codes look like:

Physical Exam, Performed: BMI LOINC Value (Negation Rationale)

But this doesn't actually show the name of the value set. In this case, the name of the value set is: "Patient Reason refused"

You should be able to access "Patient Refused" from the same measure value set list that your accessing the other codes through.

I think there are two ways this could look in our value set list.

1. As it looks in the logic: "Physical Exam, Performed: BMI LOINC Value (Not Done: Patient Reason refused)"

2. As it looks in MAT exported human readable: ""Physical Exam, Performed not done: Patient Reason refused"

c-monkey:
What a strange oversight on my part! I think I was hoping to refactor the 3 similar instances into shared code but they are just dissimilar enough that I kept them separate. That being said, this change was a 1-liner.

edeyoung:
What do you think showing it like these?

1. As it looks in the logic: "Physical Exam, Performed: BMI LOINC Value (Not Done: Patient Reason refused)"

2. As it looks in MAT exported human readable: ""Physical Exam, Performed not done: Patient Reason refused"

The current representation is different than what users will see elsewhere

c-monkey:
Fabulous suggestion, dunno why I didn't do that originally. Thanks for the great feedback. Now the data criteria text matches the logic text exactly. That's obviously preferred.

edeyoung:
Another thought -

In the MAT, they have:
"Reason: Overweight"

We have that information (the `key_title` in the `value` of `criteria.field_values`). Currently, we just show "Overweight"

Also, the MAT intersperses the "Not Done" with the data criteria value sets, but these are actually attribute value sets. Thoughts on moving these down to the "attributes" section?

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -82,14 +84,33 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
 
           # if not supplemental criteria, may contain attributes in "value" property
           # we show attributes that have type="CD", a non-empty title, and non-empty code_list_id
-          if criteria.value?.type? and criteria.value.type is "CD" and criteria.value.title? and criteria.value.code_list_id?
-            attributesCriteria.push(@._createValueSet(criteria.value, criteria.value.title))
+          if criteria.value?.type? and criteria.value.type is "CD" and criteria.value.code_list_id?
```
edeyoung, line 87: These should have `Result: ` at the beginning of them.

c-monkey:
(see comments on commit)

edeyoung (approved):

PR #674, opened 2017-01-27
using result.patient.id rather than result.get('patient_id')

c-monkey:
In addition to clone being broken, so were the adjacent buttons


edeyoung:
should this include a test for these? Is there an easy way to create a test for this?

c-monkey:
I don't think we can test this with teaspoon, but I would be happy to be proven wrong. @mayerm94 can you think of a way to test creating a patient in the patient builder and seeing the results in the PopulationCalcuation?

edeyoung:
It looks like patient_id was no longer being added to results, likely because of how cached results changes the logic flow. The code where patient_id was added to results was under a conditional (I don't know why) in app/assets/javascripts/calculator.js.coffee:

```
    # If the result is already present on the patient we will use that.
    if patient.getCalculatedResultsValues(population)
      result.set(patient.getCalculatedResultsValues(population))
      result.set({patient_id: patient.id})
      result.state = 'complete'
      return result
```

If this gets moved outside of the `if` block (I tried it under this line: `result = @resultsCache[cacheKey] ?= new Thorax.Models.Result({}, population: population, patient: patient)`), this seems to also address the problem.

I think `patient_id` was added to the results object because as an attribute, it sticks with the results object. I don't think items added directly to the object stick to it. This makes it queryable in a Thorax-y way. For example, in measure_patient_dashboard, there's `@results.where({patient_id: patient_id})`.

Thoughts on just moving this line rather than converting to `result.patient.id`?

c-monkey:
I like where you're coming from, but will defer to @hossenlopp

c-monkey:
@hossenlopp Please review, my original changes are reverted and the sum total of this patch is moving 1 line as suggested by Lizzie.

hossenlopp (approved):

PR #675, opened 2017-01-31
addressed issue where teaspoon failing if measure doesn't have 'value…

edeyoung:
…_sets' associated with it.


alexanderelliott121 (approved):
Everything looks to be in order and it runs.

PR #677, opened 2017-02-06
Updates backend test fixture loading to convert ids to bson objects, …

alexanderelliott121:
Updates test infrastructure to address issues raised by updating fixture loading.
Adding logic to convert id fields into bson objects conflicted with existing fixtures.  Those fixtures had to have their id's updated, and the tests that reference them as well.

This had the side effect of fixing a longstanding bug in the tests where value sets were being exported incorrectly due to inconsistencies with ids.

### DIFF
--- a/lib/tasks/generate_fixtures.rake
+++ b/lib/tasks/generate_fixtures.rake
@@ -1,6 +1,6 @@
 namespace :bonnie do
   namespace :fixtures do
-
+    require_relative "../../test/test_helper.rb"
     ###
     # Generates a set of front end fixtures representing a specific database state.
     #
@@ -178,54 +178,6 @@ namespace :bonnie do
         json.each_pair do |k,v|
           if k.ends_with?("_at")
             json[k] = Time.parse(v)
-          end
-        end
-      end
-    end
-
-    ###
-    # Parses json object for id fields and converts them to bson objects
-    #
-    # Code is derived from set_mongoid_ids function defined in test/test_helper.rb
-    # This version includes checks to convert specific fields into BSON objects for usability purposes.
-    #
-    # json: The json object to parse
-    def set_mongoid_ids(json)
-      if json.kind_of?( Hash)
-        json.each_pair do |k,v|
-          if v && v.kind_of?( Hash )
-            if v["$oid"]
-              json[k] = BSON::ObjectId.from_string(v["$oid"])
-            else
-              set_mongoid_ids(v)
-            end
-          elsif k == '_id' || k == 'bundle_id' || k == 'user_id'
-            json[k] = BSON::ObjectId.from_string(v)
-          end
-        end
-      end
-    end
-
-    ##
-    # Loads fixtures into the active database.
-    # Code is derived from collection_fixtures function found in test/test_helper.rb
-    # Code has been altered to address issues arising from inserting data into an active database with existing data.
-    #
-    # collection_names: array of paths leading to the relevant collections.
-    def collection_fixtures(*collection_names)
-      collection_names.each do |collection|
-        collection_name = collection.split(File::SEPARATOR)[0]
-        Mongoid.default_session[collection_name].drop
-        Dir.glob(File.join(Rails.root, 'test', 'fixtures', collection, '*.json')).each do |json_fixture_file|
-          fixture_json = JSON.parse(File.read(json_fixture_file))
-          if fixture_json.length > 0
-            convert_times(fixture_json)
-            set_mongoid_ids(fixture_json)
-            # Mongoid names collections based off of the default_session argument.
-            # With nested folders,the collection name is “records/X” (for example).
-            # To ensure we have consistent collection names in Mongoid, we need to take the file directory as the collection name.
-            puts "Inserting into collection #{collection}"
-            Mongoid.default_session[collection_name].insert(fixture_json)
           end
         end
       end
--- a/test/functional/admin_users_controller_test.rb
+++ b/test/functional/admin_users_controller_test.rb
@@ -128,7 +128,7 @@ include Devise::TestHelpers
       assert_equal 4, zip_file.glob(File.join('patients', '**', '*.json')).count
       assert_equal 3, zip_file.glob(File.join('sources', '**', '*.json')).count
       assert_equal 3, zip_file.glob(File.join('sources', '**', '*.metadata')).count
-      assert_equal 27, zip_file.glob(File.join('value_sets', '**', '*.json')).count
+      assert_equal 29, zip_file.glob(File.join('value_sets', '**', '*.json')).count
     end
     File.delete(zip_path)
   end
--- a/test/functional/admin_users_controller_test.rb
+++ b/test/functional/admin_users_controller_test.rb
@@ -17,7 +17,7 @@ include Devise::TestHelpers
     associate_user_with_measures(@user, Measure.all)
     associate_user_with_patients(@user, Record.all)
 
-    @user.measures.first.value_set_oids.uniq.each do |oid|
+    @user.measures.find('53ce63744d4d32e2cd4b0500').value_set_oids.uniq.each do |oid|
       vs = HealthDataStandards::SVS::ValueSet.new(oid: oid)
       vs.concepts << HealthDataStandards::SVS::Concept.new(code_set: 'foo', code:'bar')
       vs.user = @user
--- a/test/functional/populations_controller_test.rb
+++ b/test/functional/populations_controller_test.rb
@@ -35,12 +35,12 @@ class PopulationsControllerTest  < ActionController::TestCase
   test "update population" do
     sign_in @user
     # This particular test measure has multiple populations, and therefore can have their titles changed
-    measure = Measure.by_user(@user).find("40280381-3D61-56A7-013E-7AF612436402")
+    measure = Measure.by_user(@user).find("53ce63744d4d32e2cd4b0502")
     assert_equal("85 days",measure.populations.at("Populations1".to_i)['title'])
     # Change title of Population1 from "85 days" to "New Title Text"
-    post :update, {measure_id: "40280381-3D61-56A7-013E-7AF612436402",
+    post :update, {measure_id: "53ce63744d4d32e2cd4b0502",
       id: "Populations1", title: "New Title Text"}
-    measure = Measure.by_user(@user).find("40280381-3D61-56A7-013E-7AF612436402")
+    measure = Measure.by_user(@user).find("53ce63744d4d32e2cd4b0502")
     assert_equal("New Title Text",measure.populations.at("Populations1".to_i)['title'])
   end
 
--- a/test/functional/users_controller_test.rb
+++ b/test/functional/users_controller_test.rb
@@ -12,7 +12,7 @@ class UsersControllerTest  < ActionController::TestCase
     @user = User.by_email('bonnie@example.com').first
     associate_user_with_measures(@user, Measure.all)
     associate_user_with_patients(@user, Record.all)
-    @user.measures.first.value_set_oids.uniq.each do |oid|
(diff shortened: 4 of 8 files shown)


test/test_helper.rb
```diff
@@ -13,18 +13,49 @@ def dump_database
     end
   end
 
+  ###
+  # Parses json object for id fields and converts them to bson objects
+  #
+  # Code is derived from set_mongoid_ids function defined in test/test_helper.rb
```
edeyoung: update comment - this is when it was copy/pasted from this file into the new file you were working on.

test/test_helper.rb
```diff
@@ -13,18 +13,49 @@ def dump_database
...
+        elsif k == '_id' || k == 'bundle_id' || k == 'user_id'
+          json[k] = BSON::ObjectId.from_string(v)
+        end
+      end
+    end
+  end
+
+  ##
+  # Loads fixtures into the active database.
+  # Code is derived from collection_fixtures function found in test/test_helper.rb
```
edeyoung: update comment.

test/test_helper.rb
```diff
@@ -13,18 +13,49 @@ def dump_database
...
+  # Code has been altered to address issues arising from inserting data into an active database with existing data.
+  #
+  # collection_names: array of paths leading to the relevant collections.
   def collection_fixtures(*collection_names)
     collection_names.each do |collection|
-      Mongoid.default_session[collection].drop
+      collection_name = collection.split(File::SEPARATOR)[0]
+#      Mongoid.default_session[collection_name].drop
```
edeyoung: should this be commented? if so, why? if so, remove.

test/test_helper.rb
```diff
@@ -13,18 +13,49 @@ def dump_database
     end
   end
 
+  ###
```
edeyoung, line 16: by moving the modified functions into here, can you just remove them from wherever it is you modified them and use these instead?

test/functional/admin_users_controller_test.rb
```diff
@@ -128,7 +128,7 @@ class Admin::UsersControllerTest  < ActionController::TestCase
       assert_equal 4, zip_file.glob(File.join('patients', '**', '*.json')).count
       assert_equal 3, zip_file.glob(File.join('sources', '**', '*.json')).count
       assert_equal 3, zip_file.glob(File.join('sources', '**', '*.metadata')).count
-      assert_equal 27, zip_file.glob(File.join('value_sets', '**', '*.json')).count
+      assert_equal 29, zip_file.glob(File.join('value_sets', '**', '*.json')).count
```
edeyoung, line 131: why does the number of value sets increase?

edeyoung (changes requested):

test/functional/admin_users_controller_test.rb
alexanderelliott121: This one is weird.  The tests create value_sets to populate the measures.  29 value_sets are created, and they all have unique oids.
The 27 was a bug in the test, resulting from the way the fixtures were loaded.

I actually saw the same issue back when I was working on the vs consolidation.

test/test_helper.rb
alexanderelliott121: Should be removed.

test/functional/populations_controller_test.rb
```diff
@@ -35,12 +35,12 @@ class PopulationsControllerTest  < ActionController::TestCase
   test "update population" do
     sign_in @user
     # This particular test measure has multiple populations, and therefore can have their titles changed
-    measure = Measure.by_user(@user).find("40280381-3D61-56A7-013E-7AF612436402")
+    measure = Measure.by_user(@user).find("53ce63744d4d32e2cd4b0502")
```
edeyoung, line 38: shouldn't these be BSON ids?

test/functional/admin_users_controller_test.rb
edeyoung: if this was a bug in the test, why didn't the test fail previously? has anything changed about how the value sets are being loaded?

test/test_helper.rb
edeyoung: Can you address this comment? Should wherever these are being duplicated be deleted and these functions referenced instead?

test/test_helper.rb
```diff
@@ -13,18 +13,43 @@ def dump_database
...
+  end
+
+  ##
+  # Loads fixtures into the active database.
+  #
+  # collection_names: array of paths leading to the relevant collections.
   def collection_fixtures(*collection_names)
     collection_names.each do |collection|
-      Mongoid.default_session[collection].drop
+      collection_name = collection.split(File::SEPARATOR)[0]
```
edeyoung, line 42: why were we able to remove `Mongoid.default_session[collection].drop`?

edeyoung (changes requested):

test/test_helper.rb
alexanderelliott121: Done

test/functional/populations_controller_test.rb
alexanderelliott121: They get converted into BSON by the find method.

test/test_helper.rb
alexanderelliott121: The drop_database method that is called before most tests clears out the db.
I'm in favor of relying on drop_database over having the database clean up enforced in the loading function since I feel it opens up flexibility in execution: this way, you can load fixtures of the same type from multiple directories.

Example: lets say you want CMS160 and CMS68 from the measure_history_set fixtures, which have different paths.  Relying on loading for clean up would result in one being loaded, then being wiped before the other one is.

This does have the knock on effect of requiring a clean db for the fixture upload tool, but I think that that is acceptable.

test/functional/populations_controller_test.rb
alexanderelliott121: For some reason, I can't reply to the testing bug question or the "address comment" comment.
For the test: I'm a bit confused myself.  There are 29 unique oids, so it wasn't a matter of writing the same vs twice.

"Can you address this comment? Should wherever these are being duplicated be deleted and these functions referenced instead?"
I'm not sure what you're requesting.

test/test_helper.rb
edeyoung: ah, got it. thanks!

test/functional/populations_controller_test.rb
edeyoung: for the "can you address" comment - it was more or less a rephrasing of the first comment, which it looks like you addressed it (the functions were deleted in the other class). Maybe I missed this when I went through the second time or maybe it was happening right when I was commenting on it again? I think I thought you were done going through the comments when I wrote that.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,6 +1,6 @@
 namespace :bonnie do
   namespace :fixtures do
-
+    require_relative "../../test/test_helper.rb"
```
edeyoung, line 3: oh... interesting. these are in two completely different scopes, which i hadn't realized.

thoughts on how best to handle this? Is this creating a weird dependency? should the loading code be put somewhere else?

alexanderelliott121: It seems to be safe enough, given that it is in a rake task rather than in the main code.
The only people who will be interacting with it are devs.

lib/tasks/generate_fixtures.rake
```diff
@@ -1,6 +1,6 @@
...
     ###
     # Generates a set of front end fixtures representing a specific database state.
```
mulcahyk, line 5: I think that it might be helpful to have a comment describing the difference between front-end and back-end fixtures.  

Does it make sense to have one task with a parameter to specific front-end, back-end or both types of fixtures?

mulcahyk, line 5: I can't seem to leave the following comments where I want to so I'm lumping them all here:

- When making front-end fixtures why are all the patients dumped into one file while for the back-end they are put in separate files?
- [Here (line 73)](https://github.com/projecttacoma/bonnie/pull/677/files#diff-26c2c3094f9eb41aaa209e58589f491eR73) you have a "magic number/identifier" without explaining why or what it is used for.
- It seems like there is a fair bit of repeated code between the front and back end tasks.  Could they be converted to functions?
- I'm not clear on the purpose of the `generate_oid_to_valuesets` task.  Could you add a brief comment on when or why you would use this task?

test/functional/admin_users_controller_test.rb
```diff
@@ -17,7 +17,7 @@ class Admin::UsersControllerTest  < ActionController::TestCase
     associate_user_with_measures(@user, Measure.all)
     associate_user_with_patients(@user, Record.all)
 
-    @user.measures.first.value_set_oids.uniq.each do |oid|
+    @user.measures.find('53ce63744d4d32e2cd4b0500').value_set_oids.uniq.each do |oid|
```
mulcahyk, line 20: Magic identifier.  Why this id?

test/functional/populations_controller_test.rb
mulcahyk: Magic identifier.
If you are looking the measure up by BSON Id when why don't you do that directly?

mulcahyk, line 38: @edeyoung Do we have a standard way of looking objects up?  I like the exactness of doing it by BSON Id but that seems very brittle.  Could/should we be using something more immutable like cms_id or hqmf_id?

test/functional/users_controller_test.rb
```diff
@@ -12,14 +12,12 @@ class UsersControllerTest  < ActionController::TestCase
     @user = User.by_email('bonnie@example.com').first
     associate_user_with_measures(@user, Measure.all)
     associate_user_with_patients(@user, Record.all)
-
-    @user.measures.first.value_set_oids.uniq.each do |oid|
+    @user.measures.find('53ce63744d4d32e2cd4b0500').value_set_oids.uniq.each do |oid|
```
mulcahyk, line 15: Magic identifier.  Why this value?

test/test_helper.rb
```diff
@@ -13,18 +13,43 @@ def dump_database
...
-        # With nested folders,the collection name is “records/X” (for example).
-        # To ensure we have consistent collection names in Mongoid, we need to take the file directory as the collection name.
-        collection = collection.split(File::SEPARATOR)[0]
-        Mongoid.default_session[collection].insert(fixture_json)
+        if fixture_json.length > 0
+          convert_times(fixture_json)
+          set_mongoid_ids(fixture_json)
+          # Mongoid names collections based off of the default_session argument.
+          # With nested folders,the collection name is “records/X” (for example).
+          # To ensure we have consistent collection names in Mongoid, we need to take the file directory as the collection name.
```
mulcahyk: I'm not sure if it is a spacing issue or not but these comments are bit confusing.  I appreciate the background information but in this case I think for clarity it would be better to just say we are naming the collections after the file directory.

test/functional/admin_users_controller_test.rb
alexanderelliott121: I think using specific identifiers for tests is acceptable.  It isn't going to change unless the fixture is changed, and is more robust than the previous approach of just grabbing the first element in the collection.

alexanderelliott121: Point of clarification based on your other comment:
find converts the given argument into a bson and compares against _id

test/functional/users_controller_test.rb
alexanderelliott121: It is the bson id of the measure, ensuring that the measure retrieved is determanistic

test/functional/populations_controller_test.rb
alexanderelliott121: The id is stored in the fixture, so I wouldn't exactly call it mutable.

mulcahyk (approved):
The changes look good to me.

edeyoung (approved):

PR #678, opened 2017-02-07
Cql4bonnie qdm 5 0

edeyoung:


app/assets/javascripts/models/data_criteria.js.coffee
```diff
@@ -132,25 +139,60 @@ class Thorax.Models.PatientDataCriteria extends Thorax.Model
...
 
-  hasStopTime: ->
-    criteriaType = @get('definition')
-    return !(criteriaType in ['family_history'])
+  # determines if a data criteria has a time period associated with it: it potentially has both
+  # a start and end date.
+  isPeriod: ->
+    criteriaType = @getCriteriaType()
+    # in QDM 5.0, these are all things that are considered 'authored' - they are instances and do not have a time interval.
+    criteriaType not in ['assessment_performed', 'assessment_recommended', 'communication_from_patient_to_provider',
```
jbradl11: Was "not in" used here because the list that it could be in is larger?

edeyoung: i assume so - but I think I was wrong. I could change this to be "in" and redo all of the types. I think I thought not being a period was the exception, but I think it's actually the more common situation. Let me know if you want this changed.

jbradl11: I think if the "in" list is shorter or equal length of "not in", it should be changed. I find it more intuitive to be "in" rather than "not in".

edeyoung:
@jbradl11 I've addressed your comment.

PR #679, opened 2017-02-08
made the authodatetime changes to bonnie. no changes were necessary t…

edeyoung:
…o the model. the authordatetime will still be the model's date time. however, for periods types (e.g. encounter performed), this only shows up if negation is selected. in the qdm docs, it is specified that the intent of authordatetime for period types is to be used when the period type never happened but that needs to be recorded.


edeyoung:
this is duplicated in pull request #680.
this was a pull request into the 5.0 branch which has been incorporated into the master branch. #680 is a pull request into the master branch.

PR #680, opened 2017-02-09
Cql4bonnie qdm 5 2

edeyoung:
* will need to update the Gemfile reference to cql_qdm_patientapi

### DIFF
--- a/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
+++ b/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
@@ -115,10 +115,10 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
       faIcon: @model.faIcon()
       definition_title: definition_title
       canHaveNegation: @model.canHaveNegation()
-      startLabel: @model.startLabel()
-      stopLabel: @model.stopLabel()
-      periodLabel: @model.periodLabel()
-      isPeriod: @model.isPeriod()
+      startLabel: @startLabel(@model.get('negation'))
+      stopLabel: @stopLabel()
+      periodLabel: @periodLabel()
+      isPeriod: @model.isPeriodType() && !@model.get('negation') # if something is negated, it didn't happen so is not a period
 
   # When we serialize the form, we want to convert formatted dates back to times
   events:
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -209,6 +223,31 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
     type = @$(e.target).model().get('type')
     $(".#{type}-elements").focus()
 
+
+  startLabel: (negated) ->
+    if @model.isPeriodType() && !negated
```
jbradl11: I would add a comment here explaining the logic of this if statement. Its not complicated just a little confusing with "negated" being a variable name.

jbradl11:
Should cql4bonnie_QDM_5-2 first pull to the latest cql4bonnie? I am seeing some code in here that was changed in the QDM_5_0  update.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
thecristen: I'd agree.. consider renaming `negated`

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -188,6 +188,20 @@ class Thorax.Views.EditCriteriaView extends Thorax.Views.BuilderChildView
 
   toggleNegationSelect: (e) ->
     @$('.negation-code-list').prop('selectedIndex',0).toggleClass('hide')
+
+    if $(e.target).is(':checked')
+      @$('#periodLabel').addClass('hide')
+      @$('#stopControl').addClass('hide')
+    else
+      @$('#periodLabel').removeClass('hide')
+      @$('#stopControl').removeClass('hide')
```
thecristen: Could be more succinct with more [jquery hackery](https://api.jquery.com/toggleclass/#toggleClass-function-state) (untested example)

```coffeescript
@$('#periodLabel, #stopControl').toggleClass('hide', $(e.target).is(':checked'))
```

But I know I go nuts with that sort of stuff so feel free to leave this as is. 🙃

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
edeyoung: thoughts on a different name? I feel like you talk about something being negated rather than something happening. Happy to change the name, just wondering what to. If name isn't changed, I will definitely add a comment.

c-monkey: A comment wouldn't hurt, but I don't mind negated.

c-monkey (approved):
Tested, looks good to me.

jbradl11 (approved):

PR #684, opened 2017-02-15
BONNIE-455 fixing CQL based measures to support episode of care

c-monkey:
- changed placeholder checkboxes in patient builder to text boxes, for simplicity, as JB says these are going away when we add highlighting
 - put in temporary "This is a CQL measure" description for episodes of care at top of measure view - JB, Kristian, and Chris theorize we might be able to show actual CQL statements here

 ----------------------------------------------
 Need associated changes in bonnie_bundler
 ----------------------------------------------

Tested loading CQL and HQMF measures. It is desired behavior to skip the finalize dialog for CQL-based measures with episodes of care

### DIFF
--- a/app/assets/javascripts/templates/measure.hbs
+++ b/app/assets/javascripts/templates/measure.hbs
@@ -20,18 +20,18 @@
     <div class="measure-dsp">
       <h2 class="full-title">{{title}}</h2>
       {{#if episode_of_care}}
+        {{#if isNotCQL}}
         <div class="measure-dsp-title">
           Episode(s) of Care:
         </div>
-          {{#if cql}}
-              This is a CQL measure
-          {{else}}
-            {{#each episodesOfCare tag="ul"}}
-              {{#with attributes}}
-                <li>{{description}}</li>
-              {{/with}}
-            {{/each}}
-          {{/if}}
+          {{#each episodesOfCare tag="ul"}}
+            {{#with attributes}}
+              <li>{{description}}</li>
+            {{/with}}
+          {{/each}}
+        {{else}}
+            Episode of Care Based Measure
+        {{/if}}
         <p></p>
       {{/if}}
 
--- a/app/assets/javascripts/templates/measure.hbs
+++ b/app/assets/javascripts/templates/measure.hbs
@@ -23,11 +23,15 @@
         <div class="measure-dsp-title">
           Episode(s) of Care:
         </div>
-          {{#each episodesOfCare tag="ul"}}
-            {{#with attributes}}
-              <li>{{description}}</li>
-            {{/with}}
-          {{/each}}
+          {{#if cql}}
+              This is a CQL measure
+          {{else}}
+            {{#each episodesOfCare tag="ul"}}
+              {{#with attributes}}
+                <li>{{description}}</li>
+              {{/with}}
+            {{/each}}
+          {{/if}}
         <p></p>
       {{/if}}
 
--- a/app/assets/javascripts/views/measure_view.js.coffee
+++ b/app/assets/javascripts/views/measure_view.js.coffee
@@ -93,7 +93,7 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
     @measures = @model.collection
 
   episodesOfCare: ->
-    return null if @model.get('is_cql')
+    return null if @model.get('cql')
     @model.get('source_data_criteria').filter((sdc) => sdc.get('source_data_criteria') in @model.get('episode_ids'))
 
   updateMeasure: (e) ->
--- a/app/assets/javascripts/views/measure_view.js.coffee
+++ b/app/assets/javascripts/views/measure_view.js.coffee
@@ -93,6 +93,7 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
     @measures = @model.collection
 
   episodesOfCare: ->
+    return null if @model.get('is_cql')
     @model.get('source_data_criteria').filter((sdc) => sdc.get('source_data_criteria') in @model.get('episode_ids'))
 
   updateMeasure: (e) ->
--- a/app/assets/javascripts/views/measure_view.js.coffee
+++ b/app/assets/javascripts/views/measure_view.js.coffee
@@ -50,13 +50,9 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
       isNotCQL: !@model.has('cql') # Hide certain features in handlebars if the measure is cql.
 
   initialize: ->
+    @measureViz = Bonnie.viz.measureVisualzation().fontSize("1.25em").rowHeight(20).rowPadding({top: 14, right: 6}).dataCriteria(@model.get("data_criteria")).measurePopulation(@population).measureValueSets(@model.valueSets())
     if @model.has('cql')
       populationLogicView = new Thorax.Views.CqlLogic(model: @model)
-    else
-      populationLogicView = new Thorax.Views.PopulationLogic(model: @population)
-    @measureViz = Bonnie.viz.measureVisualzation().fontSize("1.25em").rowHeight(20).rowPadding({top: 14, right: 6}).dataCriteria(@model.get("data_criteria")).measurePopulation(@population).measureValueSets(@model.valueSets())
-
-    if @model.has('cql')
       if @populations.length > 1 # CQL Measure with multiple populations
         @logicView = new Thorax.Views.CqlPopulationsLogic model: @model, collection: @populations
         @logicView.setView populationLogicView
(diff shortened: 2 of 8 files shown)


c-monkey:
JB discovered that the finalize dialog is shown with multiple populations and gave me a CQL based measure with which to test and resolve this issue

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -93,6 +93,7 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
     @measures = @model.collection
 
   episodesOfCare: ->
+    return null if @model.get('cql')
```
edeyoung: add a comment explaining that cql measures don't need a list of defined episodes of care

app/controllers/measures_controller.rb
```diff
@@ -281,11 +267,14 @@ def finalize
     measure_finalize_data = params.values.select {|p| p['hqmf_id']}.uniq
     measure_finalize_data.each do |data|
       measure = Measure.by_user(current_user).where(hqmf_id: data['hqmf_id']).first
-      measure.update_attributes({needs_finalize: false, episode_ids: data['episode_ids']})
+      is_cql = measure == nil
```
edeyoung: can you add comments or rearrange the logic for this? on first pass, it was confusing to access the measure, then say it's cql if the measure doesn't exist (which makes sense, but on first pass it was confusing), and then access the measure again (the if statement is hidden at the end).

maybe:

```
# try to access qdm-based measure
measure = Measure.by_user(current_user).where(hqmf_id: data['hqmf_id']).first
# if measure doesn't exist in qdm measures, it must be a cql measure
is_cql = measure == nil
if is_cql
  measure = CqlMeasure.by_user(current_user).where(hqmf_id: data['hqmf_id']).first if is_cql
end
```

app/controllers/measures_controller.rb
```diff
@@ -281,11 +267,14 @@ def finalize
...
+      measure = CqlMeasure.by_user(current_user).where(hqmf_id: data['hqmf_id']).first if is_cql
+      measure['needs_finalize'] = false
+      measure['episode_ids'] = data['episode_ids'] if !is_cql
```
edeyoung: add comment that cql-based measures don't need a list of episode of care ids

app/controllers/measures_controller.rb
```diff
@@ -281,11 +267,14 @@ def finalize
...
       measure.populations.each_with_index do |population, population_index|
         population['title'] = data['titles']["#{population_index}"] if (data['titles'])
       end
-      measure.generate_js(clear_db_cache: true)
+      measure.generate_js(clear_db_cache: true) if !is_cql
```
edeyoung: add comment why this shouldn't be cleared if it's a cql-based measure.

app/controllers/measures_controller.rb
```diff
@@ -210,7 +194,7 @@ def create
         end
       end
     else
-      measure.needs_finalize = (measure_details['episode_of_care'] || measure.populations.size > 1)
+      measure.needs_finalize = (!is_cql && measure_details['episode_of_care']) || measure.populations.size > 1
```
edeyoung, line 199: for the line 182 section, the only reason I think this works is because it does not appear that cql-based measures can currently be updated. it may be worth adding a TODO in this section that this needs to be updated to not check the episode ids for cql based measures on update once that's implemented or that logic can be added now although it can't be tested.

edeyoung (changes requested):

app/controllers/measures_controller.rb
mulcahyk: Ruby style point: consider `unless is_cql`.

app/controllers/measures_controller.rb
c-monkey: @jbradl11 do you have a preference?

app/assets/javascripts/templates/measure.hbs
```diff
@@ -23,11 +23,15 @@
         <div class="measure-dsp-title">
           Episode(s) of Care:
         </div>
-          {{#each episodesOfCare tag="ul"}}
-            {{#with attributes}}
-              <li>{{description}}</li>
-            {{/with}}
-          {{/each}}
+          {{#if cql}}
```
c-monkey: @jbradl11 JB, are we considering ever supporting identification of episodes of care CQL statements? If not, then I need to annotate the code in multiple places indicating that.

app/controllers/measures_controller.rb
jbradl11: I think adding a TODO here and mentioning the TODO in the JIRA task that relates to implementing CQL measure update should be sufficient.

app/assets/javascripts/templates/measure.hbs
c-monkey: @edeyoung Do you have a better suggestion for how we might have this section in the web page look? Should we omit the whole "Episode(s) of Care:" altogether? The "This is a CQL measure" is clearly placeholder.

app/controllers/measures_controller.rb
c-monkey: @jbradl11 I'd love to do that, great suggestion. I can't find this task. Can you please confirm that it doesn't exist? I don't mind authoring it, if needed. Thanks!

jbradl11: I don't believe one exists yet.

app/assets/javascripts/templates/measure.hbs
edeyoung: Maybe for cql, have it just say:

**Episode of Care Based Measure**

And leave it at that.

The episodes of care will not be available in the logic. However, they can be parsed out from the human readable HTML, so if we wanted to show what the episode of care is, that's how we'd do it. I'd say let's add that as a low priority JIRA task (to parse out the episode of care label) and not list anything for now.

app/controllers/measures_controller.rb
c-monkey: Done

app/controllers/measures_controller.rb
c-monkey: Done

app/controllers/measures_controller.rb
edeyoung: I also added a comment similar to this in this commit: https://github.com/projecttacoma/bonnie/commit/20c49d59421943ca50000504d2d948b9672010ea#diff-9f6c25d2ea5170e4953a469c855a47d4R91

app/assets/javascripts/templates/measure.hbs
c-monkey: Done.

app/assets/javascripts/views/measure_view.js.coffee
c-monkey: @mulcahyk recommend instead to check for the existence of episode_ids; incidentally due to the change in measure.hbs this code is not currently hit for CQL-based measures

app/controllers/measures_controller.rb
c-monkey: Done

app/controllers/measures_controller.rb
c-monkey: Done

app/assets/javascripts/templates/measure.hbs
mulcahyk: Here on line 26 you have `{{#if cql}}` but up on line 7 there is `{{#if isNotCQL}}`.  Can just one of these be used to avoid confusion?

app/controllers/measures_controller.rb
```diff
@@ -266,15 +268,22 @@ def destroy
...
+        measure = CqlMeasure.by_user(current_user).where(hqmf_id: data['hqmf_id']).first
+      end
       measure['needs_finalize'] = false
-      measure['episode_ids'] = data['episode_ids'] if !is_cql
       measure.populations.each_with_index do |population, population_index|
         population['title'] = data['titles']["#{population_index}"] if (data['titles'])
       end
-      measure.generate_js(clear_db_cache: true) if !is_cql
+      # CQL-based measures don't have episode_ids field
+      if !is_cql
```
mulcahyk: Ruby style: use **unless** instead of `if !something`.

app/assets/javascripts/templates/measure.hbs
c-monkey: Done

app/controllers/measures_controller.rb
c-monkey: Done

jbradl11 (approved):
Looks Good

app/controllers/measures_controller.rb
```diff
@@ -223,9 +209,11 @@ def create
 
     end
 
-    Measures::ADEHelper.update_if_ade(measure)
+    if !is_cql
```
edeyoung: here's another place for `unless`

c-monkey: Done

edeyoung (approved):

PR #686, opened 2017-03-03
BONNIE-491 Fix for external BONNIE-224

c-monkey:
- patient dashboard erroneously treated nil data as a 0 causing failure status when stratification actually passed
- see https://jira.mitre.org/browse/BONNIE-491

### DIFF
--- a/app/assets/javascripts/models/patient_dashboard_patient.js.coffee
+++ b/app/assets/javascripts/models/patient_dashboard_patient.js.coffee
@@ -66,10 +66,7 @@ class Thorax.Models.PatientDashboardPatient extends Thorax.Model
     expectedResults = {}
     expected_model = @patient.get('expected_values').findWhere(measure_id: @measure.get('hqmf_set_id'), population_index: @populationSet.get('index'))
     for population in @populations
-      if population not in expected_model.keys()
-        expectedResults[population] = 0
-      else
-        expectedResults[population] = expected_model.get(population)
+      expectedResults[population] = expected_model.get(population)
     expectedResults
 
   ###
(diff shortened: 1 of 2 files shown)


edeyoung (commented):
This seems to address the calculation issue. However, populations not in the population set are still showing up (e.g. there's a STRAT column for the first population that shouldn't be there).

I think this is good to have so that we don't show failing, but it would be nice to show the correct populations as well. I have some thoughts. I'll add them soon.

edeyoung:
Proposed update to only show relevant populations (conceptual only - I haven't even run this):

In app/assets/javascripts/views/patient_dashboard/measure_patient_dashboard.js.coffee line 5:

```coffee
  initialize: ->
    # Grab all populations related to this measure
    codes = (population['code'] for population in @measure.get('measure_logic'))
    @populations = _.intersection(Thorax.Models.Measure.allPopulationCodes, codes)
```

to

```coffee
  initialize: ->
    # Grab all populations related to this measure
    codes = @measure.get('populations').at(@populationSet.get('index')).keys()
    @populations = _.intersection(Thorax.Models.Measure.allPopulationCodes, codes)
```

This should just get the population keys relevant to the given population index.

edeyoung (approved):
I'm approving to get this in ASAP. If the change I described in my comments can be done and tested for a deployment tomorrow, let's do that. Otherwise, I'll take it as is and we can make this mod at a later date.

holmesie:
Lizzie - your change makes sense, and I have tested it. I think it should be included.

edeyoung:
@holmesie can you add it to this pull request? I'll review again and we'll get one other person to review.

Thanks,
Lizzie

holmesie (approved):

jbradl11 (approved):

PR #688, opened 2017-03-08
Addressed code review comments; pulled out CQL and QDM calculators

jbradl11:


holmesie (approved):

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -86,21 +62,23 @@
     cql_map = population.collection.parent.get('populations_cql_map')
     # Grab the correct expected for this population
     index = population.get('index')
-    expected = patient.get('expected_values').models[index]
-    # Loop over all population codes ("IPP", "DENOM", etc.)
+    expected = patient.get('expected_values').findWhere(measure_id: population.collection.parent.get('hqmf_set_id'), population_index: index)
+     # Loop over all population codes ("IPP", "DENOM", etc.)
```
c-monkey: extraneous space

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -86,21 +62,23 @@
...
       if cql_map[popCode]
+        # This code is supporting measures that were uploaded 
+        # before the parser returned multiple populations in an array.
         if _.isString(cql_map[popCode])
           defined_pops = [cql_map[popCode]]
         else
           defined_pops = cql_map[popCode]
-        target_map_index = if defined_pops.length > 1 then index else 0
-        cql_population = defined_pops[target_map_index]
+        index = if defined_pops.length > 1 then index else 0
```
c-monkey: why not `index = 0 unless defined_pops.length > 1`?

c-monkey (approved):
All changes requested in https://github.com/projecttacoma/bonnie/commit/20c49d59421943ca50000504d2d948b9672010ea appear to have been resolved.

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -76,6 +51,7 @@
         result.set {'patient_id': patient['id']} # Add patient_id to result in order to delete patient from population_calculation_view
         result.state = 'complete'
     catch error
+      result.state = 'cancelled'
```
edeyoung, line 54: This is for the code a few lines up. Should this be addressed? Or is it ok to have it be 'undefined' for new patients?

image

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -86,21 +62,23 @@
...
     # Grab the correct expected for this population
     index = population.get('index')
-    expected = patient.get('expected_values').models[index]
-    # Loop over all population codes ("IPP", "DENOM", etc.)
+    expected = patient.get('expected_values').findWhere(measure_id: population.collection.parent.get('hqmf_set_id'), population_index: index)
+     # Loop over all population codes ("IPP", "DENOM", etc.)
     for popCode in Thorax.Models.Measure.allPopulationCodes
       if cql_map[popCode]
+        # This code is supporting measures that were uploaded 
+        # before the parser returned multiple populations in an array.
```
edeyoung: when can we move away from this? Can you add a TODO here? I'd like to have this not be present for when we move the code to production when we'll be starting from scratch. Can we make a list of the measures that this is necessary for and ask the users to delete/re-upload them?

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -86,21 +62,23 @@
...
         if _.isString(cql_map[popCode])
           defined_pops = [cql_map[popCode]]
         else
           defined_pops = cql_map[popCode]
-        target_map_index = if defined_pops.length > 1 then index else 0
-        cql_population = defined_pops[target_map_index]
+        index = if defined_pops.length > 1 then index else 0
+        cql_population = defined_pops[index]
```
edeyoung: I'm not sure this was addressed - i think more logic changes would have been needed. let's discuss to make sure I'm understanding the code properly.

image

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -86,21 +62,23 @@
...
         # Is there a patient result for this population?
         if results['patientResults'][patient.id][cql_population]?
           # Grab CQL result value and adjust for Bonnie
           value = results['patientResults'][patient.id][cql_population]
-          if typeof value is 'object' and value.length > 0
+          if Array.isArray(value) and value.length > 0
```
edeyoung, line 82: Does this work with the episode of care changes?

app/controllers/measures_controller.rb
```diff
@@ -59,10 +59,9 @@ def create
       end
     end
     begin
```
edeyoung: I think we still need to discuss this:

image

edeyoung:
For this comment, please add a TODO.
It is not present in the current diff so can't comment directly.


image

edeyoung (changes requested):

app/controllers/measures_controller.rb
edeyoung: explained - the upload logic is lower down and split into xml and zip loading types. the cql logic will be further integrated over time as we move towards supporting updates.

app/assets/javascripts/cql_calculator.js.coffee
jbradl11: This works as expected.

app/assets/javascripts/cql_calculator.js.coffee
jbradl11: Good catch. Updated.

app/assets/javascripts/cql_calculator.js.coffee
jbradl11: Added a TODO

app/assets/javascripts/cql_calculator.js.coffee
jbradl11: I like it. Makes sense to me.

app/assets/javascripts/cql_calculator.js.coffee
jbradl11: Yes it does. I believe this was your suggestion, and its much clearer compared to the previous version.

app/assets/javascripts/cql_calculator.js.coffee
edeyoung: Tohline, Adam, and I discussed. This is something Tohline will take a look at.

edeyoung (approved):
This set of changes appears to require a hard-reload for the user. When we deploy to staging, we'll need to confirm if users will need to clear their cache for this to work. if so, is there something we can do to make this easier for users?

PR #690, opened 2017-03-14
Fix OBSERV showing failure in patient dashboard

hossenlopp:
made actual results check for observ. have actual results follow the same behavior as expected results here: you make this into a string. If there are multiple values, it will look like '120,60'. Not a perfect solution but it seems to work. Then use the 'actual' that has been recorded on the patient rather than recalculating the patient. this ensures we're using the same formatting.


edeyoung (approved):
Approving on behalf of Hoss. This was my change, but he submitted it for PR review and said he approved it.

jbradl11 (approved):
Looks good.

PR #693, opened 2017-03-23
Updated VSAC profile to be the latest 2017 profile.

hossenlopp:
This has already been hotfixed on production.

### DIFF
--- a/lib/tasks/vsac.rake
+++ b/lib/tasks/vsac.rake
@@ -9,6 +9,7 @@ namespace :bonnie do
       effective_date = Date.parse(ENV['DATE']).strftime('%Y%m%d') if ENV['DATE']
       include_draft = !!ENV['DRAFT']
       config = APP_CONFIG["nlm"]
+      config["profile"] = ENV['PROFILE'] if ENV['PROFILE']
       
       # We are using V2 API:
       api = HealthDataStandards::Util::VSApiV2.new(config["ticket_url"], config["api_url"], ENV['USERNAME'], ENV['PASSWORD'])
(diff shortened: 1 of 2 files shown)


lib/tasks/vsac.rake
```diff
@@ -9,6 +9,7 @@ namespace :bonnie do
       effective_date = Date.parse(ENV['DATE']).strftime('%Y%m%d') if ENV['DATE']
       include_draft = !!ENV['DRAFT']
       config = APP_CONFIG["nlm"]
+      config["profile"] = ENV['PROFILE'] if ENV['PROFILE']
```
edeyoung, line 12: what does this line do?

edeyoung: nm, i get it. it grabs it from the environment variable if its specified.

edeyoung (commented):
So, I checked CMS169 which was one of the measures mentioned. I was looking for the missing codes (RxNorm 1738483). I saw them in my newly uploaded measure on my account. However, when I tried the rake task, this code wasn't showing up.

Thoughts on why the rake task may not be working?

If you don't have time to look into this now, can you delete the mod to the rake task, and I'll accept this profile change?

hossenlopp:
@edeyoung Did you use the DRAFT=true in the rake task. The profile defines the code system versions, not the value set versions.

lib/tasks/vsac.rake
hossenlopp: Yeah, I wanted to be able to change the profile when using this rake task. I was tired of changing the config file and running again.

edeyoung:
I see them now, thanks!

edeyoung (approved):

jbradl11 (approved):

PR #694, opened 2017-03-27
Initial implementation of client side error handling using Costanza

holmesie:
Client side errors are now passed to the backend for inspection. For known errors, the backend crafts a more helpful error message that is passed back to the front end, to be displayed to the user in a flash message. Errors are also now emailed to the development team if seen in a production deployment.

Instructions for adding new types of known frontend errors:
https://github.com/projecttacoma/bonnie/wiki/Adding-custom-error-messages-for-front-end-errors

This prevents the usual "blank screen" issue when JavaScript errors prevent the rest of certain pages from being loaded.

For testing, see: https://oncprojectracking.healthit.gov/support/projects/BONNIE/issues/BONNIE-226

Before:
before

After:
after

### DIFF
--- a/app/assets/javascripts/application.js.coffee
+++ b/app/assets/javascripts/application.js.coffee
@@ -32,18 +32,22 @@
 #= require jquery-color/jquery.color
 #= require bootstrap-file-input/bootstrap.file-input
 #= require MutationObserver-shim/dist/mutationobserver.min
-#= require cql_qdm_patientapi
-#= require ace-builds/src/ace.js
-#= require ace-builds/src/ext-language_tools.js
-#= require ace-builds/src/mode-cql_highlight_rules.js
-#= require ace-builds/src/mode-cql.js
-#= require ace-builds/src/theme-chrome.js
+#
+#= require costanza/js/costanza
+#= require costanza/thorax
 #
 #= require datatables.net/js/jquery.dataTables
 #= require datatables.net-bs/js/dataTables.bootstrap
 #= require datatables.net-fixedcolumns/js/dataTables.fixedColumns
 #= require jquery.scrollTo/jquery.scrollTo
 #= require eonasdan-bootstrap-datetimepicker/build/js/bootstrap-datetimepicker.min
+#
+#= require cql_qdm_patientapi
+#= require ace-builds/src/ace.js
+#= require ace-builds/src/ext-language_tools.js
+#= require ace-builds/src/mode-cql_highlight_rules.js
+#= require ace-builds/src/mode-cql.js
+#= require ace-builds/src/theme-chrome.js
 #
 #= require helpers
 #= require_tree ./templates
@@ -54,7 +58,6 @@
 #= require measure_libraries
 #= require_self
 #= require_tree .
-
 
 # Make all JST templates available as Handlebars templates, so the {{template 'foo'}} helper works as expected
 Handlebars.templates[name] = template for name, template of JST
--- a/app/assets/javascripts/calculator.js.coffee
+++ b/app/assets/javascripts/calculator.js.coffee
@@ -66,10 +66,9 @@
           result.state = 'unstarted'
           return
         result.state = 'complete'
-        try
+        # Capture calculation errors that may arise and handle them using Costanza
+        Costanza.run 'measure-calculation', {cms_id: result.measure.get('cms_id')}, () =>
           result.set @calculator[calcKey](patient.toJSON())
-        catch error
-          bonnie.showError({title: "Measure Calculation Error", summary: "There was an error calculating the measure #{result.measure.get('cms_id')}.", body: "One of the data elements associated with the measure is causing an issue.  Please review the elements associated with the measure to verify that they are all constructed properly.  Error message: #{error.message}."})
 
       setTimeout deferredCalculation, 0
 
--- a/app/assets/javascripts/error_handler.js.coffee
+++ b/app/assets/javascripts/error_handler.js.coffee
@@ -2,7 +2,7 @@
 Costanza.init (info, rawError) ->
   $.ajax
     url: 'application/client_error'
-    method: 'GET'
+    method: 'POST'
     dataType: 'json'
     data: info
     success: (message) ->
--- a/app/assets/javascripts/error_handler.js.coffee
+++ b/app/assets/javascripts/error_handler.js.coffee
@@ -1,11 +1,17 @@
 # Initialize Costanza front-end error handling/tracking
 Costanza.init (info, rawError) ->
-  $.ajax
-    url: 'application/client_error'
-    method: 'POST'
-    dataType: 'json'
-    data: info
-    success: (message) ->
-      bonnie.showError(message) if message
-    error: (info) ->
-      bonnie.showError({'body': info.status + ': ' + info.statusText}) if info
+  # Ignore mirrored Costanza error messages (we have already handled these).
+  # This check stops doubled error message popups from appearing.
+  unless /Costanza/i.test rawError.toString()
+    $.ajax
+      url: 'application/client_error'
+      method: 'POST'
+      dataType: 'json'
+      data: info
+      success: (message) ->
+        bonnie.showError(message) if message
+      error: (jqXHR, textStatus, errorThrown) ->
+        connectionError =
+          title: 'Connection Issues'
+          body: 'An error occurred in Bonnie. You may be experiencing network connectivity issues. Please check your network connection and try reloading Bonnie.'
+        bonnie.showError(connectionError)
--- a/app/assets/javascripts/error_handler.js.coffee
+++ b/app/assets/javascripts/error_handler.js.coffee
@@ -1,0 +1,11 @@
+# Initialize Costanza front-end error handling/tracking
+Costanza.init (info, rawError) ->
+  $.ajax
+    url: 'application/client_error'
+    method: 'GET'
+    dataType: 'json'
+    data: info
+    success: (message) ->
+      bonnie.showError(message) if message
+    error: () ->
+      console.error(info) if info
--- a/app/assets/javascripts/error_handler.js.coffee
+++ b/app/assets/javascripts/error_handler.js.coffee
@@ -7,5 +7,5 @@ Costanza.init (info, rawError) ->
     data: info
     success: (message) ->
       bonnie.showError(message) if message
-    error: () ->
-      console.error(info) if info
+    error: (info) ->
+      bonnie.showError({'body': info.status + ': ' + info.statusText}) if info
--- a/app/assets/javascripts/views/logic/data_criteria_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/data_criteria_logic_view.js.coffee
@@ -14,31 +14,33 @@ class Thorax.Views.DataCriteriaLogic extends Thorax.Views.BonnieView
     # TODO: Find a better way than logic in the template to disable highlighting when called from code other than patient builder.
 
   initialize: ->
-    @dataCriteria = @measure.get('data_criteria')[@reference]
-    # handle reference to source data criteria (this is used for displaying variables)
-    unless @dataCriteria
-      @dataCriteria = @measure.get('source_data_criteria').findWhere({'source_data_criteria': @reference}).attributes
-      @dataCriteria.key = @reference
+    # Capture logic view building errors that may arise and handle them using Costanza
+    Costanza.run 'data-criteria-logic-view-creation', {reference: @reference, cms_id: @measure.get('cms_id')}, () =>
+      @dataCriteria = @measure.get('data_criteria')[@reference]
+      # handle reference to source data criteria (this is used for displaying variables)
+      unless @dataCriteria
+        @dataCriteria = @measure.get('source_data_criteria').findWhere({'source_data_criteria': @reference}).attributes
+        @dataCriteria.key = @reference
 
-    # we need to do this because the view helper doesn't seem to be available in an #each.
-    if @dataCriteria.field_values
-      for key, field of @dataCriteria.field_values
-        # timing fields can have a null value
-        unless field?
-          field = {}
-          @dataCriteria.field_values[key] = field
-        field['key'] = key
-        field['key_title'] = @translate_field(key)
+      # we need to do this because the view helper doesn't seem to be available in an #each.
+      if @dataCriteria.field_values
+        for key, field of @dataCriteria.field_values
+          # timing fields can have a null value
+          unless field?
+            field = {}
+            @dataCriteria.field_values[key] = field
+          field['key'] = key
+          field['key_title'] = @translate_field(key)
 
-    if @dataCriteria.references
-      for key, field of @dataCriteria.references
-        field['key'] = @translate_reference_type(key)
-        field["key_title"] = @translate_reference_to_title(field.referenced_criteria)
-    @dataCriteria.references = null if @dataCriteria && _.isEmpty(@dataCriteria.references)
-    @dataCriteria.field_values = null if @dataCriteria && _.isEmpty(@dataCriteria.field_values)
-    @isSatisfies = @dataCriteria.definition in @satisfiesDefinitions
-    @isDerived = @dataCriteria.type == 'derived'
-    @hasChildren = @isDerived && (!@dataCriteria.variable || @expandVariable)
+      if @dataCriteria.references
+        for key, field of @dataCriteria.references
+          field['key'] = @translate_reference_type(key)
+          field["key_title"] = @translate_reference_to_title(field.referenced_criteria)
+      @dataCriteria.references = null if @dataCriteria && _.isEmpty(@dataCriteria.references)
+      @dataCriteria.field_values = null if @dataCriteria && _.isEmpty(@dataCriteria.field_values)
+      @isSatisfies = @dataCriteria.definition in @satisfiesDefinitions
+      @isDerived = @dataCriteria.type == 'derived'
+      @hasChildren = @isDerived && (!@dataCriteria.variable || @expandVariable)
 
   isSetOp: => @dataCriteria.derivation_operator in _(@set_operator_map).keys()
 
--- a/app/helpers/error_helper.rb
+++ b/app/helpers/error_helper.rb
@@ -1,0 +1,45 @@
+# Helper module for handling client side errors.
+module ErrorHelper
+  # Handles the given error by categorizing it based on the included
+  # section, as defined in the front end code. If the given error can
+  # not be categorized, no error message will be shown (but an email
+  # will still be sent to the development team, if on production).
+  def self.describe_error(error_info, exception, request)
+    # Do not process errors if their message includes Costanza. These are
+    # errors passed up to Thorax.onException from Costanza, which we do
+    # not care about (we've already handled them).
+    return if error_info[:msg].include? 'Costanza'
+
+    # If enabled, send an email to the development team containing
+    # information about this error.
+    ErrorHelper.send_email(error_info, exception, request) if APP_CONFIG['enable_client_error_email']
+
+    case error_info[:section]
+    # Handles errors that arise when generating the human readable logic from
+    # the measure data criteria.
+    when /data-criteria-logic-view-creation/
+      {
+        title: 'Measure Logic Error',
+        summary: 'Error Building Measure Logic View',
+        body: 'You have imported a measure that contains logic that cannot be rendered. This is usually caused by errors in the measure logic. Please review the measure logic for correctness, and re-upload the measure.<br>Data criteria reference that failed: <b>' + error_info[:reference] + '</b>'
+      }
+    # Handles errors that arise when trying to calculate a measure against
+    # patients.
+    when /measure-calculation/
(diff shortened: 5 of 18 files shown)


app/assets/javascripts/calculator.js.coffee
```diff
@@ -66,10 +66,8 @@
           result.state = 'unstarted'
           return
         result.state = 'complete'
-        try
+        Costanza.run 'measure-calculation', {cms_id: result.measure.get('cms_id')}, () =>
```
jbradl11: I would recommend a comment here (and in the data_criteria_logic_view) explaining what is happening on this line.

app/assets/javascripts/views/logic/data_criteria_logic_view.js.coffee
```diff
@@ -14,31 +14,32 @@ class Thorax.Views.DataCriteriaLogic extends Thorax.Views.BonnieView
     # TODO: Find a better way than logic in the template to disable highlighting when called from code other than patient builder.
 
   initialize: ->
-    @dataCriteria = @measure.get('data_criteria')[@reference]
-    # handle reference to source data criteria (this is used for displaying variables)
-    unless @dataCriteria
-      @dataCriteria = @measure.get('source_data_criteria').findWhere({'source_data_criteria': @reference}).attributes
-      @dataCriteria.key = @reference
+    Costanza.run 'data-criteria-logic-view-creation', {reference: @reference}, () =>
```
jbradl11: Recommend adding a comment. Same as above.

app/assets/javascripts/application.js.coffee
```diff
@@ -32,19 +32,23 @@
 #= require jquery-color/jquery.color
 #= require bootstrap-file-input/bootstrap.file-input
 #= require MutationObserver-shim/dist/mutationobserver.min
-#= require cql_qdm_patientapi
```
c-monkey, line 35: Why was this block moved? It makes the diff less clear. The only actual change to this file is the addition of costanza. Or, was there a need to move this block due to a compiler error?

app/assets/javascripts/error_handler.js.coffee
```diff
@@ -0,0 +1,9 @@
+# Initialize Costanza front-end error handling/tracking
+Costanza.init (info, rawError) ->
+  $.ajax
+    url: 'application/client_error'
+    method: 'GET'
+    dataType: 'json'
+    data: info
+    success: (message) ->
```
hossenlopp: Perhaps the error handler should be implemented here too. If the server fails for some reason responding to the client_error the user will not see an error message at all.

config/routes.rb
```diff
@@ -17,6 +17,10 @@
   get '404', :to => 'application#page_not_found'
   get '500', :to => 'application#server_error'
 
+  # The following route is for the reporting of front end (javascript) errors
+  # to the backend.
+  match 'application/client_error' => 'application#client_error', :via => :get
```
hossenlopp: Why not use the following line?
```
get '/application/client_error' => 'application#client_error'
```

Why is this implemented as a GET? Wouldn't a POST make more sense for this sort of thing?

app/helpers/error_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+      {
+        title: 'Measure Calculation Error',
+        summary: 'There was an error calculating measure ' + error_info[:cms_id] + '.',
+        body: 'One of the data elements associated with the measure is causing an issue. Please review the elements associated with the measure to verify that they are all constructed properly.<br>Error message: <b>' + error_info[:msg] + '</b>'
+      }
+    end
+  end
+
+  # Emails an error description notification (if on production).
+  def self.send_email(error_info)
```
hossenlopp: Have you gotten to test the email sending to see if the mail received has enough info to denote it is a front end error and where it happened?

app/helpers/error_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+    return if error_info[:msg].include? 'Costanza'
+
+    # If enabled, send an email to the development team containing
+    # information about this error.
+    ErrorHelper.send_email(error_info) if APP_CONFIG[:enable_client_error_email]
+
+    case error_info[:section]
+    # Handles errors that arise when generating the human readable logic from
+    # the measure data criteria.
+    when /data-criteria-logic-view-creation/
```
c-monkey, line 20: Managing a lot of errors would get messy here, but assuming we refactor the implementation if it becomes burdensome, this is fine for now.

c-monkey (commented):
I didn't run it, but seems logical. Your screen capture seems superior.

hossenlopp (changes requested):

app/assets/javascripts/error_handler.js.coffee
holmesie: Agreed, that makes sense

config/routes.rb
holmesie: So the communication workflow goes: frontend => backend => frontend; meaning the frontend is essentially 'requesting' more information about an error, which made me think a GET makes the most sense, although I'm open to debate about it.

app/helpers/error_helper.rb
holmesie: For this I'd have to defer to what exactly a Costanza error looks like (see my link to that repo for more info), which is already way better than the usual `window.onerror`. That in addition to the Rails `current_user` being included in this email gives us enough info to look into potential issues (in my opinion).

app/assets/javascripts/application.js.coffee
holmesie: Yeah, this was a tricky change. The ace editor seems to seriously mess with our `window` namespaces; specifically, loading Costanza (which uses things like `module`) after ace.js causes serious issues (ace.js is used in the learn CQL tool). Loading Costanza before ace.js fixes those errors completely. I think this might justify more investigation into what ace.js is actually doing - but since we are planning on removing ace.js eventually in favor of generating our own human readable, perhaps that isn't worth the effort.

app/helpers/error_helper.rb
holmesie: Agreed - perhaps a more "human readable" YAML configuration file containing this information might be more maintainable. I am not opposed to doing that work in this pull request (not too difficult). Thoughts?

config/routes.rb
holmesie: As to my using `match` here - just a habit with the Rails work I've done. What is the best practice in this case?

app/helpers/error_helper.rb
c-monkey: I'm not a fan of over-engineering a solution before we have a problem. It is desirable to validate correspondence between the string in the error handler and the one in the `Costanza.run` match. Perhaps defining the set of errors in code that is shared via an .erb coffeescript module? That sounds trivial and can't get out of sync.

config/routes.rb
c-monkey: My only comment, w/out knowing Ruby best practices, is that consistency is golden. All the rest of the string matches are done with the `get` statement, so this wouldn't look different. There is no reason not to be consistent in this case.

app/assets/javascripts/application.js.coffee
c-monkey: Thanks for the explanation. I figured it wasn't arbitrary on your part, but it is good to confirm.

app/assets/javascripts/calculator.js.coffee
c-monkey: Or, instead of a comment, a call to a helper method, as I ruminated on in my comment in `app/helpers/error_helper.rb` - the method would be documented once, rather than having to comment everywhere in the code.

app/assets/javascripts/error_handler.js.coffee
c-monkey: That's a really good idea. And trivial to test, just shut down your server.

app/helpers/error_helper.rb
c-monkey: LOL, okay, now I can't consent to this pull request until I've tested seeing an email. I thought I was just poking at this code review, but it seems like a reasonable request that all its functionality has be executed.

c-monkey (changes requested):

config/routes.rb
hossenlopp: I think `match` is used for more complex routing situations.

In regards to using a POST: I see it as the frontend is POSTing an error to the server for it's logging purpose. It happens to maybe get useful stuff in the response. Using a GET request means all that data goes in query parameters, which means character encoding fun. Some browsers also chop off long query parameters. Additionally, if the same error happens the GET request could get cached and we won't get a report of that error happening.

holmesie: @hossenlopp Agreed, I think that reasoning makes sense and is enough to merit a change

app/helpers/error_helper.rb
holmesie: Not so sure that would be as trivial as you think; mixing ERB with a JavaScript framework like Thorax and also pulling from that in the back end seems like it would introduce more complication than is necessary, while not really giving much benefit. I think your first comment seems to make the most sense on second thought.

app/helpers/error_helper.rb
edeyoung: There's an outstanding PR that hasn't been incorporated that sends emails because we haven't been able to test that. I believe someone mentioned that we can send up bonnie-review to send emails. Since this also has email settings, can we do that? I think this is a really important PR to incorporate and it would be nice to get that other one incorporated too.

edeyoung (commented):
I'll defer to Tohline and Hoss's acceptance with an emphasis on Tohline's requirement to test that email works.

Looks good and will be a really great modification for Bonnie. Thanks!

c-monkey:
I loaded the issue described in https://oncprojectracking.healthit.gov/support/browse/BONNIE-226
(https://oncprojectracking.healthit.gov/support/secure/attachment/24379/OCM7_PrCA_AdjuvantHormonalTx_v4_3_Artifacts%20%286%29.zip)

I saw this work in some context when I was looking at it with you Adam, but don't remember where...anyway, it isn't working now. Getting this error in my local console, and no user error in the browser, although I do get a console error [Failed to load resource: the server responded with a status of 400 (Bad Request)]

Invalid request: HTTP element QUERY_STRING is longer than the (1024 * 10) allowed length.
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/request.rb:84:in `execute'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/request.rb:84:in `parse'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/connection.rb:39:in `receive_data'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:187:in `run_machine'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:187:in `run'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/backends/base.rb:73:in `start'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/server.rb:162:in `start'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/rack-1.5.5/lib/rack/handler/thin.rb:16:in `run'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/rack-1.5.5/lib/rack/server.rb:264:in `start'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/server.rb:69:in `start'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/commands_tasks.rb:81:in `block in server'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/commands_tasks.rb:76:in `tap'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/commands_tasks.rb:76:in `server'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/commands_tasks.rb:40:in `run_command!'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands.rb:17:in `<top (required)>'
	/Users/ctohline/Projects/Bonnie/bonnie/bin/rails:9:in `require'
	/Users/ctohline/Projects/Bonnie/bonnie/bin/rails:9:in `<top (required)>'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/ruby-debug-ide-0.6.1.beta4/lib/ruby-debug-ide.rb:91:in `debug_load'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/ruby-debug-ide-0.6.1.beta4/lib/ruby-debug-ide.rb:91:in `debug_program'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/ruby-debug-ide-0.6.1.beta4/bin/rdebug-ide:156:in `<top (required)>'
	-e:1:in `load'
	-e:1:in `<main>'

Invalid request: HTTP element QUERY_STRING is longer than the (1024 * 10) allowed length.
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/request.rb:84:in `execute'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/request.rb:84:in `parse'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/connection.rb:39:in `receive_data'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:187:in `run_machine'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/eventmachine-1.0.7/lib/eventmachine.rb:187:in `run'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/backends/base.rb:73:in `start'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/thin-1.6.3/lib/thin/server.rb:162:in `start'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/rack-1.5.5/lib/rack/handler/thin.rb:16:in `run'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/rack-1.5.5/lib/rack/server.rb:264:in `start'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/server.rb:69:in `start'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/commands_tasks.rb:81:in `block in server'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/commands_tasks.rb:76:in `tap'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/commands_tasks.rb:76:in `server'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands/commands_tasks.rb:40:in `run_command!'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/railties-4.1.12/lib/rails/commands.rb:17:in `<top (required)>'
	/Users/ctohline/Projects/Bonnie/bonnie/bin/rails:9:in `require'
	/Users/ctohline/Projects/Bonnie/bonnie/bin/rails:9:in `<top (required)>'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/ruby-debug-ide-0.6.1.beta4/lib/ruby-debug-ide.rb:91:in `debug_load'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/ruby-debug-ide-0.6.1.beta4/lib/ruby-debug-ide.rb:91:in `debug_program'
	/Users/ctohline/.rvm/gems/ruby-2.3.0/gems/ruby-debug-ide-0.6.1.beta4/bin/rdebug-ide:156:in `<top (required)>'
	-e:1:in `load'
	-e:1:in `<main>'

holmesie:
I think I understand what is happening, I will get back to you on Monday.

c-monkey (approved):

jbradl11 (approved):

app/assets/javascripts/error_handler.js.coffee
```diff
@@ -0,0 +1,11 @@
...
+Costanza.init (info, rawError) ->
+  $.ajax
+    url: 'application/client_error'
+    method: 'POST'
+    dataType: 'json'
+    data: info
+    success: (message) ->
+      bonnie.showError(message) if message
+    error: () ->
+      console.error(info) if info
```
hossenlopp: This completely eats the error as far as the user is concerned. Perhaps the bonnie.showError() should be called to let them know there was an error but an issue with Bonnie or their connection prevented from more details being available.

hossenlopp (changes requested):

app/assets/javascripts/error_handler.js.coffee
```diff
@@ -7,5 +7,5 @@ Costanza.init (info, rawError) ->
     data: info
     success: (message) ->
       bonnie.showError(message) if message
-    error: () ->
-      console.error(info) if info
+    error: (info) ->
+      bonnie.showError({'body': info.status + ': ' + info.statusText}) if info
```
hossenlopp: This results in very non-descript and ugly error. That info object is just jQXHR object. I ran a test where I loaded a broken measure after shutting the server down. (This could be similar to Bonnie going down or the user's network connection dying). It got double popups and didn't have much info. 
uglyerror

hossenlopp (changes requested):
Perhaps needs a more friendly message for errors from POSTing to client_error.

app/assets/javascripts/error_handler.js.coffee
c-monkey: Thanks for testing this, we discussed it but I don't recall doing so.

holmesie: Opinions for what the error message should say and consist of? I definitely do not want to be parsing 404s on the client side to make them look prettier, as in this case we are dealing with network issues, etc - and not calculation/measure logic issues.

c-monkey: Random thoughts:
- Hoss, is your animated GIF showing the "ugly error" you refer to?
- Adam, can we fix the double-popup? When I load in some measure developers' dashboards there are tons of popups -- might we want to resolve this in the same fashion? If the 2 popups are identical, can we tell that and cull them? If that's the solution perhaps we should open another ticket.
- When Bonnie is getting an error due to a network outage, we don't really care, right? The existing text seems sufficient to me. Do we want power users to know the exact error? Assuming this is happening because we can't report the error to the mailing list, what is best for our needs?

holmesie: Hmm, not so sure we can (or at least easily can). Errors are sent to the backend and dealt with on a case by case basis. If we tried to bundle similar things together, we might start approaching something that gets complicated quickly (e.g. do we wait X seconds an see if the same error comes in, and then tell the front end to only show things once?) - I'd really prefer to keep it simple in this case (even though the the same error showing multiple times isn't ideal).

To your third point, I totally agree - the error was caused by extenuating circumstances in this edge case, and isn't even remotely as important as the primary types of errors this pull request is trying to identify and describe (bad measure logic?, etc.). If a user did see this, and contacted us via mailing list/JIRA, we would instantly know either the server is down, or they had (seriously) intermittent network connectivity issues.

hossenlopp: So the popups I captured were brought up in the error handler of the AJAX POST to /application/client_error. I turned off the server before clicking on the measure that breaks so the AJAX call would error out. I was surprised that the error was 404, since that's not a server response, the browser decided on 404 because the connection was rejected. I do wonder if this is consistent between browsers.

I think that if it is unable to succeed in the client_error call we should assume that the user's network connection dropped or Bonnie is down. The resulting error should say that.

hossenlopp: What do people think of this? 
connectionerror

holmesie: I'm thinking it should be more generic, what if it happens to be something other than 404? So rather than saying "Additionally, it appears...", we could say, "You may be experiencing network connectivity issues. Please check your network connection and reload." (or something like that).

Thoughts?

c-monkey: Adam's solution we can implement in 5 minutes.

Hoss's solution we can implement in 60 minutes (unless we can't see the HTTP error code).

I think there's value in having different messages based on error code, but I guess I don't have enough information to be completely sure. What is the result on IE? Safari? Chrome? Firefox?

hossenlopp (approved):

c-monkey (approved):
tested on my machine using BonedensitywithPCaandADT_v4_6_eMeasure (1).xml

PR #696, opened 2017-03-31
Bug fix of CQL support for stratifications

c-monkey:
See https://jira.mitre.org/browse/BONNIE-423

Note: must update this pull request with Gemfile change to point to https://github.com/projectcypress/health-data-standards/pull/484 after that change is merged.

### DIFF
--- a/app/assets/javascripts/cql_calculator.js.coffee
+++ b/app/assets/javascripts/cql_calculator.js.coffee
@@ -94,12 +94,13 @@
           defined_pops = [cql_map[popCode]]
         else
           defined_pops = cql_map[popCode]
-        # We have to special case the stratification because the stratification results
-        # array doesn't include the population criteria
+        # We have to special-case the stratification population code ("STRAT")
         if popCode == 'STRAT'
+          # Ignore the STRAT popCode for "Population Criteria Section"
           if index == 0
-            # Ignore the STRAT popCode for "Population Criteria Section"
+            # do not add "STRAT" to population_results
             continue
+          # stratification results array starts with 1st stratification 0-indexed
           index--
         target_map_index = if defined_pops.length > 1 then index else 0
         cql_population = defined_pops[target_map_index]
--- a/app/assets/javascripts/cql_calculator.js.coffee
+++ b/app/assets/javascripts/cql_calculator.js.coffee
@@ -94,13 +94,23 @@
           defined_pops = [cql_map[popCode]]
         else
           defined_pops = cql_map[popCode]
-        # We have to special-case the stratification population code ("STRAT")
+        # Special-case the stratification population code to not include a STRAT in the
+        # population_results for the overall "Population Criteria Section" and then to
+        # read the 0-indexed population_results (e.g. Stratification1 is the 0th item
+        # in the cql_population array)
         if popCode == 'STRAT'
-          # Ignore the STRAT popCode for "Population Criteria Section"
-          if index == 0
-            # do not add "STRAT" to population_results
-            continue
-          # stratification results array starts with 1st stratification 0-indexed
+          # do not add 'STRAT' popCode to population_results for "Population Criteria Section"
+          # this value of index is for the overall population results, not a specific
+          # stratification, therefore this item 'STRAT' shouldn't be included
+          continue if index == 0
+          # Bonnie's indexing of stratifications is considered to be a separate population,
+          # so index 1 corresponds to Stratification 1, however the stratification results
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -94,6 +94,13 @@
           defined_pops = [cql_map[popCode]]
         else
           defined_pops = cql_map[popCode]
+        # We have to special case the stratification because the stratification results
+        # array doesn't include the population criteria
+        if popCode == 'STRAT'
+          if index == 0
+            # Ignore the STRAT popCode for "Population Criteria Section"
+            continue
+          index--
```
jbradl11: I would suggest a more detailed description as to what the special case is doing. For example, why are you calling "index--"?

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -94,6 +94,14 @@
           defined_pops = [cql_map[popCode]]
         else
           defined_pops = cql_map[popCode]
+        # We have to special-case the stratification population code ("STRAT")
+        if popCode == 'STRAT'
+          # Ignore the STRAT popCode for "Population Criteria Section"
```
jbradl11: Is "Population Criteria Section" a user defined name? If so maybe we should change the description to match that.

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -94,6 +94,14 @@
...
+          if index == 0
+            # do not add "STRAT" to population_results
+            continue
```
jbradl11: I would recommend rewriting this to "continue if index == 0"

app/assets/javascripts/cql_calculator.js.coffee
edeyoung: what is the "population criteria section" and why does `index == 0` indicate the population criteria section?

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -94,6 +94,14 @@
...
+          # stratification results array starts with 1st stratification 0-indexed
```
edeyoung: I'm confused about this.

why are we reducing the index? Do we assume all indices are 1 digit too high? But there's a 0th indexed one it looks like, we just don't add it? When/why does the 0th indexed one exists?

I think it would be worth adding a longer comment at the top of the `STRAT` section that explains the entire situation. The reason behind the special casing is not clear to me.

edeyoung (changes requested):
I think this needs clearer, longer comments. It's not clear to me why the given code had to be added.

app/assets/javascripts/cql_calculator.js.coffee
c-monkey: I guess it's the default in the MAT. Bonnie pulls the name from the HQMF section.

app/assets/javascripts/cql_calculator.js.coffee
c-monkey: Fine by me

app/assets/javascripts/cql_calculator.js.coffee
c-monkey: It's the name of the overall section #2 referenced in https://ecqi.healthit.gov/hqmf describing the High-level Structure.

I've added documentation to the code.

c-monkey:
Let's see how these comments read. Thanks

jbradl11 (approved):
This is much better. The new comments give me a better idea of what is going on.

edeyoung (approved):

edeyoung (changes requested):
This is good, but I never ran it - i got caught up in, "my comments were addressed". Waiting for measure from Tohline. Once that works, I'll approve and we can incorporate.

edeyoung (approved):

c-monkey:
Sadly, this code doesn't work for CMS137!

Problem is we have 2 populations.

Our extraction code in HDS really seems to need to have an array of stratifications per population @holmesie @jbradl11 advice?

edeyoung (changes requested):
the underlying hds commit did not properly accommodate non-strat functions. It's possible that this bonnie UI would actually do that. These need to be retested together with both a strat and a non-strat measure. the non-strat measure needs to be confirmed to NOT HAVE "STRAT" in the evaluation portion above the CQL view.

c-monkey:
Matt is reinventing the wheel on this puppy, closing this deprecated PR now.

PR #698, opened 2017-04-07
Add ELM JSON function generation for CV support in CQL

mayerm94:


mulcahyk:
Before this can be merged we need to merge in the changes for HDS.  (Fix support for CV Measures #487 in HDS)

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -20,6 +20,25 @@
...
+          "type": "FunctionRef",
+          "operand": [
+            {
+              "name": parameter,
+              "type": "ExpressionRef"
+            }
+          ]
+        }
+      }
+    elmFunction
```
holmesie: If you're going to go this route to build the ELM statement (rather than using Handlebars for something like a `elm_function.json.hb`), I'd suggest making this object in CoffeeScript form. So along the lines of (not tested):
```
elmFunction =
  name: 'BonnieFunction'
  context: 'Patient'
  accessLevel: 'Public'
  expression =
    name: functionName
    type: 'FunctionRef'
    operand =
      parameter: parameter
      type: 'ExpressionRef'
```

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -20,6 +20,25 @@
   cancelCalculations: ->
     result.state = 'cancelled' for key, result of @resultsCache when result.state == 'pending'
 
+  # Returns a JSON function to add to the ELM before ELM JSON is used to calculate results
+  generateELMJSONFunction: (functionName, parameter) ->
+    elmFunction = {
+        "name": "BonnieFunction",
```
mulcahyk: Is it worth appending the name of the function that we are adding here?  What if in the future we need to add another function like this; will it be confusing to have two names of "BonnieFunction"?

jbradl11: I agree with Kristian, there may be unintended consequences if we have more than one define statement with the same name.

edeyoung (commented):
Looks good to me other than comments mentioned (I don't have an opinion either way on Adam's comment about coffeescript form). I'll defer to Adam, Kristian, and James' review.

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -22,22 +22,16 @@
...
-      }
-    elmFunction
+    elmFunction = 
+        name: 'BonnieFunction' + functionName,
+        context: 'Patient',
+        accessLevel: 'Public',
+        expression:
+          name: functionName,
+          type: 'FunctionRef',
+          operand:
```
edeyoung, line 32: `operand` is an array in the other version. Does this properly reflect an array?

edeyoung: or does it no longer take an array?

jbradl11: I believe this is what Matt was talking about yesterday at standup. We agreed that the function should only support taking one parameter and not an array. Matt, correct me if I am mistaken.

edeyoung: got it, thanks!

jbradl11: I will leave this as an optional change. But maybe naming the function with a "_" between the BonnieFunction and the functionName. Just for easier reading when debugging.
`
name: 'BonnieFunction_' + functionName,`

jbradl11 (approved):

mayerm94:
[CMS55_v5_1_Artifacts 10.zip](https://github.com/projecttacoma/bonnie/files/930622/CMS55_v5_1_Artifacts.10.zip)

This is a measure to test CV with. Should see Quantity result of 500 minutes in patient results

mayerm94:
CQL Function needs to be adjusted to iterate over arguments

PR #702, opened 2017-04-18
Fix bug that allows a CQL measure to be loaded twice

mayerm94:


c-monkey (approved):
tested extensively; I wish I knew what happened, but it all appears to be good now

app/controllers/measures_controller.rb
```diff
@@ -67,7 +67,7 @@ def create
         measure = Measures::MATLoader.load(params[:measure_file], current_user, measure_details, params[:vsac_username], params[:vsac_password]) # overwrite_valuesets=true, cache=false, includeDraft=true
 
         existing = CqlMeasure.by_user(current_user).where(hqmf_set_id: measure.hqmf_set_id)
-        if existing.count > 1
+        if existing.count > 0
```
edeyoung: Please add a comment here that notes that this `existing.count` is intentionally different than the qdm one (1 vs 0). Please add a `TODO` noting that this should be more smoothly managed when the cql code is better incorporated in a future PR.

edeyoung (changes requested):

edeyoung (approved):

PR #704, opened 2017-04-19
ELMJSON Function Generation For CV Measures

mayerm94:


edeyoung:
can you add the measure to test this with, what to build on that measure to test this, and how to check the results?
I'd like to check this with two "observations" on the patient.

mayerm94:
[CMS55_v5_1_Artifacts 10.zip](https://github.com/projecttacoma/bonnie/files/938854/CMS55_v5_1_Artifacts.10.zip)
screen shot 2017-04-19 at 6 50 56 pm

You can test CV Functions by uploading the attached measure and creating a patient with the 2 datacriteria seen in the screenshot. In order to have multiple observations on the patient, duplicate the 2 datacriteria seen and give them different dates.

holmesie:
Looks good to me; my only comment is that I'd like to see a fixup of the last 8 commits into the first one to clean things up - 9 commits is a bit much for this small of a change.

c-monkey:
@holmesie Does Matt have to do the squash, or will the "squash and merge" button on github suffice for your needs, Adam?

mulcahyk:
I would like to see a comment somewhere explaining the formatting of what we are adding or at least a reference to where we got it from.  This went through several evolutions and in a year we aren't going to remember why we did this.

holmesie:
@c-monkey I'm not actually sure about that; I prefer to use the command line because I'm much more comfortable with it - but if GitHub actually just does it for you if you click that button I'm fine with that.

c-monkey (approved):
tested locally

PR #710, opened 2017-04-26
Fix issue with Patient Bank not loading (cql4bonnie branch)

mulcahyk:
When a user goes to use the patient bank it appears to them that Bonnie has stopped responding.  The cause is a mongoid error.  This fixes that error.

Replace find method with keep_if

### DIFF
--- a/app/controllers/patients_controller.rb
+++ b/app/controllers/patients_controller.rb
@@ -11,7 +11,7 @@ class PatientsController < ApplicationController
     # record for cms_id and user_email; prepopulate these values using lookup tables, with lookups
     # for email by user_id and cms_id by user_id and measure_id
     user_ids = records.map(&:user_id).uniq
-    users = User.only(:_id, :email).to_a.keep_if { |u| user_ids.include?(u.id) }
+    users = User.only(:_id, :email).where(:_id.in => user_ids)
     email_lookup = users.each_with_object({}) { |u, h| h[u.id] = u.email }
     measure_ids = records.map { |r| r.measure_ids.first }.uniq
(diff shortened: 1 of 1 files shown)


c-monkey (approved):
tested locally; fixes the bug

app/controllers/patients_controller.rb
```diff
@@ -11,7 +11,7 @@ def index
     # record for cms_id and user_email; prepopulate these values using lookup tables, with lookups
     # for email by user_id and cms_id by user_id and measure_id
     user_ids = records.map(&:user_id).uniq
-    users = User.only(:_id, :email).find(user_ids)
+    users = User.only(:_id, :email).to_a.keep_if { |u| user_ids.include?(u.id) }
```
holmesie: Same as my comment on your master pull request:

I'm worried the `to_a` is making this less efficient than it needs to be.

Could you use `where(id: user_ids)`? `find` raises an exception if it couldn't find the object by the given id, but `where` will not.

That being said, this is MongoDB, so my concern might not even be worth it.

holmesie (changes requested):

app/controllers/patients_controller.rb
c-monkey: +1 I agree with Adam.
Look at line 17 for almost exactly what you want, of course that's on Measure rather than User. Similar feedback was just added by @hossenlopp for the master branch PR.

c-monkey (changes requested):

app/controllers/patients_controller.rb
mulcahyk: Done.

c-monkey (approved):
tested locally; fixes the bug

holmesie (approved):

PR #712, opened 2017-04-27
Fix issue with Patient Bank not loading (master branch)

mulcahyk:
Replace mongoid find with mongoid where.


c-monkey (approved):
tested locally; fixes the bug

holmesie (approved):

PR #715, opened 2017-05-03
Cql4bonnie bonnie 595

mayerm94:
See QDM5.02 3.1.3 Assessment Performed Attributes table https://ecqi.healthit.gov/system/files/qdm_5_02_draft_2016nov30_authortimeerrata_508_compliant_v2.pdf

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/edit_value.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_value.hbs
@@ -53,9 +53,7 @@
       <select id="type_{{@cid}}" name="type" class="selectBoxIt-native">
           <option value="PQ" data-icon="fa fa-pencil-square-o">Scalar</option>
           <option value="CD" data-icon="fa fa-list">Coded</option>
-          {{#if fieldValue}}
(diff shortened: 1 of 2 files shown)


mayerm94:
Can test with the cql_qdm_patientapi cql4bonnie_595 branch. Using 
[TESTCMS9_GA_v5_1_Artifacts (1).zip](https://github.com/projecttacoma/bonnie/files/973466/TESTCMS9_GA_v5_1_Artifacts.1.zip)
To test. Notice, this does not fix the error that will pop up saying that durationBetween cannot be called on undefined. That error will persist until there is a data criteria that has the dateTime result that duration between can be called on. Create a patient with an Assessment Performed Data Criteria (TEST estimated due date) and add a result date to the criteria

c-monkey (approved):
Tested locally.

Of note, the result methods in cql_qdm_patientapi all can only handle a single value. If multiple results are added by users, we might have confusion on their part. This isn't introduced by this pull request, of course.

holmesie (approved):

PR #719, opened 2017-05-09
BONNIE-571

mulcahyk:
set suppressDataCriteriaHighlight for measure_view
Setting to true (so that highlighting is suppressed).
The only other place in the code that this variable was being set was in the patient builder (where it was set to `false`).


jbradl11 (approved):

c-monkey (approved):
tested locally

hossenlopp (approved):

PR #722, opened 2017-05-10
Add handling of MeasureLoadingException

mayerm94:


mayerm94:
@c-monkey decided we shouldn't be displaying the line numbers of where the translator is crashing to the user. No longer need the front end code that supports this.

c-monkey:
rock on!

PR #723, opened 2017-05-11
CQL JSON ELM based logic views with highlighting

hossenlopp:
- Replaces use of ACE Editor in the logic view.
- View for each define statement that gets highlighted from execution engine results.
- Used in patient builder as well with patient data highlighting for defines that result in a list of entries.

### DIFF
--- a/app/assets/javascripts/templates/logic/cql_logic.hbs
+++ b/app/assets/javascripts/templates/logic/cql_logic.hbs
@@ -1,3 +1,5 @@
-<div class="measure-viz rationale panel-group">
-  <div class="cql-editor" id="editor"></div>
+<div class="cql-logic measure-viz rationale panel-group">
+  {{#each statementViews}}
+    {{view .}}
+  {{/each}}
 </div>
--- a/app/assets/javascripts/templates/logic/cql_logic.hbs
+++ b/app/assets/javascripts/templates/logic/cql_logic.hbs
@@ -1,3 +1,4 @@
+{{! TODO: Remove this outdated message once CQL settles for production. Currently needed because the volatile annotation structure. }}
 {{#if isOutdatedUpload}}
 <p>
   <div class="alert alert-warning">
--- a/app/assets/javascripts/templates/logic/cql_logic.hbs
+++ b/app/assets/javascripts/templates/logic/cql_logic.hbs
@@ -1,3 +1,10 @@
+{{#if isOutdatedUpload}}
+<p>
+  <div class="alert alert-warning">
+    <strong>Outdated Measure!</strong> This measure appears to be uploaded with an earlier version of Bonnie Alpha. This view may not function properly until the measure is reuploaded.
+  </div>
+</p>
+{{/if}}
 <div class="cql-logic measure-viz rationale panel-group">
   {{#each statementViews}}
     {{view .}}
--- a/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
@@ -48,9 +48,12 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
   # Expects model to be a Measure model object of a CQL measure.
   ###
   initialize: ->
+    @isOutdatedUpload = false
     @statementViews = []
     _.each @model.get('elm').library.statements.def, (statement) =>
       if (statement.annotation)
+        if (statement.annotation[0].s.value[0] == "define ")
+          @isOutdatedUpload = true  # if the annotation only has "define" then this measure upload may be out of date.
         @statementViews.push new Thorax.Views.CqlStatement(statement: statement, highlightPatientDataEnabled: @highlightPatientDataEnabled)
 
   ###*
--- a/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
@@ -83,8 +83,6 @@ class Thorax.Views.CqlStatement extends Thorax.Views.BonnieView
       @$('code').attr('class', 'eval-true')
     else
       @$('code').attr('class', 'eval-false')
-  
-  _
 
   clearRationale: ->
     @$('code').attr('class', '')
--- a/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
@@ -40,6 +40,9 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
 
   template: JST['logic/cql_logic']
 
+  # List of statements added by the MAT that are not useful to the user.
+  @SKIP_STATEMENTS = ["SDE Ethnicity", "SDE Payer", "SDE Race", "SDE Sex"]
+
   events:
     "ready": ->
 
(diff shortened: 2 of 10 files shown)


app/assets/javascripts/templates/logic/cql_logic.hbs
```diff
@@ -1,3 +1,12 @@
-<div class="measure-viz rationale panel-group">
-  <div class="cql-editor" id="editor"></div>
+{{#if isOutdatedUpload}}
```
holmesie: This won't be required when we eventually deploy to staging. I'd recommend adding a TODO here mentioning that.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -35,41 +33,157 @@ class Thorax.Views.CqlPopulationsLogic extends Thorax.LayoutView
...
-
+###*
+# View for CQL Logic. Contains CqlStatement views.
+###
 class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
 
   template: JST['logic/cql_logic']
 
+  # List of statements added by the MAT that are not useful to the user.
+  @SKIP_STATEMENTS = ["SDE Ethnicity", "SDE Payer", "SDE Race", "SDE Sex"]
```
holmesie, line 44: Why the capitals? Coffeescript doesn't have a `const`, and I don't think we do this anywhere else. Doesn't the `@` make it important enough?

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -35,41 +33,157 @@ class Thorax.Views.CqlPopulationsLogic extends Thorax.LayoutView
       isActive:  population is population.measure().get('displayedPopulation')
       populationTitle: population.get('title') || population.get('sub_id')
 
-
+###*
+# View for CQL Logic. Contains CqlStatement views.
+###
 class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
```
holmesie, line 39: Thoughts on moving this class to its own file? It seems big enough.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -35,41 +33,157 @@ class Thorax.Views.CqlPopulationsLogic extends Thorax.LayoutView
...
+  # Initializes the view. Creates the CqlStatement views.
+  # Expects model to be a Measure model object of a CQL measure.
+  ###
   initialize: ->
-
-  context: -> _(super).extend cqlLines: @model.get('cql').split("\n")
-
+    @isOutdatedUpload = false
+    @statementViews = []
+    _.each @model.get('elm').library.statements.def, (statement) =>
```
holmesie: Maybe add a few `?` null checks here?

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -35,41 +33,157 @@ class Thorax.Views.CqlPopulationsLogic extends Thorax.LayoutView
...
+      if statement.annotation
+        if (statement.annotation[0].s.value[0] == "define ")
```
holmesie: Why the space at the end of `"define "`? Is that what is in the ELM? A comment might help with my confusion about this.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -35,41 +33,157 @@ class Thorax.Views.CqlPopulationsLogic extends Thorax.LayoutView
...
+    @name = @statement.name
+
+  ###*
+  # Show the results of this statement's calculation by highlighing appropiately. 
+  # @param {boolean|Object[]} result - The result for this statement. May be a boolean or an array of entries.
+  ###
+  showRationale: (result) ->
+    @latestResult = result
+
+    if result == true  # Specifically a boolean true
```
holmesie: I think there is similar logic in at least the cql_calculator (checking for the result type). Since this is the second place this logic exists, maybe we should move this to a central location. Thoughts?

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -35,41 +33,157 @@ class Thorax.Views.CqlPopulationsLogic extends Thorax.LayoutView
...
+    'mouseover code': 'highlightEntry'
+    'mouseout code': 'clearHighlightEntry'
+
+  ###*
+  # Initializes the CqlStatement view. Expects statement to be the JSON ELM statement that should be shown.
+  # highlightPatientDataEnabled can optionally be set to true when constructing to turn on the highlighing patient 
+  # data functionality.
+  ###
+  initialize: ->
+    @text = @statement.annotation[0].s.value[0]
```
holmesie: Maybe null check here? Who knows what this structure might look like in weird cases.

holmesie (changes requested):

app/assets/javascripts/templates/logic/cql_logic.hbs
hossenlopp: Won't we need it when we deploy to staging since some people may still have uploaded on previous versions of the translation service?

holmesie: Typo, meant production.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
hossenlopp: I was following general guidelines that have the constant things capitalized. I think you may be right that in some places in Bonnie we don't capitalize constants. Here is a relevant article on the topic. https://coffeescript-cookbook.github.io/chapters/classes_and_objects/class-variables-and-instance-variables.

Let me know if you think I should change it.

app/assets/javascripts/templates/logic/cql_logic.hbs
hossenlopp: Ahh yes. I will add the comment on that so we remember to remove once CQL settles down for production.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
hossenlopp: I'll add some comments explaining this check better. The space comes after the "define" in the clause level annotation probably because it needs the space to end the keyword.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
hossenlopp: This is looking at the specific result for a statement (define). I will look at that code and see if it makes sense.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
holmesie: Nope, your link was very informative. I think you should keep it.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
hossenlopp: In QDM land these are in the same file. I think The CQL Statement view should be moved to it's own file.

jbradl11 (approved):
Looks good.

holmesie (approved):

hossenlopp:
Rebased against the cql4bonnie branch.

c-monkey:
I tested this against "TESTCMS9_GA" from https://oncprojectracking.healthit.gov/support/browse/BONNIE-237 and am confused by the highlighting. Let's discuss.

c-monkey (approved):
This runs locally and looks great.

PR #728, opened 2017-05-16
Reenabling QRDA export feature for CQL measures

holmesie:
Reenabling QRDA export feature for CQL measures.

Requires https://github.com/projecttacoma/bonnie_bundler/pull/83

NOTE: You need to upload a new copy of a measure to get QRDA exporting properly.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -31,7 +31,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/bonnie_bundler.git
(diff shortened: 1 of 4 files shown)


mayerm94 (approved):
Tested locally with QDM and CQL measures, QRDA correctly exports.

c-monkey (approved):
I kinda wish we had polymorphism here, but understand that we don't want to do that in mongo

jbradl11 (approved):

PR #730, opened 2017-05-18
Fix for highlight suppression on multiple population measures

hossenlopp:
On multiple population set measures, changing population sets created new views without the suppression setting persisted. This fix passes the suppression option to the PopulationsLogic view so it can be properly passed to the PopulationLogic views when the use changes the population set.

Test this by using any measure with multiple population sets and switching to a measure other than the first one and rolling over the measure logic. If no errors result, then it was properly suppressed. To ensure suppression doesn't go too far, test the data element highlighting in the patient builder.

### DIFF
--- a/app/assets/javascripts/views/measure_view.js.coffee
+++ b/app/assets/javascripts/views/measure_view.js.coffee
@@ -56,7 +56,7 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
 
     # display layout view when there are multiple populations; otherwise, just show logic view
     if @populations.length > 1
-      @logicView = new Thorax.Views.PopulationsLogic collection: @populations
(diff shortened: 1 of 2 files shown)


jbradl11:
Considering we are now adding this suppression check everywhere, should we add it in patient_bank_view.js.coffee? Where its creating PopulationLogic and PopulationsLogic views. I don't believe it makes a difference in this case. I was mainly thinking this would act as a sanity check.

alexanderelliott121:
I've tested it, and I'm not seeing the flood of error messages.
I concur if there are other places where the method is called they should be updated, for consistencies' sake.

jbradl11 (approved):
Looks good.

c-monkey (approved):
tested locally

PR #731, opened 2017-05-19
Change the way strats are displayed in CQL

mayerm94:


mayerm94:
[pauldcms126v4_v5_0_Artifacts (2).zip](https://github.com/projecttacoma/bonnie/files/1015253/pauldcms126v4_v5_0_Artifacts.2.zip)
This can be tested with any measure with strats, attached is one from the original JIRA task. This should be tested with HDS branch cql4bonnie_546

mayerm94:
This will not yet work with multiple strats/pops

PR #737, opened 2017-05-29
coverage is broken -- pending revised CQL annotations

c-monkey:


mayerm94:
Is this broken for both QDM and CQL logic? If it works for QDM could this be turned into a conditional disabling of the coverage?

jbradl11:
I agree with Matt. This should be similar to how we "hide" other features that are not supported in CQL yet.

PR #738, opened 2017-05-29
update is broken -- quick fix is to disable update button

c-monkey:


jbradl11 (changes requested):
You should use the same conditional check that we use in other parts of Bonnie. I believe it's called "isCQL". It's a flag that we pass into the handlebars templates. This way we still allow updating of non cql based measures.

jbradl11 (approved):
Looks good now!

mayerm94 (approved):
Tested locally, looks good.

PR #739, opened 2017-05-31
Cql4bonnie 546

mayerm94:
Should be tested with health data standards branch Cql4bonnie_546. The 137 measure (attached) has 2 strats and 2 populations if you look at the HQMF you will see that BOTH populations have BOTH of the stratifications. That is why when you load it, there will be 6 groups that show up, population 1 and 2 and stratifications 1-4. That is pop1,pop2, pop1-strat1, pop1-strat2, pop2-strat1, pop2-strat2. Measure 126, also attached shows that this still works with single pop single strats. Some things to look for is that all of the highlighting and population boxing works, and that the calculations are correct. A known (very small) issue is that in the edit patient view under CQL Calculation results section, the STRAT title still shows up when you are selecting a population criteria.
[cqltest137_v5_0_artifacts.zip](https://github.com/projecttacoma/bonnie/files/1042532/cqltest137_v5_0_artifacts.zip)
[pauldcms126v4_v5_0_Artifacts.zip](https://github.com/projecttacoma/bonnie/files/1042533/pauldcms126v4_v5_0_Artifacts.zip)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -11,7 +11,7 @@ GIT
 
 GIT
   remote: https://github.com/projectcypress/health-data-standards.git
-  revision: 6b73cffd3892452e9cf51416f5e81b1bc36d7dc1
+  revision: a82737e4728a2f9b568089a32cb336b994e7c2c7
   branch: cql4bonnie
   specs:
     health-data-standards (3.6.1)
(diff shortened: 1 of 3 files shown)


c-monkey (approved):
tested with https://github.com/projectcypress/health-data-standards/pull/499

jbradl11 (approved):
Tested and looks good.

PR #743, opened 2017-06-05
Updated exception recipient for master on production

holmesie:


config/environments/production.rb
```diff
@@ -81,7 +81,7 @@
   config.middleware.use ExceptionNotification::Rack, email: {
     email_prefix: "[Bonnie] ",
     sender_address: %{"Bonnie (#{APP_CONFIG['hostname']})" <bonnie@#{APP_CONFIG['hostname']}>},
-    exception_recipients: %w{bonnie-feedback-list@lists.mitre.org},
+    exception_recipients: %w{bonnie-production-errors-list@LISTS.MITRE.ORG},
```
mayerm94, line 84: any reason why LISTS.MITRE.ORG is capitalized?

c-monkey: I don't think it is case sensitive. I don't understand why bonnie-feedback-list is sent to the tacoma list, but my real question is: how do we test this?

c-monkey: We decided @holmesie is going to load production onto the Bonnie Alpha Staging machine (...245) so we can test this email.

holmesie: Pull request example emails sent and received 1:44pm EDT, check your inboxes for verification

c-monkey (approved):
emails look good

PR #746, opened 2017-06-06
Switch default qrda export version to r4 from r3_1

mayerm94:
WE NEED TO WAIT UNTIL https://github.com/projectcypress/health-data-standards/tree/mongoid5 IS MERGED


mayerm94:
I'm not sure what the PR expectations are for a one liner, speak up if there is something you would like to see here.

okeefm:
Do we still need to have projectcypress/health-data-standards#491 merged before this is merged? Because that PR was closed but not merged.

edeyoung:
Was that merged into the mongoid5 branch? Because that's what we're now using. This PR should be redirected to be on top of mongoid5. Then, after these outstanding issues are resolved, it would be good to double check Dave's PR of mongoid5 onto master and work that so everything can be pointed to master again. Thanks!

okeefm (dismissed):
Looks good to me. Works using the same principle as Cypress when we used the HDS exporter.

edeyoung:
Looks like this PR still needs to be moved to sit on top of mongoid5

edeyoung:
Did this get checked with the open tickets related to this? BONNIE-298 and BONNIE-330?

okeefm:
BONNIE-298 continues to fail, and patients will no longer export, because the "Risk Category Assessment" datatype (and therefore template) was removed in the R4 release of QRDA Cat I. 

BONNIE-330 continues to fail, because HDS exports entries with negations with negated valuesets (which looks like `<code nullFlavor="NA" sdtc:valueSet="2.16.840.1.113883.3.600.1.1525"/>`), not negated individual codes. 

Brandy Todd's issue has been fixed, as the Assessment, Performed datatype/template was added in QRDA Category I R4.

okeefm (approved):
QRDA R4 works as expected.

PR #747, opened 2017-06-06
Change default qrda export version from r3_1 to r4

mayerm94:
NEED TO WAIT TILL https://github.com/projectcypress/health-data-standards/pull/491 IS MERGED


edeyoung:
Not relevant for CQL-based measures. Will need entirely new set of QRDA updates.

PR #751, opened 2017-06-13
Set denexcep to 0 when in numer

mayerm94:


mulcahyk (commented):
We should probably add a (documentation) reference here as to why we are doing what we are doing (within handlePopulationValues).

mayerm94:
To test, make a measure pass the logic in numer and denexcep and make sure the denexcep Actual result is set zero

holmesie (approved):

mulcahyk (approved):

c-monkey (approved):
tested locally, has the highlighting bug https://jira.mitre.org/browse/BONNIE-669 but any fix for that is unrelated to this change, as it was already present for other populations

PR #755, opened 2017-06-15
Updated libv8 to make installing on mac less stressful.

hossenlopp:


c-monkey (approved):
tested locally, now you don't need to use brew!

PR #756, opened 2017-06-16
Warning in patient.materialize if the user session expires

hossenlopp:
Addressing an issue for users who leave their Bonnie tab open overnight and try to continue using it...

sessionexpired

The existing alert said there is something wrong with the patient data. No indication that they need to reload.


c-monkey:
Are you planning on cherry picking or repeating this change into the cql4bonnie branch?

hossenlopp:
@c-monkey Should we just wait for the rebase on master? Or is this something needed now? There may also be other places that the session expiration errors could arise. So this may be an incomplete solution.

c-monkey:
Fair enough. I am planning on deploying a new cql4bonnie release this afternoon, but not to production yet, so was curious only for that reason.

app/assets/javascripts/models/patient.js.coffee
```diff
@@ -131,8 +131,12 @@ class Thorax.Models.Patient extends Thorax.Model
       @trigger 'materialize' # We use a new event rather than relying on 'change' because we don't want to automatically re-render everything
       callback() if callback?
       $('#ariaalerts').html "This patient has been updated" #tell SR something changed
-    .fail ->
-      bonnie.showError({title: "Patient Data Error", summary: 'There was an error handling the data associated with this patient.', body: 'One of the data elements associated with the patient is causing an issue.  Please review the elements associated with the patient to verify that they are all constructed properly.'})
+    .fail (xhr, status, error) ->
+      # if status is 422 that means that the user's session has expired
```
mulcahyk, line 135: Like comment.
While we are here, are there any other issues we can catch?

PR #758, opened 2017-06-20
BONNIE-671 Improves performance.

alexanderelliott121:
Changes the algorithm to use maps to reduce number of required iterations.

### DIFF
--- a/app/assets/javascripts/views/measure_value_sets.js.coffee
+++ b/app/assets/javascripts/views/measure_value_sets.js.coffee
@@ -132,23 +132,33 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
     # determines if one or more codes in a value set are in another value set.
     summaryValueSets = @filterValueSets(@summaryValueSets)
 
-    for valueSet1 in summaryValueSets
-      for valueSet2 in _(summaryValueSets).without(valueSet1)
-        matchedCodes = []
-
-        # Because codes are pageable, we need to reference the fullCollection to compare all codes
-        valueSet1.codes.fullCollection.each (code1) =>
-          hasOverlap = valueSet2.codes.fullCollection.some (code2) ->
-            overlapsCode = code2.get('code') == code1.get('code')
-            overlapsCodeSystem = code2.get('code_system_name') == code1.get('code_system_name')
-            return overlapsCode && overlapsCodeSystem
-
-          if hasOverlap then matchedCodes.push(code1)
-        if matchedCodes.length > 0
-          @overlappingValueSets.add
-            cid: valueSet1.cid + "_" + valueSet2.cid
-            codes: new Backbone.PageableCollection(@sortAndFilterCodes(matchedCodes), @pagination_options)
-            oid1: valueSet1.oid
-            name1: valueSet1.name
-            oid2: valueSet2.oid
-            name2: valueSet2.name
+    codeToVs = {}
+    overlapCodes = {}
+    overlapValueSets = {}
+    for curValueSet in summaryValueSets
+      curValueSet.codes.fullCollection.each (curCode) =>
+        workingCode = curCode.get('code') + ":::" + curCode.get('code_system_name')
+        vsAndCode = {valueSet: curValueSet, code: curCode}
+        if codeToVs[workingCode] == undefined
+          codeToVs[workingCode] = []
+        else
+          for overlap in codeToVs[workingCode]
+            overlapKey = overlap['valueSet'].cid + "_" + curValueSet.cid
+            if overlapCodes[overlapKey] == undefined
+              overlapCodes[overlapKey] = []
+            if overlapValueSets[overlapKey] == undefined
+              overlapValueSets[overlapKey] = [overlap['valueSet'], curValueSet]
+            overlapCodes[overlapKey].push(curCode)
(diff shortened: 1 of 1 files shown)


c-monkey:
This was previous last revised in: https://github.com/projecttacoma/bonnie/commit/07474bb69e2b2a7f2851c66272c8836414b38a4e

Alex's change gets rid of bi-directional overlapping codes. Should he add these back in? Do we know the customer's needs?

c-monkey:
This PR is in response to:
https://oncprojectracking.healthit.gov/support/browse/BONNIE-271

On my local machine, run time for the 6/19 production backup version of "laura.vandenbosch@relayhealth.com" account measure CMS166V6
http://localhost:3000/#measures/B6016B47-B65D-4BE0-866F-1D397886CA89
this takes about 70s w/out the fix and 12s with it.

The general issue is these value sets are giant and the algorithm to check for overlapping value sets is very slow.

c-monkey:
screen shot 2017-06-20 at 4 51 59 pm

Screen shot of new (left) versus old (right), the difference is in the overlapping value sets. If you want to play with this yourself, try CMS53V5

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -132,23 +132,33 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
...
+      valueSet1 = overlapValueSets[overlapKey][0]
+      valueSet2 = overlapValueSets[overlapKey][1]
+      matchedCodes = overlapCodes[overlapKey]
+      @overlappingValueSets.add
+        cid: overlapKey
+        codes: new Backbone.PageableCollection(@sortAndFilterCodes(matchedCodes), @pagination_options)
+        oid1: valueSet1.oid
+        name1: valueSet1.name
+        oid2: valueSet2.oid
+        name2: valueSet2.name
```
mulcahyk: Looks like you might be missing an ending CRLF.

hossenlopp (changes requested):
Add some comments explaining what is going on here. Especially explain the hashes used for accumulating data while looping through.

Also add that missing CLRF.

mulcahyk (commented):
I second Chris H's comment.
It would also be good to add back the comment about the codes being pageable.

edeyoung:
Having the lists duplicated I think was a very conscious decision to make sure that the items were as findable as possible. I think that this can still be done with the more efficient approach. I think we could just take the list that's there and copy everything and flip the primary and secondary value sets.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -132,23 +132,33 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
...
+            if overlapValueSets[overlapKey] == undefined
+              overlapValueSets[overlapKey] = [overlap['valueSet'], curValueSet]
+            overlapCodes[overlapKey].push(curCode)
+        codeToVs[workingCode].push(vsAndCode)
+    
+    for overlapKey in Object.keys(overlapValueSets)
+      valueSet1 = overlapValueSets[overlapKey][0]
+      valueSet2 = overlapValueSets[overlapKey][1]
+      matchedCodes = overlapCodes[overlapKey]
+      @overlappingValueSets.add
```
edeyoung, line 171: in order to get both sets of entries like existed before these changes, you can just add another object but swap the `oid1`/`name1` and `oid2`/`name2` entries. This will make the output match what it was before.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -132,13 +132,26 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
     # determines if one or more codes in a value set are in another value set.
     summaryValueSets = @filterValueSets(@summaryValueSets)
 
+    #Map of codes to the ValueSets they are found in.
     codeToVs = {}
+    #Map whose key is a pair of ValueSet CIDs (vs1.cid + "_" + vs2.cid), and whose value is a list of codes shared between them.
     overlapCodes = {}
+    #Map whose key is a pair of ValueSet CIDs (vs1.cid + "_" + vs2.cid), and whose value is the pair of valuesets involved.
     overlapValueSets = {}
+    #Outer loop itterates over valueSets, inner loop itterates over contained codes.
```
c-monkey: There is only 1 't' in the word "iterates"

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -132,13 +132,26 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
...
+    #For each code, check codeToVs to see if that code has been seen before.
+    #If it hasn't populate it with an empty array.
```
c-monkey: Add a comma after "hasn't" and consider rewriting to start this sentence on the previous line.

c-monkey (commented):
I have not had time to test. Have we tested against CMS53V5?

hossenlopp:
Tested CMS53v5 with the latest revision.

image

hossenlopp (approved):

c-monkey (approved):

PR #759, opened 2017-06-22
Use field_title instead of key for title in field value

mayerm94:
The field values are getting their titles form the array of hashes in health-data-standards/lib/hqmf-model/data_criteria.rb and were using the key as the title instead of the field_title 
screen shot 2017-06-22 at 10 22 26 am
screen shot 2017-06-22 at 10 22 50 am

Addresses the problem about "discharge disposition" vs "discharge status" described in https://oncprojectracking.healthit.gov/support/browse/BONNIE-282


edeyoung:
Where in the UI does this issue come up so I can test the fix?

mayerm94:
@edeyoung This happens in the text of any field value that is added in a data criteria.

edeyoung:
screen shot please!

mayerm94:
@edeyoung Add a field value to the data criteria, in this case the field value added was ANATOMICAL_STRUCTURE
screen shot 2017-06-26 at 7 30 04 am

edeyoung:
can you note a measure where a field value is used as part of the logic whose name is significantly changed as a result of this change? I want to confirm that logic calculations are unchanged.

mayerm94:
[EH_CMS71v5_NQF0436_Stroke3.zip](https://github.com/projecttacoma/bonnie/files/1102181/EH_CMS71v5_NQF0436_Stroke3.zip)

screen shot 2017-06-26 at 9 12 11 am
Field Value Ordinality: Principal needs to be added to the  Diagnosis: Active: Hemorrhagic Stroke data criteria in order to pass the IPP

mayerm94:
To check the origin of this bug, see the hmdmlantana account on alpha patient "Expired MSRPOPLEXStrat3Pass" I don't have the MAT package, so you'll have to pull the alpha database backup using the script at https://gitlab.mitre.org/bonnie/server-scripts/blob/master/grab-latest-alpha-backup.rb

edeyoung (approved):

hossenlopp (approved):

ssayer (approved):
Regression Passed

PR #760, opened 2017-06-22
CQL Logic Improvements BONNIE-669

hossenlopp:
-Does not highlight statements that define populations that are not being considered.
-The logic for this is stopgap until we move it and have clause level highlighting.

### DIFF
--- a/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
@@ -100,8 +100,58 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
   ###
   showRationale: (result) ->
     @latestResult = result
+    showResultsMap = @_makePopulationResultShownMap result
     for statementView in @statementViews
-      statementView.showRationale result.get('statement_results')[statementView.name]
+      # check to see if highlighting should be supressed because "not calculated"
+      popName = statementView.cqlPopulation?.split(', ')[0]
+      showHighlighting = if showResultsMap[popName]? then showResultsMap[popName] else true
+      
+      statementView.showRationale(result.get('statement_results')[statementView.name], showHighlighting)
+
+  _makePopulationResultShownMap: (result) ->
+    # initialize to true for every population
+    resultShown = {}
+    _.each(_.without(result.keys(), 'statement_results', 'patient_id'), (population) -> resultShown[population] = true)
+    
+    # If STRAT is 0 then everything else is not calculated
+    if result.get('STRAT')? && result.get('STRAT') == 0
+      resultShown.IPP = false if resultShown.IPP?
+      resultShown.NUMER = false if resultShown.NUMER?
+      resultShown.NUMEX = false if resultShown.NUMEX?
+      resultShown.DENOM = false if resultShown.DENOM?
+      resultShown.DENEX = false if resultShown.DENEX?
+      resultShown.DENEXCEP = false if resultShown.DENEXCEP?
+    
+    # If IPP is 0 then everything else is not calculated
+    if result.get('IPP') == 0
+      resultShown.NUMER = false if resultShown.NUMER?
+      resultShown.NUMEX = false if resultShown.NUMEX?
+      resultShown.DENOM = false if resultShown.DENOM?
+      resultShown.DENEX = false if resultShown.DENEX?
+      resultShown.DENEXCEP = false if resultShown.DENEXCEP?
+
+    # If DENOM is 0 then DENEX, DENEXCEP, NUMER and NUMEX are not calculated
+    if result.get('DENOM')? && result.get('DENOM') == 0
+      resultShown.NUMER = false if resultShown.NUMER?
+      resultShown.NUMEX = false if resultShown.NUMEX?
+      resultShown.DENEX = false if resultShown.DENEX?
+      resultShown.DENEXCEP = false if resultShown.DENEXCEP?
+
+    # If DENEX is 1 then NUMER, NUMEX and DENEXCEP not calculated
+    if result.get('DENEX')? && result.get('DENEX') >= 1
+      resultShown.NUMER = false if resultShown.NUMER?
+      resultShown.NUMEX = false if resultShown.NUMEX?
+      resultShown.DENEXCEP = false if resultShown.DENEXCEP?
+
+    # If NUMER is 0 then NUMEX is not calculated
+    if result.get('NUMER')? && result.get('NUMER') == 0
+      resultShown.NUMEX = false if resultShown.NUMEX?
+
+    # If NUMER is 1 then DENEXCEP is not calculated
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -100,8 +100,65 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
...
+      statementView.showRationale(result.get('statement_results')[statementView.name], showHighlighting)
+
+  ###*
+  # Make a map of population to boolean of if the result of the define statement result should be shown or not. This is
+  # what determines if we don't highlight the statements that are "not calculated".
+  # TODO: This is stop gap solution, should be moved to the calculator.
+  # @private
+  # @param {Result} result = The result object from the calculator.
+  ###
+  _makePopulationResultShownMap: (result) ->
```
edeyoung, line 118: how does this work with multiple populations or multiple stratifications? Is the `result` object already split up by population so you only have one IPP inside, or are there potentially `IPP_1`, `IPP_2`, etc.?

app/assets/javascripts/views/logic/cql_statement_view.js.coffee
```diff
@@ -19,19 +19,28 @@ class Thorax.Views.CqlStatement extends Thorax.Views.BonnieView
...
-    if result == true  # Specifically a boolean true
+    if highlightResult == false  # If the result shouldn't be highlighted
+      @$('code').attr('class', '')
+    else if result == true  # Specifically a boolean true
       @_setResult true
     else if result == false  # Specifically a boolean false
       @_setResult false
     else if Array.isArray(result)  # Check if result is an array
       @_setResult result.length > 0  # Result is true if the array is not empty
+      if result.length == 1 && result[0] == null # But if the array has one element that is null. Then we should make it red.
```
edeyoung, line 36: I know this is an edge case that you're handling - do we have a ticket to investigate why this is something that's being returned? It doesn't seem like this is a good value to have be returned and mean false.

app/assets/javascripts/views/logic/cql_statement_view.js.coffee
```diff
@@ -19,19 +19,28 @@ class Thorax.Views.CqlStatement extends Thorax.Views.BonnieView
...
+        @_setResult false
+    else if result instanceof cql.Interval  # make it green if and Interval is returned
+      @_setResult true
+    else if result == null  # Specifically no result
```
edeyoung, line 40: should no result be colored white instead of false? What does no result mean? Since everything gets executed, I'm guessing it shouldn't be colored white.

However - if these is a valid result than all my grandiose ideas about doing the "should this exist" at the calculation stage before the result gets passed back may not work out. How would we appropriately signal "this shouldn't be processed" if null is taken up already? Maybe a separate hash like what hoss has here?

app/assets/javascripts/views/logic/cql_statement_view.js.coffee
```diff
@@ -48,7 +57,8 @@ class Thorax.Views.CqlStatement extends Thorax.Views.BonnieView
   # Clear the result for this statement.
   ###
   clearRationale: ->
-    @latestResult = null
+    # Clear out the latestResult. Using undefined because it is different than null which is a valid result.
+    @latestResult = undefined
```
edeyoung, line 61: maybe this answers my question above...

edeyoung (changes requested):
Mostly questions

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -100,8 +100,65 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
...
+
+    # If STRAT is 0 then everything else is not calculated
+    if result.get('STRAT')? && result.get('STRAT') == 0
+      resultShown.IPP = false if resultShown.IPP?
+      resultShown.NUMER = false if resultShown.NUMER?
+      resultShown.NUMEX = false if resultShown.NUMEX?
+      resultShown.DENOM = false if resultShown.DENOM?
+      resultShown.DENEX = false if resultShown.DENEX?
+      resultShown.DENEXCEP = false if resultShown.DENEXCEP?
+
```
holmesie, line 131: I really like how readable all of this is! Makes me want to adjust the cql_calculator population rules to look something like this (if it makes sense)...

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
hossenlopp: The calculations are done with the scope of a population set or stratification. So the results are already normalized to their population names. Anyways the `IPP_N` is a structure only in QDM measures in Bonnie. The `populations_cql_map` used to map populations to the CQL statements that define them.

app/assets/javascripts/views/logic/cql_statement_view.js.coffee
hossenlopp: To me, it looks like an issue with the CQL. The issue in cqltest137_v5_0 in our Sharepoint repo, which also is good for testing the populations and stratifications. This check for an array with one `null` allows the logic to be highlighted as it is in this screenshot below when there is are no elements that would resolve in that `First`.
screen shot 2017-06-23 at 10 00 11 am
My guess is that the `{ }` around the `First` is turning the single element result into an array even if the result is nothing.

app/assets/javascripts/views/logic/cql_statement_view.js.coffee
hossenlopp: I'm treating a `null` result as a failed (red) highlight. I am assuming a `null` means that the statement had a single element result from a `First` or `Single`, etc... I think I found a measure that exhibits this instead of wrapping it in an array like the specific check above. I'll look for that again and report back.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
hossenlopp: We do want to move this stuff into the calculator so the result of this logic can be used elsewhere. This is just temporarily here and very verbose because I felt it was quite complicated. It doesn't handle the OBSERV and MSRPOPL stuffs too...

edeyoung:
What is a good alpha account and measure to test this with?

edeyoung (approved):

edeyoung:
My questions were answered and I checked it out locally. looks good!

hossenlopp:
I rebased this on my local machine and noticed that the stratifications are broken. None of the population defining statements are being labeled. Looking into this now.

hossenlopp:
Upon further investigation, I found that uploading the measure again gets the population labels to come back for stratifications. I think something changed in the upload process for stratifications since the measure was loaded into the account I was testing with.

holmesie:
Looks good

holmesie (approved):

c-monkey (approved):
tested locally

PR #763, opened 2017-06-27
Medication order fulfillment delete

edeyoung:
Script for deleting the fulfillment history for medication orders.

Medication orders previously contained fulfillment histories. This actually doesn't make sense as an order is a future concept and a fulfillment history is what happened. This also does not align with the QDM specification. This became an issue when trying to do QRDA exports because the 
medication order QDM model did not support fulfillment history information. The UI for medication orders was changed to align with QDM, but previously entered medication orders still have fulfillment history which is no longer editable in the UI and affects calculations.

The script below removes the fulfillment history information. This will likely make patients fail. We will need to send out an email when we make this change.

**To Test**

See instructions here: https://jira.mitre.org/browse/BONNIE-660?focusedCommentId=97482&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-97482

### DIFF
--- a/lib/tasks/bonnie_run_once.rake
+++ b/lib/tasks/bonnie_run_once.rake
@@ -43,8 +43,11 @@ namespace :bonnie do
     # Medication orders previously contained fulfillment histories. This actually
     # doesn't make sense as an order is a future concept and a fulfillment history
     # is what happened. This also does not align with the QDM specification.
-    #  This became an issue when trying to do QRDA exports because the 
+    # This became an issue when trying to do QRDA exports because the 
     # medication order QDM model did not support fulfillment history information.
+    # The UI for medication orders was changed to align with QDM, but previously
+    # entered medication orders still have fulfillment history which is no longer
+    # editable in the UI and affects calculations.
     #
     # The script below removes the fulfillment history information. This will likely
     # make patients fail. We will need to send out an email when we make this change.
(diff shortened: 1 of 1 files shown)


edeyoung:
This code exists in HQMF2JS (which is where the processing issues are occuring):

```coffeescript
# returns cumulativeMedicationDuration in terms of days
hQuery.Medication::cumulativeMedicationDuration = (dateRange) ->
  #assuming that the dose is the same across fills and that fills is  stated in individual
  #doses not total amount. Will need to flush this out more at a later point in time.
  #Considering that liquid meds are probaly dispensed as total volume ex 325ml with a dose of
  #say 25ml per dose.  Will definatley need to revisit this.
  if this.administrationTiming() && this.dose() && @json['fulfillmentHistory']
    this.fulfillmentTotals(dateRange)
  else if this.administrationTiming() && this.allowedAdministrations()
    # this happens if we have a Medication, Order.
    cumulativeMedicationDuration = this.allowedAdministrations() / this.administrationTiming().dosesPerDay()
    # need to do in case of divide by zero error
    cumulativeMedicationDuration = 0 if isNaN(cumulativeMedicationDuration)
    cumulativeMedicationDuration
```

Should we also add a check for if the medication is an order here? If so, we do "administration timings". Otherwise, we do "fulfillment history" for calculations. I still think it's a good idea to do the script for DB cleanup purposes.

okeefm:
I ran the calculation tests against this code; 4 patients in 1 measure are failing after the rake task is run.

okeefm (approved):

PR #764, opened 2017-06-27
Added support to call new execution engine; and changed views to supp…

jbradl11:
…ort multiple libraries.

This is the final step to integrate multi library support. You should only have to pull this branch down and run it as is. If you have to change the Gemfile please let me know.


holmesie (approved):
Looks good

c-monkey (approved):
tested locally

PR #766, opened 2017-06-30
Updating Gemfile.lock to point to new cql_qdm_patientapi

holmesie:


edeyoung (approved):
saw it working on alpha staging account

PR #769, opened 2017-07-12
Implemented a rake task that rebuilds the ELM for CqlMeasure models i…

holmesie:
…n the local db using a locally running translation service

### DIFF
--- a/lib/tasks/cql.rake
+++ b/lib/tasks/cql.rake
@@ -15,7 +15,7 @@ namespace :bonnie do
           # Grab the user, we need this to output the name of the user who owns
           # this measure. Also comes in handy when detecting measures uploaded
           # by accounts that have since been deleted.
-          user = User.find_by('_id': measure[:user_id])
+          user = User.find_by(_id: measure[:user_id])
           # Grab the measure cql
           if measure[:cql].instance_of?(Array) # New type of measure
             cql = measure[:cql]
--- a/lib/tasks/cql.rake
+++ b/lib/tasks/cql.rake
@@ -12,28 +12,34 @@ namespace :bonnie do
       orphans = 0
       CqlMeasure.all.each do |measure|
         begin
+          # Grab the user, we need this to output the name of the user who owns
+          # this measure. Also comes in handy when detecting measures uploaded
+          # by accounts that have since been deleted.
           user = User.find_by('_id': measure[:user_id])
-          if measure[:cql].instance_of?(Array) # Newer version of the CqlMeasure model
-            # NOTE: The following three lines are adapted from the functionality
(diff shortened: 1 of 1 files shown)


lib/tasks/cql.rake
```diff
@@ -2,6 +2,51 @@
...
+            # NOTE: The following three lines are adapted from the functionality
+            # of the 'load_mat_cql_exports' method in the cql_loader.rb in bonnie_bundler
+            elms, elm_annotations = Measures::CqlLoader.translate_cql_to_elm(measure[:cql])
+            cql_definition_dependency_structure = Measures::CqlLoader.populate_cql_definition_dependency_structure(measure[:main_cql_library], elms, measure[:populations_cql_map])
+            cql_definition_dependency_structure = Measures::CqlLoader.populate_used_library_dependencies(cql_definition_dependency_structure, measure[:main_cql_library], elms)
+            measure.update(elm: elms, cql_statement_dependencies: cql_definition_dependency_structure)
+            measure.save!
+            update_passes += 1
+            puts '[Success] Measure ' + measure[:cms_id] + ': "' + measure[:title] + '" with id ' + measure[:id] + ' in account ' + user[:email] + ' successfully updated ELM!'
+          elsif measure[:cql].instance_of?(String) # Older version of the CqlMeasure model
```
hossenlopp: We should be updating the model here instead of keeping it out of date. I would suggest checking if `cql` is a string and make it an array and then go to town with the code in the if it's an Array part. Note that we'll also have to populate the main_cql_library field.

hossenlopp (changes requested):

lib/tasks/cql.rake
holmesie: Done!

lib/tasks/cql.rake
```diff
@@ -2,6 +2,57 @@
...
+    task :rebuild_elm => :environment do
+      update_passes = 0
+      update_fails = 0
+      orphans = 0
+      CqlMeasure.all.each do |measure|
+        begin
+          # Grab the user, we need this to output the name of the user who owns
+          # this measure. Also comes in handy when detecting measures uploaded
+          # by accounts that have since been deleted.
+          user = User.find_by('_id': measure[:user_id])
```
hossenlopp: `'_id':` isn't a valid symbol. `_id:` will do the trick.

hossenlopp (changes requested):

lib/tasks/cql.rake
holmesie: Fixed

hossenlopp (approved):

PR #773, opened 2017-07-14
BONNIE-800 no end date for patient builder elements with Authored time

pkmitre:
Fixes BONNIE-800

This pull request changes Bonnie so that in the patient builder, patient elements that have an "Authored" time (and therefore no end time to specify in the UI) do not store an end time when being saved. This addresses an issue where, if an element with an "Authored" time was being dragged onto another element with a start and stop time, the stop time would be invisibly stored on the new element and cause possible validation issues.

JIRA test: https://jira.mitre.org/browse/BONNIE-811


lukoktonos (approved):
Passed JIRA test.

holmesie (approved):
Looks good!

c-monkey (approved):
I ran the https://jira.mitre.org/browse/BONNIE-744 regression, this change introduced no changes.

PR #775, opened 2017-07-14
CQL4Bonnie With Coverage, Clause Level Highlighting, and Multi-Library Support

mayerm94:
For CODE REVIEW... you can test it if you want, use the 1.2.8 translator and run the update measures rake task


app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -47,7 +50,7 @@
         if elm_library['library']['identifier']['id'] == population.collection.parent.get('main_cql_library')
           main_library_version = elm_library['library']['identifier']['version']
           main_library_index = index
-
+
```
c-monkey: This whitespace change is sloppy. Also line 73.

lib/tasks/cql.rake
```diff
@@ -2,6 +2,57 @@
...
+      CqlMeasure.all.each do |measure|
+        begin
+          # Grab the user, we need this to output the name of the user who owns
+          # this measure. Also comes in handy when detecting measures uploaded
+          # by accounts that have since been deleted.
+          user = User.find_by(_id: measure[:user_id])
+          # Grab the measure cql
+          if measure[:cql].instance_of?(Array) # New type of measure
+            cql = measure[:cql]
+          elsif measure[:cql].instance_of?(String) # Old type of measure
```
c-monkey, line 22: What if cql isn't an array or a string? Obviously we have exceptions handlers, but why not explicitly call 'raise'?

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -1,5 +1,8 @@
 @CQLCalculator = class CQLCalculator extends Calculator
 
+  # List of statements added by the MAT that are not useful to the user.
+  @SKIP_STATEMENTS = ["SDE Ethnicity", "SDE Payer", "SDE Race", "SDE Sex"]
```
c-monkey: This is duplicated from Thorax.Views.CqlPopulationLogic could it be shared?

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+    if result is true  # Specifically a boolean true
+      return true
+    else if result is false  # Specifically a boolean false
+      return false
+    else if Array.isArray(result)  # Check if result is an array
+      if result.length == 0  # Result is true if the array is not empty
+        return false
+      else if result.length == 1 && result[0] == null # But if the array has one element that is null. Then we should make it red.
+        return false
+      else
```
lukoktonos: Can the array ever have multiple null values?

Gemfile
```diff
@@ -14,10 +14,10 @@ gem 'less-rails'
 gem "non-stupid-digest-assets"
 
 gem 'health-data-standards', :git => 'https://github.com/projectcypress/health-data-standards.git', :branch => 'cql4bonnie'
-gem 'cql_qdm_patientapi', :git => 'https://github.com/projecttacoma/cql_qdm_patientapi.git', :branch => 'cql4bonnie'
+gem 'cql_qdm_patientapi', :git => 'https://github.com/projecttacoma/cql_qdm_patientapi.git', :branch => 'cql4bonnie_706_707'
```
holmesie, line 17: This is fine for testing, but why is this included in a pull request?

Gemfile
```diff
@@ -14,10 +14,10 @@ gem 'less-rails'
...
 gem 'simplexml_parser', :git => 'https://github.com/projecttacoma/simplexml_parser.git', :branch => 'master'
 gem 'hqmf2js', :git => 'https://github.com/projecttacoma/hqmf2js.git', :branch => 'master'
-gem 'bonnie_bundler', :git => 'https://github.com/projecttacoma/bonnie_bundler.git', :branch => 'cql4bonnie'
+gem 'bonnie_bundler', :git => 'https://github.com/projecttacoma/bonnie_bundler.git', :branch => 'statement_dep_and_clause_elm_merged'
```
holmesie, line 20: Same as above

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+        # If retrieving the STRAT, set the index to the correct STRAT in the cql_map
+        index = populationSet.get('stratification_index') if population == "STRAT" && populationSet.get('stratification')?
+
+        relevantStatement = measure.get('populations_cql_map')[population][index]
+        @_markStatementRelevant(measure.get('cql_statement_dependencies'), statementRelevance, measure.get('main_cql_library'), relevantStatement, relevance)
+
+    return statementRelevance
+
+  ###*
+  # Mark a statement as relevant if it hasn't already been marked relevant. Mark all dependent statement
```
c-monkey: This comment confused me since relevance is 3 values, no longer a boolean.

How about replace comment with:
Mark statement and its dependent statements relevance. Can change relevance from 'NA' to 'TRUE' or 'FALSE', can change 'FALSE' to 'TRUE', but does not change 'TRUE' if previously identified.

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -114,7 +119,7 @@
         cql_population = defined_pops[index]
         # Is there a patient result for this population? and does this populationCriteria contain the population
         # We need to check if the populationCriteria contains the population so that a STRAT is not set to zero if there is not a STRAT in the populationCriteria
-        if results['patientResults'][patient.id][cql_population]? && population.get(popCode)?
```
holmesie: Why was this check removed?

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
         population_results["DENEXCEP"] = 0
     return population_results
 
+  ###*
```
holmesie, line 214: This is not the style of function comments we've been using for this file. If you are going to start using this style, you should change all of the other function comments.

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+  # Build a map of population to boolean of if the result of the define statement result should be shown or not. This is
+  # what determines if we don't highlight the statements that are "not calculated".
+  # @private
+  # @param {Result} result - The population result object.
+  # @return {object} Map that tells if a population calculation was considered/relevant.
+  ###
+  _buildPopulationRelevanceMap: (result) ->
```
holmesie: This looks like view related code; why is it in this file?

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+
+  ###*
+  # Builds a map of relevance for the statements in this calculation.
+  # @private
+  # @param {object} populationRelevance - The map of population relevance used as the starting point.
+  # @param {Measure} measure - The measure.
+  # @param {population} populationSet - The population set being calculated.
+  # @return {object} The statement_relevance map that tells if each statement was relevant for calculation.
+  ###
+  _buildStatementRelevanceMap: (populationRelevance, measure, populationSet) ->
```
holmesie: Same as previous comment; this looks like view related code, why is it in this file?

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+  ###*
+  # Mark a statement as relevant if it hasn't already been marked relevant. Mark all dependent statement
+  # as relevant.
+  # @private
+  # @param {object} cql_statement_dependencies - Dependency map for the measure.
+  # @param {object} statementRelevance - The relevance map to fill.
+  # @param {string} libraryName - The library name.
+  # @param {string} statementName - The statement name.
+  ###
+  _markStatementRelevant: (cql_statement_dependencies, statementRelevance, libraryName, statementName, relevant) ->
```
holmesie: And again, same as previous comment; this looks like view related code, why is it in this file?

app/assets/javascripts/models/coverage.js.coffee
```diff
@@ -3,27 +3,46 @@ class Thorax.Model.Coverage extends Thorax.Model
...
+                if allClauses[key]
+                  if !@rationaleCriteria[libraryName]?
+                    @rationaleCriteria[libraryName] = []
+                  @rationaleCriteria[libraryName][localId] = clauseResult
+                
+        # Count total number of clauses that evalueated to true
+        for localId of allClauses
+          if allClauses[localId]
+            passedClauses += 1
+        # Set coverage to the ratio of evaluated clauses to total clauses
```
lukoktonos: I think 'ratio' should be 'percentage' here

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+  ###*
+  # Builds the result structures for the statements and the clauses.
+  # @private
+  # @param {Measure} measure - The measure.
+  # @param {object} rawStatementResults - The raw statement results from the calculation engine.
+  # @param {object} rawClauseResults - The raw clause results from the calculation engine.
+  # @param {object} statementRelevance - The statement relevance map.
+  # @return {object} Object with statement_results and clause_results built.
+  ###
+  _buildStatementAndClauseResults: (measure, rawClauseResults, statementRelevance) ->
```
holmesie: Same as previous comment

app/assets/javascripts/models/coverage.js.coffee
```diff
@@ -3,27 +3,46 @@ class Thorax.Model.Coverage extends Thorax.Model
...
-      @set coverage: ( covered_dc_count * 100 / dc_count ).toFixed()
+      # Coverage is 0 if there are no patients
+      if @clauseResults.length == 0
+        @set coverage: 0
+      else
+        allClauses = {}
+        @rationaleCriteria = {}
+        totalClauses = 0
+        passedClauses = 0
+        # Initialize all_clauses list to all false and count number of clauses
```
lukoktonos: change all_clauses in comment to be camel case to match variable name

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+  # @private
+  # @param {object} rawClauseResults - The raw clause results from the calculation engine.
+  # @param {object} statementRelevance - The statement relevance map.
+  # @param {object} statementName - The name of the statement the clause is in
+  # @param {object} lib - The name of the libarary the clause is in
+  # @param {object} localId - The localId of the current clause
+  # @param {object} clause - The clause we are getting the final result of
+  # @param {Array|Object|Interval|??} rawResult - The raw result from the calculation engine.
+  ###
+  _setFinalResults: (params) ->
```
holmesie: Same as previous comment, this seems way too much unrelated stuff to add to the cql_calculator

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+
+  ###*
+  # Finds the clause localId for a statement and gets the result for that clause.
+  # @private
+  # @param {Measure} measure - The measure.
+  # @param {string} libraryName - The library name.
+  # @param {string} statementName - The statement name.
+  # @param {object} rawClauseResults - The raw clause results from the engine.
+  ###
+  _findResultForStatementClause: (measure, libraryName, statementName, rawClauseResults) ->
```
holmesie: Same as previous comment

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+    statement = library.library.statements.def.find((statement) -> statement.name == statementName)
+    return rawClauseResults[libraryName]?[statement.localId]
+
+  ###*
+  # Determines if a result (for a statement or clause) is a pass or fail.
+  # @private
+  # @param result - The result from the calculation engine.
+  # @return {boolean} true or false
+  ###
+  _doesResultPass: (result) ->
```
holmesie: Same as previous comment

app/assets/javascripts/models/measure.js.coffee
```diff
@@ -125,6 +126,95 @@ class Thorax.Models.Measure extends Thorax.Model
...
+        return @_localIdCache[libraryName][statementName]
+
+  ###*
+  # Finds all localIds in a statement by it's library and statement name.
+  # @private
+  # @param {string} libraryName - The name of the library the statement belongs to.
+  # @param {string} statementName - The statement name to search for.
+  # @return {Hash} List of local ids in the statement.
+  ###
+  _findAllLocalIdsInStatementByName: (libraryName, statementName) ->
```
holmesie: Is this view related? Why is it in the model?

app/assets/javascripts/models/measure.js.coffee
```diff
@@ -125,6 +126,95 @@ class Thorax.Models.Measure extends Thorax.Model
...
+
+    return localIds
+
+  ###*
+  # Finds all localIds in the statement structure recursively.
+  # @private
+  # @param {Object} statement - The statement structure or child parts of it.
+  # @return {Array[Integer]} List of local ids in the statement.
+  ###
+  _findAllLocalIdsInStatement: (statement, libraryName, localIds, aliasMap, emptyResultClauses) ->
```
holmesie: See previous comment

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+    return statementRelevance
+
+  ###*
+  # Mark a statement as relevant if it hasn't already been marked relevant. Mark all dependent statement
+  # as relevant.
+  # @private
+  # @param {object} cql_statement_dependencies - Dependency map for the measure.
+  # @param {object} statementRelevance - The relevance map to fill.
+  # @param {string} libraryName - The library name.
+  # @param {string} statementName - The statement name.
```
c-monkey: missing 'relevant' param

app/assets/javascripts/models/coverage.js.coffee
```diff
@@ -3,27 +3,46 @@ class Thorax.Model.Coverage extends Thorax.Model
...
+                  if !@rationaleCriteria[libraryName]?
+                    @rationaleCriteria[libraryName] = []
+                  @rationaleCriteria[libraryName][localId] = clauseResult
+                
+        # Count total number of clauses that evalueated to true
+        for localId of allClauses
+          if allClauses[localId]
+            passedClauses += 1
+        # Set coverage to the ratio of evaluated clauses to total clauses   
+        @set coverage: ( passedClauses * 100 / totalClauses ).toFixed()
```
lukoktonos: Should there be a check to make sure this always stays between 0 and 100?

app/assets/javascripts/models/coverage.js.coffee
```diff
@@ -35,3 +54,18 @@ class Thorax.Model.Coverage extends Thorax.Model
 
       # Set coverage to the fraction of measure criteria that were true in the rationale
       @set coverage: ( @rationaleCriteria.length * 100 / @measureCriteria.length ).toFixed()
+
+  determineCovered: (clause) ->
+    if clause["final"] == "TRUE"
```
lukoktonos: might be able to just say `return clause["final"] == "TRUE"`

app/assets/stylesheets/cql_logic.less
```diff
@@ -13,4 +13,41 @@
...
+
+    span {
+      border-bottom-width: 4px;
+      line-height: 1.4;
+    }
+
+    span.clause-true {
+      background-color: @state-success-bg;
+      color: @state-success-text;
+      border-bottom-color: @state-success-text !important;
```
holmesie: Generally it's not the nicest thing seeing `!important` in CSS, why is that necessary here?

lib/tasks/cql.rake
```diff
@@ -2,6 +2,57 @@
 namespace :bonnie do
   namespace :cql do
 
+    desc %{Recreates the JSON elm stored on CQL measures using an instance of
+      a locally running CQLTranslationService JAR.
+
+    $ rake bonnie:cql:rebuild_elm}
+    task :rebuild_elm => :environment do
```
holmesie, line 9: This looks pretty good!

holmesie (changes requested):

app/assets/javascripts/cql_calculator.js.coffee
c-monkey, line 214: This language reads too awkward for me. How about:
Build a population to boolean map indicating if each define statement result should be highlighted, thus statements that are "not calculated" are unhighlighted in the UI.

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+      for statementName of statements
+        statementRelevance[lib][statementName] = "NA"
+
+    for population, relevance of populationRelevance
+      # If the population is values, that means we need to mark relevance for the OBSERVs
+      if (population == 'values')
+        for observation in measure.get('observations')
+          @_markStatementRelevant(measure.get('cql_statement_dependencies'), statementRelevance, measure.get('main_cql_library'), observation.function_name, relevance)
+      else
+        index = populationSet.get('index')
```
c-monkey: index is calculated identically each time through this loop except for a single "if"

How about we pull this calculation out of the "for"?

Gemfile
hossenlopp: This is a first pass PR for just reviewing code. Having the gemfile changes is here so people don't have to remember to checkout the correct branches and modify their gemfile to point to the other local checkouts. This also allows us to put it on ALPHA-STG, which we have done for a short period yesterday.

Gemfile
hossenlopp: Same response as above.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Looking at the commit that was made the log message provides more context. https://github.com/projecttacoma/bonnie/commit/072de32102ed34bd12284514a9b78e8987fac406

So... I believe I removed it because the coffeescript `?` was effectively causing to not look at the result if it was `null` or `false`. But those ARE valid results and they are handled below. Perhaps someone should look at that... Perhaps this code should just call `@_doesResultPass`.

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -67,13 +70,15 @@
 
       # Calculate results for each CQL statement
       results = executeSimpleELM(elm, patientSource, @valueSetsForCodeService(), population.collection.parent.get('main_cql_library'), main_library_version, params)
-
+      
       # Parse CQL statement results into population values
       population_results = @createPopulationValues population, results, patient, observation_defs
 
       if population_results?
```
c-monkey: @hossenlopp as we discussed, add a comment for why we are returning all these values in the hash.

I like your suggestion that we document each of these hash values.

app/assets/javascripts/cql_calculator.js.coffee
```diff
@@ -187,6 +192,233 @@
...
+  _buildStatementAndClauseResults: (measure, rawClauseResults, statementRelevance) ->
+    statementResults = {}
+    clauseResults = {}
+    emptyResultClauses = []
+    for lib, statements of measure.get('cql_statement_dependencies')
+      statementResults[lib] = {}
+      clauseResults[lib] = {}
+      for statementName of statements
+        rawStatementResult = @_findResultForStatementClause(measure, lib, statementName, rawClauseResults)
+        statementResults[lib][statementName] = { raw: rawStatementResult}
```
c-monkey: like below in line 338, I think it would add clarity to call out that this section determines the statement results

app/assets/javascripts/views/logic/cql/cql_clause.js.coffee
```diff
@@ -0,0 +1,98 @@
+class Thorax.Views.CqlClauseView extends Thorax.Views.BonnieView
```
c-monkey: this is the only module in the 'cql' folder, we should get rid of the folder, since related modules (and this one, too) are all prefixed with 'cql_'

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -53,24 +53,17 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
...
+    if @model.get('elm_annotations')?
+      for libraryName, annotationLibrary of @model.get('elm_annotations')
+        for statement in annotationLibrary.statements
+          # skip if this is a statement the user doesn't need to see
+          if !Thorax.Views.CqlPopulationLogic.SKIP_STATEMENTS.includes(statement.define_name) && statement.define_name?
             popNames = []
+            popName = null
             # if a population (population set) was provided for this view it should mark the statment if it is a population defining statement  
             if @population
               for pop, popStatements of @model.get('populations_cql_map')
```
c-monkey, line 66: this block of code is the same as line 285 of _buildStatementRelevanceMap (replacing population with populationSet)

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: This is view and coverage related code. Some of the big work in this branch to allow coverage calculation to work and be consistent with patient result highlighting is to use the same exact set of `clause_result`s that get returned from the CQLCalculator.

This function in particular determines which populations were even considered for calculation for this patient.  The structure generated here is used later to determine which CQL define statements should be considered touched (aka. relevant) in for this population calculation.

I will add some more comments to this function.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Same as above comment in general. It is used to determine the `clause_results` used in coverage calculation and logic views.

This function in particular makes use of the population_relevance structure generated in `_buildPopulationRelevanceMap` to determine the statements that are relevant in the calculation for this patient.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Similar to above comment. This is a helper recursive function for `_buildStatementRelevanceMap`.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Similar to above. This generates `statement_results` and `clause_results` that are added to the Result's attributes. Coverage and logic highlight both use these results. Right now I think there may be no usage of `statement_results` but I see it as pretty important for ease of debugging and for the inevitable implementation of patient dashboard for CQL.

app/assets/javascripts/models/coverage.js.coffee
```diff
@@ -3,27 +3,46 @@ class Thorax.Model.Coverage extends Thorax.Model
...
-      all_dc = []
-      for data_criteria, details of @population.measure().get('data_criteria')
-        if details.type != 'characteristic'
-          all_dc.push(data_criteria)
-      dc_count = all_dc.length
-
-      # Count number of covered data criteria
-      covered_dc = []
-      for patient in @population.measure().get('patients').models
-        covered_dc = covered_dc.concat(new CQL_QDM.CQLPatient(patient).getUniqueCoveredDataCriteria())
```
c-monkey, line 22: with this change, we should remove the placeholder function getUniqueCoveredDataCriteria from CQL_QDM.CQLPatient

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Similar to above. Used in coverage calculation and logic highlighting. (I already mentioned this in the function comment). I'm going to fill more info into the code comment.

This function in particular figures out if a clause should be NA, UNHIT, TRUE or FALSE.

- NA - Means Not Applicable. This means the clause was not relevant to this population set at all. This won't be highlighted or considered a clause that needs to be covered.
- UNHIT - This specific patient doesn't hit this clause. This is most likely because the statement it resides was not relevant to a population calculation for this patient.
- TRUE - This clause is relevant and has a truthy result. `_doesResultPass` determines this.
- FALSE - This clause is relevant, is hit, and has a falsey result.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Similar to previous response. Important for coverage and logic highlighting.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Similar to previous response. Important for coverage and logic highlighting.

This function in particular is the sole source of if a statement or clause result is truthy or falsey (aka. highlight green or highlight red).

c-monkey (changes requested):
I'm generally happy with this PR.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Yes, if a CQL measure does something like this:
```
define "Array of two nulls": { first("something with no result"), first("somethingelse with no result") }
```

app/assets/javascripts/models/coverage.js.coffee
mayerm94: https://github.com/projecttacoma/cql_qdm_patientapi/tree/cql4bonnie_706_707

c-monkey: Don't forget to make the PR for this.

app/assets/javascripts/models/coverage.js.coffee
mayerm94: There was never a check on QDM coverage. Not a bad idea, but I'm not sure what we would do about it if it ever fell outside those bounds.

app/assets/javascripts/cql_calculator.js.coffee
jbradl11: I think even if the code is related to the view,  it's ok to be here for now. In the future, it might make sense to move alot of the logic here in a separate file.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: I believe it can. I will look into this.

app/assets/javascripts/cql_calculator.js.coffee
holmesie: @jbradl11 I disagree, this change should be made now, and not relegated to the "never-going-to-happen" future.

The code itself looks good, but I strongly believe it does not belong here. This file is already way too big (compare it to the QDM version).

app/assets/javascripts/cql_calculator.js.coffee
holmesie: Make sense

lib/tasks/cql.rake
mayerm94: This file was already reviewed in another PR

c-monkey: https://github.com/projecttacoma/bonnie/pull/769

If we don't need a performance optimization - Alex says this takes only a minute or two - then as long as we're okay with the spam, doesn't matter to me.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: We will not use `@_doesResultPass` right now because this may deal with collecting the results for episode of care based numbers, this means they need to know the number of things in an Array result.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: This is getting moved to a new helper file in a helpers folder.

app/assets/stylesheets/cql_logic.less
hossenlopp: This was a copy and paste from a previous implementation where it was needed. Removed them in ff8c5170de4ef78e0236531539970b5cd003e725.

app/assets/javascripts/models/coverage.js.coffee
hossenlopp: @mayerm94 handled this.

app/assets/javascripts/models/coverage.js.coffee
hossenlopp: @mayerm94 handled this.

lib/tasks/cql.rake
hossenlopp: Since the beginning of cql4bonnie having CqlMeasure model that model has had `cql` as a field. The change to multi_library support meant we changed the type from `String` to `Array`. We need to handle upgrading to `cql` to an `Array` if isn't already.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: @mayerm94 addressed this

app/assets/javascripts/models/measure.js.coffee
hossenlopp: We had a discussion about this. I forgot the result. This creates a structure relevant any calculation. It's additional calculated data that we could generate on the back end at upload, but having at the front end makes it easier to fix if there's an issue because it won't need a re-upload or rake task.

Now that I remember, this was something is ok to have on the Measure model, but the functions for calculating it should be shifted to a helper file. I will do that.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Oh gosh, I hate this block of code. It was copy pasted from the sister view code in `cql_logic_view.js.coffee`.

Basically what it is doing here is looking at each population in the population set (or stratification), figuring out which CQL statement defines the population. This should be moved to the Measure or the Population thorax model so it can be shared by calculator code and the logic view, since they both need to know which CQL statement defines which population.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
hossenlopp: Sure is... I commented on the other copy of this. I decided it belongs somewhere both calculator and this code can handle it. I have it as a tasker to refactor to a shared location.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: Will move this to `constants.js.coffee.erb`.

c-monkey (approved):
My comments have been addressed.

c-monkey (changes requested):
@hossenlopp says more changes are comin'

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: This has been partially addressed. The rest of the functions will not be updated right now due to priorities.

app/assets/javascripts/cql_calculator.js.coffee
hossenlopp: @mayerm94 addressed this.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
hossenlopp: @mayerm94 Addressed this.

holmesie (approved):

jbradl11 (approved):
Looks good.

PR #780, opened 2017-07-15
updated gemfile.lock to point to the QDM 5.3 parsing changes, the res…

edeyoung:
…ult decimal and integer changes, and the single source query changes.


holmesie (approved):
Looks good

PR #781, opened 2017-07-16
Regression Testing Script

ssayer:
This script uses selenium+chrome to automate the Bonnie regression test. It logs results to the terminal and will note any regressions it finds.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -54,6 +54,8 @@ group :test, :development, :ci do
   gem 'minitest'
   gem 'webmock'
   gem 'vcr'
+  gem 'thor'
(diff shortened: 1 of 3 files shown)


config/mongoid.yml
```diff
@@ -1,7 +1,7 @@
 development:
   sessions:
     default:
-      database: bonnie_development
+      database: bonnie_alpha
```
edeyoung: revert this back to bonnie_development for the commit

lukoktonos:
Moving this script to gitlab/bonnie/server-scripts.

PR #782, opened 2017-07-16
Display warning for any Internet Explorer users.

lukoktonos:
- Add gem 'browser'
  - Replace _ie8_alert.html.erb with _ie_alert.html.erb
    This now detects any Internet Explorer rather than just versions <8
  - Replace references to 'ie8_alert' with 'ie_alert'

JIRA Test: https://jira.mitre.org/browse/BONNIE-845

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,3 +84,6 @@ gem 'foreman'
 
 gem 'handlebars_assets', '0.16'
 gem 'jquery-rails'
+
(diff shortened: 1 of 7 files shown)


holmesie (approved):
Tested with ie, edge, chrome, firefox, and safari; behaves as expected; looks good!

alexanderelliott121 (approved):

PR #783, opened 2017-07-16
Made error catching more generic in order for the elm regeneration script to not fail when a single measure fails

holmesie:
Made error catching more generic in order for the elm regeneration script to not fail when a single measure fails


edeyoung:
this will be added onto the PR that catches errors on load.

PR #785, opened 2017-07-16
Bonnie 831 quickfix based cql error display

edeyoung:


edeyoung:
this change needs to be modified to be on the correct branch. just doing a PR didn't do this automatically.

PR #786, opened 2017-07-16
Bonnie 831 quickfix based cql error display

edeyoung:


edeyoung:
this PR has already been reviewed but against a different branch. That has been re-confirmed.

PR #788, opened 2017-07-17
Bonnie reverts

holmesie:


edeyoung:
Not necessary to include. Updated to new coloring scheme which includes these changes.

PR #795, opened 2017-07-20
BONNIE-776 Fixes strat selection after save issue.

alexanderelliott121:
JIRA Ticket: https://jira.mitre.org/browse/BONNIE-776
JIRA Test: https://jira.mitre.org/browse/BONNIE-876

### DIFF
--- a/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
+++ b/app/assets/javascripts/views/logic/cql_logic_view.js.coffee
@@ -2,10 +2,20 @@ class Thorax.Views.CqlPopulationsLogic extends Thorax.LayoutView
 
   template: JST['logic/cql_layout']
 
+  ###*
+  # This is a button click event handler. The button being clicked is the population tab in the cql_layout.hbs.
+  # This function figures out which population was clicked then updates the view accordingly.
+  # @param {Event} e - JS click event.
+  ###
   switchPopulation: (e) ->
     population = $(e.target).model()
     @switchToGivenPopulation(population)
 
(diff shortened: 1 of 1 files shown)


jbradl11 (approved):
Ran Jira test locally. Looks good.

hossenlopp (approved):

hossenlopp:
`bundle exec rake test` and `bundle exec rake teaspoon` both passed locally. Travis' mongod is having issues with our tests out of nowhere.

edeyoung:
Given that travis is borking after just a comment was added, had passed previously, and appears to failing for reasons unrelated to this PR, I am approving and merging.

mayerm94:
I checked out the failing travis build. The failing tests are due to that duplicate key issue because mongo was not successfully dropping the collections before each test. My solution to this was in test/test_helper.rb change  
```
def dump_database
    Mongoid.default_session.collections.each do |c|
      c.drop()
    end
  end
```
to.....
```
dump_database
    Mongoid.default_session.drop()
 end
```

hossenlopp:
@mayerm94 But why is it just suddenly happening out of nowhere? Does this fix work for all versions of mongoid and mongo though?

mayerm94:
@hossenlopp I know this is not a good answer, but it seems like it starts happening randomly. It started happening randomly on my machine and on Sam's machine. I think it is a problem with newer versions of mongo (possibly due to a deprecated function?). My machine no longer behaves like this, I just wanted to offer what my solution was when it acted up.

PR #797, opened 2017-07-20
Changed database drop command to comply with new versions of mongo

hossenlopp:
This is to fix issues with travis and some of the team having updated versions of mongo that have issues with dropping individual collections. This address BONNIE-886. This still works with older versions of mongo.


mayerm94 (approved):

edeyoung (approved):
tests run on my machine

jbradl11 (approved):
Tested locally. Looks good.

PR #798, opened 2017-07-20
Adjusted observation handler code to account for zero values

holmesie:
JIRA test: https://jira.mitre.org/browse/BONNIE-891


lukoktonos (approved):
JIRA and Regression test passed. Looks good.

alexanderelliott121 (approved):

PR #801, opened 2017-07-21
Replaced usage of Numner.toFixed() with Math.floor() 

hossenlopp:
for coverage ratio percentage calculation


holmesie (approved):
Code change looks good

hossenlopp:
Here is a PR test. https://jira.mitre.org/browse/BONNIE-893

lukoktonos (approved):
JIRA test passes.  Regression test script also passes.  Looks good.

PR #802, opened 2017-07-24
Prevent finalize dialog from looping upon error

lukoktonos:
JIRA Test: https://jira.mitre.org/browse/BONNIE-899

Regression test passed.

### DIFF
--- a/app/controllers/measures_controller.rb
+++ b/app/controllers/measures_controller.rb
@@ -280,16 +280,23 @@ class MeasuresController < ApplicationController
       if is_cql
         measure = CqlMeasure.by_user(current_user).where(hqmf_id: data['hqmf_id']).first
       end
-      measure['needs_finalize'] = false
-      measure.populations.each_with_index do |population, population_index|
-        population['title'] = data['titles']["#{population_index}"] if (data['titles'])
+      begin
+        measure.populations.each_with_index do |population, population_index|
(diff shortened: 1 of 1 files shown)


hossenlopp (approved):
Looks good. I hope the issue seen in the bug report was indeed happening in the begin, rescue block.

I also independently ran the JIRA test. I added some comments in the execution.

c-monkey (approved):
https://jira.mitre.org/browse/BONNIE-899 passed.

app/controllers/measures_controller.rb
```diff
@@ -280,16 +280,23 @@ def finalize
...
+          population['title'] = data['titles']["#{population_index}"] if (data['titles'])
+        end
+        # CQL-based measures don't have episode_ids field
+        unless is_cql
+          measure['episode_ids'] = data['episode_ids']
+          measure.generate_js(clear_db_cache: true)
+        end
+      rescue Exception => e
+        operator_error = true
+        flash[:error] = {title: "Error Loading Measure", summary: "Error Finalizing Measure", body: "An unexpected error occurred while finalizing this measure.  Please delete the measure, re-export the measure from the MAT, and re-upload the measure."}
```
c-monkey: In cql_logic.hbs we use this message:
Please re-package and re-export the measure from the MAT.

Perhaps we should explicitly add that to this error message, too.

PR #803, opened 2017-07-25
Fixed coverage model from not updating after patient delete.

hossenlopp:
Moved the code that grabs the clause_results used for calculation of coverage into the update function so calculations actually use the updated data. Also shortened it to use the .map function on the collection.

Test. https://jira.mitre.org/browse/BONNIE-900


lukoktonos (approved):
JIRA test passed. Regression test passed. Looks good.

c-monkey (approved):
Test https://jira.mitre.org/browse/BONNIE-900 passes.

Code change seems good.

PR #804, opened 2017-07-25
prevent infinite finalize loop if 2 measures have same hqmf_set_id

lukoktonos:
This change checks if there is another measure with the same hqmf_set_id, which should be unique, upon creation. Previously, only CQL measures were checked for a new CQL measure and only QDM measures were checked for a new QDM measure.  Now both CQL and QDM measures are checked in both cases.

JIRA Test: https://jira.mitre.org/browse/BONNIE-906

Regression test script passed.

### DIFF
--- a/app/controllers/measures_controller.rb
+++ b/app/controllers/measures_controller.rb
@@ -79,9 +79,10 @@ class MeasuresController < ApplicationController
 
         measure = Measures::MATLoader.load(params[:measure_file], current_user, measure_details, params[:vsac_username], params[:vsac_password], true, false, effectiveDate, includeDraft, get_ticket_granting_ticket) # Note: overwrite_valuesets=true, cache=false
         existing = CqlMeasure.by_user(current_user).where(hqmf_set_id: measure.hqmf_set_id)
+        qdm_existing = Measure.by_user(current_user).where(hqmf_set_id: measure.hqmf_set_id)
         # Check if there is already a CQL measure with the given hqmf_set_id, this is intentionally different than checking if qdm based is already uploaded (>0 vs >1)
         # TODO: Duplication checks should be more smoothly managed before CQL is fully incorperated
-        if existing.count > 0
+        if existing.count > 0 || qdm_existing.count > 0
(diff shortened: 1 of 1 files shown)


app/controllers/measures_controller.rb
```diff
@@ -79,9 +79,10 @@ def create
 
         measure = Measures::MATLoader.load(params[:measure_file], current_user, measure_details, params[:vsac_username], params[:vsac_password], true, false, effectiveDate, includeDraft, get_ticket_granting_ticket) # Note: overwrite_valuesets=true, cache=false
         existing = CqlMeasure.by_user(current_user).where(hqmf_set_id: measure.hqmf_set_id)
+        qdm_existing = Measure.by_user(current_user).where(hqmf_set_id: measure.hqmf_set_id)
         # Check if there is already a CQL measure with the given hqmf_set_id, this is intentionally different than checking if qdm based is already uploaded (>0 vs >1)
         # TODO: Duplication checks should be more smoothly managed before CQL is fully incorperated
```
hossenlopp: Can this TODO be removed? Or do we need to even more gracefully handle things?

app/controllers/measures_controller.rb
```diff
@@ -291,7 +293,7 @@ def finalize
         end
       rescue Exception => e
         operator_error = true
-        flash[:error] = {title: "Error Loading Measure", summary: "Error Finalizing Measure", body: "An unexpected error occurred while finalizing this measure.  Please delete the measure, re-package and re-export the measure from the MAT, and re-upload the measure."}
+        flash[:error] = {title: "Error Loading Measure", summary: "Error Finalizing Measure", body: "An unexpected error occurred while finalizing this measure.  Please delete the measure, re-export the measure from the MAT, and re-upload the measure."}
```
hossenlopp: Why was the "re-package and" re-export part of this message removed?

hossenlopp (changes requested):

c-monkey (approved):
Test https://jira.mitre.org/browse/BONNIE-906 passes

hossenlopp (approved):

PR #810, opened 2017-08-01
Add XFS Protection

hossenlopp:
Sets the `X-Frame-Options` response header to `DENY` for every request. Adds a framebreaker script. Will blank out the page and attempt to navigate the window to bonnie if it finds it is in a frame.

JIRA test for framebreaker can only be run in development because it needs a code change to actually be capable on a modern browser. https://jira.mitre.org/browse/BONNIE-947

JIRA test for header changes PR and regression. https://jira.mitre.org/browse/BONNIE-946

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,12 @@ class ApplicationController < ActionController::Base
   before_filter :log_additional_data
   layout :layout_by_resource
 
+  after_action :allow_no_iframe
+
+  def allow_no_iframe
+    response.headers['X-Frame-Options'] = 'DENY'
+  end
+
   def layout_by_resource
     if devise_controller?
       "devise"
--- a/app/views/home/show.html.erb
+++ b/app/views/home/show.html.erb
@@ -8,6 +8,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <%= stylesheet_link_tag "application", :media => "all" %>
     <%= render 'layouts/shim' %>
+    <%= render 'layouts/framebreaker' %>
   </head>
   <body>
     <div class="landing-masthead">
--- a/app/views/layouts/_framebreaker.html.erb
+++ b/app/views/layouts/_framebreaker.html.erb
@@ -1,0 +1,9 @@
+<style id="antiClickjack">body{display:none !important;}</style>
+<script type="text/javascript">
+   if (self === top) {
+       var antiClickjack = document.getElementById("antiClickjack");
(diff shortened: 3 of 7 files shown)


hossenlopp:
JIRA tests added original description.

lukoktonos (approved):
JIRA tests both passed.  Code looks good.

app/controllers/application_controller.rb
```diff
@@ -4,6 +4,12 @@ class ApplicationController < ActionController::Base
   before_filter :log_additional_data
   layout :layout_by_resource
 
+  after_action :allow_no_iframe
+
+  def allow_no_iframe
+    response.headers['X-Frame-Options'] = 'DENY'
```
holmesie, line 10: Could this be added to the reverse proxy, like this [guy](https://gitlab.mitre.org/bonnie/internal-documentation/wikis/run-book#resolution), rather than the application controller?

hossenlopp: We needed the reverse proxy to change the request header in that situation so we know it's https that's being reverse proxied for us. X-Frame-Options is a response header. The application should control the response headers. There are sometimes valid reasons to allow a page to load with a different value.

hossenlopp: Also, it looks like another Apache module would need to be installed to support that.

c-monkey (approved):
Tested locally, both pass:
https://jira.mitre.org/browse/BONNIE-946
https://jira.mitre.org/browse/BONNIE-947

Bonnie's functionality seems unchanged on each of these pages.

app/controllers/application_controller.rb
holmesie: mod_headers?

hossenlopp: Yeah.

holmesie: Yeah, that's what I figured. Just was curious, I'm not a reviewer on this anyways. What you are doing obviously works, but wearing my "sys admin" hat, I'd consider passenger/apache as "part of our application", and vulnerabilities like this feel like they should be handled at that level, by the sys admins.

hossenlopp: But our application should decide when it is allowed to be put in a frame. We obviously don't want it to be allowed for the login page. But, once a session with the user is established we may want to allow it. For example we could let the MAT embed a view of tests results in the future. In that case we would want to control the header in our application.

Additionally, having the reverse proxy server, which is shared between apps, handle an app specific thing doesn't make much sense to me.

PR #812, opened 2017-08-07
Bonnie 871 cql fixture export

alexanderelliott121:
Export tool for cql measures, frontend and backend, example test.


edeyoung:
Please resolve failed travis test.

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,254 @@
+# NOTE: Tasks use array arguments to execute.
+# In order for z-shell to execute, noglob is required.
+# e.g., noglob bundle exec rake bonnie:fixtures:load_backend_fixtures[test/fake]
```
mulcahyk, line 3: What is z-shell and what is it used/needed for?

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,254 @@
...
+namespace :bonnie do
+  namespace :fixtures do
+    ###
+    # Generates a set of front end fixtures representing a specific database state.
+    #
+    # cms_hqmf: indicates if a CMS id or an HQMF id is used.
+    #   values: cms, hqmf
+    # path: Path to fixture files, derived from the data-type directories (EX: measure_data/${path}).
+    # user_email: email of user to export.  Measure and patients exported will be taken from this user account.
+    # measure_id: id of measuer to export, taken from account of given user.
```
mulcahyk: typo: measuer

holmesie: sounds french

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,254 @@
...
+        oid_to_vs_map[vs.oid] = vs
+      end
+      
+      value_sets_file = File.join(fixtures_path, 'measure_data', args[:path], 'value_sets.json')
+      create_fixture_file(value_sets_file, JSON.pretty_generate(JSON.parse(oid_to_vs_map.to_json)))
+      puts 'exported value sets to ' + value_sets_file
+
+      #Exports patient data
+      records = Record.by_user_and_hqmf_set_id(user, measure.hqmf_set_id)
+      if (!args[:patient_first_name].nil? && !args[:patient_last_name].nil?)
```
mulcahyk, line 41: Add something to the comments for this task about the usage of **pateint_first_name** and **patient_last_name**. Like when/why you might use them.

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,254 @@
...
+            filtered_records << record
+          end
+        end
+        records = filtered_records
+      end
+      record_file = File.join(fixtures_path, 'records', args[:path], "patients.json")
+      create_fixture_file(record_file, JSON.pretty_generate(JSON.parse(records.to_json)))
+      puts 'exported patient records to ' + record_file
+
+#Commented out for when measure history is fully implemented for CQL
```
mulcahyk: Either remove this or replace it with a TODO.  We don't even know yet what we are going to be storing, or how, for measure history.

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,254 @@
...
+    # user_email: email of user to export.  Measure and patients exported will be taken from this user account.
+    # measure_id: id of measuer to export, taken from account of given user.
+    #
+    # e.g., bundle exec rake bonnie:fixtures:generate_backend_fixtures[cms,test/fake,bonnie@test.org,CMS68v5]
+    desc "Exports a set of fixtures that can be loaded for testing purposes"
+    task :generate_backend_cql_fixtures, [:cms_hqmf, :path, :user_email, :measure_id] => [:environment] do |t, args|
+      fixtures_path = File.join('test', 'fixtures')
+      #Exports the user
+      user = User.find_by email: args[:user_email]
+      bonnie_user_id = '501fdba3044a111b98000001'
```
mulcahyk: I like to have explanations for _magic numbers/strings_.

lib/tasks/generate_cql_fixtures.rake
alexanderelliott121: If I recall correctly, it is something Lizzie uses in her dev environment.

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,218 @@
...
+
+    def get_cql_measure(user, cms_hqmf, measure_id)
+      CqlMeasure.by_user(user).each do |measure|
+        if (cms_hqmf.downcase  == 'cms' && measure.cms_id == measure_id)
+          return measure
+        elsif (cms_hqmf.downcase == 'hqmf' && measure.hqmf_id == measure_id)
+          return measure
+        end
+      end
+      throw('Argument: "' + cms_hqmf + '" does not match expected: cms or hqmf')
```
mulcahyk: This throw is always going to be called.  Consider testing **cms_hqmf** as the first step in the method.

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,218 @@
...
+      records_collection = File.join 'records', args[:path]
+#      upload_summaries_collection = File.join 'upload_summaries', args[:path]
+      users_collection = File.join 'users', 'export_user'
+      collection_fixtures(measure_collection, cql_measure_collection, value_sets_collection, records_collection, users_collection)
+    end
+
+    def get_cql_measure(user, cms_hqmf, measure_id)
+      CqlMeasure.by_user(user).each do |measure|
+        if (cms_hqmf.downcase  == 'cms' && measure.cms_id == measure_id)
+          return measure
```
mulcahyk: You need a **measure.first** here otherwise you are going to get back a Mongoid object and while that object is just one record it isn't what you are expecting elsewhere in the code.

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,218 @@
...
+        elsif (cms_hqmf.downcase == 'hqmf' && measure.hqmf_id == measure_id)
+          return measure
```
mulcahyk, line 149: See above

lib/tasks/generate_cql_fixtures.rake
edeyoung: yeah, because this is a rake task, we should support different terminals. I happen to use z-shell. Others maybe, too?

lib/tasks/generate_cql_fixtures.rake
alexanderelliott121: It only gets thrown if the return in the loop (line 148) doesn't get hit.

lib/tasks/generate_cql_fixtures.rake
alexanderelliott121: by_user returns a list of measures rather than mongoid object.

lib/tasks/generate_cql_fixtures.rake
alexanderelliott121: Added to the method comments

holmesie (approved):
Looks good to me. My only gripe (although I don't feel it is big enough to reject this pull request) is the def statements in the rake file. I'd prefer to see those moved to a module/class in lib, maybe something called `fixture_helpers.rb` or something. Food for thought.

lib/tasks/generate_cql_fixtures.rake
mulcahyk: Maybe update this message.  You can also end up here if you either fat-fingered the value of _measure_id_ or there isn't a measure for that user.

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -0,0 +1,218 @@
...
+    #
+    # e.g., bundle exec rake bonnie:fixtures:load_backend_fixtures[test/fake]
+    desc "Loads set of fixtures into a running instance of BONNIE"
+    task :load_backend_fixtures, [:path] => [:environment] do |t, args|
+#      archived_measures_collection = File.join 'archived_measures', args[:path]
+      measure_collection = File.join 'draft_measures', args[:path]
+      cql_measure_collection = File.join 'cql_measures', args[:path]
+      value_sets_collection = File.join 'health_data_standards_svs_value_sets', args[:path]
+      records_collection = File.join 'records', args[:path]
+#      upload_summaries_collection = File.join 'upload_summaries', args[:path]
```
mulcahyk: Minor - remove reference to measure history, here and a few lines up.

mulcahyk (approved):
Runs great locally.

PR #813, opened 2017-08-07
Fix let keyword and identifiers highlighting

hossenlopp:
Fix for let keyword and identifiers highlighting.

Keyword has no localid for it. So if the parent localId is hit (i.e. TRUE or FALSE) it will highlight green. It does not always highlight for coverage. This can be open to discussion.

Identifiers now point to the localId for the expression that defines them.

JIRA Task: https://jira.mitre.org/browse/BONNIE-948
JIRA Test: https://jira.mitre.org/browse/BONNIE-963

### DIFF
--- a/app/assets/javascripts/helpers/cql_measure_helpers.js.coffee
+++ b/app/assets/javascripts/helpers/cql_measure_helpers.js.coffee
@@ -85,6 +85,14 @@ class CQLMeasureHelpers
       else if k == 'sort'
         # Sort is a special case that we need to recurse into separately and set the results to the result of the statement the sort clause is in
         @_findAllLocalIdsInSort(v, libraryName, localIds, aliasMap, emptyResultClauses, parentNode)
+      else if k == 'let'
+        # let is a special case where it is an array, one for each defined alias. These aliases work slightly different
+        # and execution engine does return results for them on use. The Initial naming of them needs to be properly pointed
+        # to what they are set to.
+        for aLet in v
+          # Add the localId for the definition of this let to it's source.
+          localIds[aLet.localId] = { localId: aLet.localId, sourceLocalId: aLet.expression.localId }
+          @_findAllLocalIdsInStatement(aLet.expression, libraryName, localIds, aliasMap, emptyResultClauses, statement)
       # If 'First' and 'Last' expressions, the result of source of the clause should be set to the expression
       else if k=='type' && (v =='First' || v == 'Last')
         if statement.source && statement.source.localId?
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/helpers/cql_measure_helpers.js.coffee
```diff
@@ -85,6 +85,14 @@ class CQLMeasureHelpers
...
         # Sort is a special case that we need to recurse into separately and set the results to the result of the statement the sort clause is in
         @_findAllLocalIdsInSort(v, libraryName, localIds, aliasMap, emptyResultClauses, parentNode)
+      else if k == 'let'
+        # let is a special case where it is an array, one for each defined alias. These aliases work slightly different
+        # and execution engine does return results for them on use. The Initial naming of them needs to be properly pointed
+        # to what they are set to.
+        for aLet in v
+          # Add the localId for the definition of this let to it's source.
+          localIds[aLet.localId] = { localId: aLet.localId, sourceLocalId: aLet.expression.localId }
+          @_findAllLocalIdsInStatement(aLet.expression, libraryName, localIds, aliasMap, emptyResultClauses, statement)
```
jbradl11, line 95: This looks ok to me. Not that it matters right now but, what made you guys decide to do this on the frontend instead of in Bonnie Bundler?

hossenlopp: This function being in the front end allows us to be good bit more agile in how we handle displaying the logic. If this was in bonnie_bundler and we found an issue in the source of what it uses to highlight a clause, then a rebuild_elm would be needed. It would also immensely slow down development.

It felt like a display related calculation, so I think it belongs in the front end.

mayerm94: I think it was so that if something changes in the way that ELM is represented we won't have to do a measure reupload to get highlighting to cooperate

jbradl11: I agree that it probably allows you to be more agile. Would "findAllLocalIdsInStatementByName" really change?

jbradl11:
You say "It doesn't always highlight for coverage" when will it highlight and when will it not?

mayerm94 (approved):
JIRA test worked

hossenlopp:
Since the let keyword doesn't have its own localId. It cannot be individually counted for coverage or results. I see this as a problem with the translation library. But maybe because `let` doesn't have any result since it's just defining alias(es) it didn't get its own localId.

The `let` keyword in the measure in the JIRA test is part of the clause that determines the result for the entire define statement. So before this PR, it would highlight when the `where` lower in the statement is satisfied. So to properly get the let covered, you should create a test that satisfies that entire statement.

jbradl11 (approved):
Test passes. Code looks good.

PR #815, opened 2017-08-10
Cql4bonnie qdm5 3 937 facility

lukoktonos:
JIRA Test: https://jira.mitre.org/browse/BONNIE-968

Related PRs:
- https://github.com/projecttacoma/bonnie/pull/815
- https://github.com/projecttacoma/cql_qdm_patientapi/pull/65
- https://github.com/projectcypress/health-data-standards/pull/528

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/edit_value.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_value.hbs
@@ -47,52 +47,52 @@
   </div>
 {{else}}
   {{#ifCond type '==' 'FAC'}}
-  <div class="row">
-    <div class="col-md-2">
-      <label for="type_coded" class="control-label">Code</label>
+    <div class="row">
+      <div class="col-md-2">
+        <label for="type_coded" class="control-label">Code</label>
+      </div>
+      <div class="col-md-10">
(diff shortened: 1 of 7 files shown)


lukoktonos:
The Travis CI test fails because the corresponding branch in hds is needed.  It passes locally with the gemfile pointing hds and cql_qdm_patientapi to the branch of the same name.

app/assets/javascripts/templates/patient_builder/edit_value.hbs
```diff
@@ -46,58 +46,108 @@
     </div>
   </div>
 {{else}}
+  {{#ifCond type '==' 'FAC'}}
```
mayerm94: this should be en else if. If Cmp (CMP layout)..... else if FAC(Facility layout) ... else... (logic used for Code, TimeStamp... Scalar layouts). Indenting here seems weird too

lukoktonos: I wanted to use an elseif, but that's not supported in hbs, so I had to use an 
```
if
    (normal if logic goes here)
else
    if 
        (normal elseif logic goes here)
    else 
        (normal else logic goes here)
```
in order to achieve that effect.

edeyoung:
Additional work required. Closing for now.

PR #819, opened 2017-08-16
Cql4bonnie qdm5 3 937 facility

lukoktonos:
JIRA Test: https://jira.mitre.org/browse/BONNIE-968

Related PRs:
- https://github.com/projecttacoma/cql_qdm_patientapi/pull/72
- https://github.com/projecttacoma/bonnie/pull/819
- https://github.com/projectcypress/health-data-standards/pull/530

Note: Patients will need to be re-materialized in the db after these changes have been merged.


lukoktonos:
The Travis CI tests fail because this change requires the branch cql4bonnie-QDM5_3_966 in health_data_standards.  Tests should be run locally with the Gemfile pointing to the corresponding branches in hds and cql_qdm_patientapi.

mayerm94 (dismissed):
Changes look good, my only comment is that when you add a facility location that only has an end date, the date is not shown in the field_value.

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -508,20 +528,26 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
-       col.get('values').push @model.toJSON()
-       @values.add col
+    model_key = @model.get('key')
+    if (@model.get('type') == "CMP" ||
+        @model.get('type') == "FAC" ||
+        model_key == 'FACILITY_LOCATION' ||
+        model_key  == 'DIAGNOSIS'   ||
+        model_key  == 'RELATED_TO')
+
+      compare_collection = @values.findWhere(key: @model.get('key'))
```
c-monkey: replace this additional lookup with `model_key`

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -508,20 +528,26 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
...
+      if compare_collection
+        col = compare_collection
+        # We remove the collection and then re add it to trigger the UI to update
+        @values.remove compare_collection
+      if !col
+        # Create a thorax model collection
+        col = new Thorax.Model()
+        col.set('key', @model.get('key'))
```
c-monkey: replace this additional lookup with `model_key`

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -192,7 +192,13 @@
...
-                  {{start_date}} {{start_time}}
+                  {{#if end_date}}
+                    {{start_date}} {{start_time}} - {{end_date}} {{end_time}},
+                    {{title}}
+                  {{else}}
+                    {{start_date}} {{start_time}} - Undefined, 
+                    {{title}}
+                  {{/if}}
                 {{else}}
                   {{value}} {{unit}}
```
c-monkey: Perhaps this is undesired behavior: if there is no start_date, the end_date is not shown. @losborne and I discussed and think we don't need to enforce anything here but do need to update the display for the case where a user accidentally enters end_date instead of start_date but that isn't visible in the display.

lib/measures/patient_builder.rb
```diff
@@ -253,6 +253,11 @@ def self.recursive_field_value_derivation(value, value_sets, name, entry)
         field_value["code"] = Measures::PatientBuilder.select_code(field.code.code_list_id, value_sets)
         field_value["result"] =  {"scalar"=>value["value"], "units"=>value["unit"]}
         # TODO: will have to add code here if range exists
+      elsif field.type == "FAC"
+        field_value["code"] = Measures::PatientBuilder.select_code(field.code_list_id, value_sets)
+        field_value["display"] = field.title
+        field_value["locationPeriodLow"] = value['locationPeriodLow']
```
c-monkey: inconsistent quotes, other places in this function use double-quotes for hash keys of `value`

c-monkey (changes requested):

config/bonnie.yml
```diff
@@ -1,5 +1,5 @@
 defaults: &defaults
-  version: 2.0.0-alpha.13
```
edeyoung: has the underlying cql4bonnie-QDM5_3 been updated yet? Is this just an artifact from an old ancestor? If cql4bonnie-QDM5_3 hasn't been updated yet (latest cql4bonnie merged in) it should be

lukoktonos: cql4bonnie-QDM5_3 is showing as `version: 2.0.0-alpha.16`, so this is probably an artifact from an old ancestor

c-monkey (dismissed):

edeyoung (approved):
approving based on the approval of previous reviews and that the changes since were very minor (gemfile update and version update)

PR #822, opened 2017-08-18
BONNIE-972: Relevant Period for CMD

c-monkey:
Cumulative Medication Duration support requires Relevant Period for "Medication, Order" and "Medication, Dispensed"

JIRA: https://jira.mitre.org/browse/BONNIE-972
TESTS:
- new unit tests for Medication, Ordered and Medication, Dispensed
- manual test for Medication, Dispensed is https://jira.mitre.org/browse/BONNIE-982

### DIFF
--- a/spec/javascripts/models/data_criteria_spec.js.coffee
+++ b/spec/javascripts/models/data_criteria_spec.js.coffee
@@ -1,0 +1,11 @@
+describe "MeasureDataCriteria", ->
+  beforeEach ->
+    # Clear the fixtures cache so that getJSONFixture does not return stale/modified fixtures
+    jasmine.getJSONFixtures().clearCache()
+    @patient = new Thorax.Models.Patient getJSONFixture('patients.json')[1], parse: true
+
+  # changed from Authored to Relevant Period for QDM 5.3
+  it "specifies 'Medication, Ordered' to have a relevant period", ->
+    dataCriteria = @patient.get('source_data_criteria').at(2)
+    expect(dataCriteria.getCriteriaType()).toBe 'medication_ordered'
+    expect(dataCriteria.isPeriodType()).toBe true
--- a/spec/javascripts/models/data_criteria_spec.js.coffee
+++ b/spec/javascripts/models/data_criteria_spec.js.coffee
@@ -2,7 +2,6 @@ describe "MeasureDataCriteria", ->
   beforeEach ->
     # Clear the fixtures cache so that getJSONFixture does not return stale/modified fixtures
     jasmine.getJSONFixtures().clearCache()
-#    @cqlMeasures = new Thorax.Models.Measure getJSONFixture('measure_data/CQL/CMS136/CMS741v0.json'), parse: true
 
   # changed from Authored to Relevant Period for QDM 5.3
   it "specifies 'Medication, Dispensed' to have a relevant period", ->
--- a/spec/javascripts/models/data_criteria_spec.js.coffee
+++ b/spec/javascripts/models/data_criteria_spec.js.coffee
@@ -2,10 +2,20 @@ describe "MeasureDataCriteria", ->
   beforeEach ->
     # Clear the fixtures cache so that getJSONFixture does not return stale/modified fixtures
     jasmine.getJSONFixtures().clearCache()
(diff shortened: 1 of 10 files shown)


spec/javascripts/models/data_criteria_spec.js.coffee
```diff
@@ -0,0 +1,11 @@
...
+  beforeEach ->
+    # Clear the fixtures cache so that getJSONFixture does not return stale/modified fixtures
+    jasmine.getJSONFixtures().clearCache()
+    @patient = new Thorax.Models.Patient getJSONFixture('patients.json')[1], parse: true
+
+  # changed from Authored to Relevant Period for QDM 5.3
+  it "specifies 'Medication, Ordered' to have a relevant period", ->
+    dataCriteria = @patient.get('source_data_criteria').at(2)
+    expect(dataCriteria.getCriteriaType()).toBe 'medication_ordered'
+    expect(dataCriteria.isPeriodType()).toBe true
```
edeyoung: were you going to create a new fixture with medication dispensed?

Is this based off of an old QDM-based fixture? If so, should we create an updated CQL-based fixture?

Do these changes require fixtures in order to be tested, or could you just create a "new" object of the correct type and check that?

spec/javascripts/views/patient_builder_spec.js.coffee
```diff
@@ -85,8 +85,7 @@ describe 'PatientBuilderView', ->
     it "acquires the dates of the drop target when dropping on an existing criteria", ->
       startDate = @patientBuilder.model.get('source_data_criteria').first().get('start_date')
       endDate = @patientBuilder.model.get('source_data_criteria').first().get('end_date')
-      @patientBuilder.model.get('source_data_criteria').first().set end_date: endDate
-      @addEncounter 1, '.criteria-data.droppable:first'
+      @addEncounter 5, '.criteria-data.droppable:first'
```
mayerm94: Why did you have to change the position of the encounter fro 1 to 5?

c-monkey: The dragged data criteria was changed in MeasureDataCriteria.isPeriodType() and thus no longer had an end date

mayerm94: Why does the fact that the data criteria no longer has an end date change the position of the data criteria that you are adding?

spec/javascripts/models/data_criteria_spec.js.coffee
c-monkey: Yes this is testing a QDM-based measure. I wasn't planning at this time to make a new fixture but that would be reasonable. I was concerned about making a trivial fixture just to test this. I don't know the impact of proliferating a lot of fixtures - how are they organized? Indexed? I figured until a decision was made, making a manual test would suffice.

spec/javascripts/views/patient_builder_spec.js.coffee
c-monkey: I don't know. I just poked at the data in the debugger until I found a criteria that matched. Is the original test fragile? It hasn't changed since Pete originally wrote it.

Perhaps this risk would be best captured by introducing another test that shows a data type that isn't a period, showing there is no end date?

c-monkey: @mayerm94 and I discussed this. We think making a new test showing dragging a data criteria that isPeriodType() onto one with only an authored date

spec/javascripts/models/data_criteria_spec.js.coffee
c-monkey: Are we happy with these 2 new fixtures I created? If so, then I believe I've addressed this feedback.

spec/javascripts/views/patient_builder_spec.js.coffee
c-monkey: I believe I've addressed this question by adding a comment. We would have to massively restructure this test to load another measure to have one that didn't have an end date for dropping purposes.

jbradl11 (approved):
Tested locally. Looks good.

mayerm94 (approved):
PR comments have been addressed. Looks good

PR #823, opened 2017-08-21
Change export script to use hqmf_set_id

alexanderelliott121:
Changes the export script to use the hqmf_set_id

### DIFF
--- a/lib/tasks/generate_cql_fixtures.rake
+++ b/lib/tasks/generate_cql_fixtures.rake
@@ -145,7 +145,7 @@ namespace :bonnie do
       CqlMeasure.by_user(user).each do |measure|
         if (cms_hqmf.downcase  == 'cms' && measure.cms_id == measure_id)
           return measure
-        elsif (cms_hqmf.downcase == 'hqmf' && measure.hqmf_id == measure_id)
+        elsif (cms_hqmf.downcase == 'hqmf' && measure.hqmf_set_id == measure_id)
           return measure
         end
       end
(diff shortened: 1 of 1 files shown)


edeyoung (dismissed):

mulcahyk (commented):
Consider reworded this error message.  "measure_id" for me in this context is ambiguous.  There are four possible identifiers for "measure_id": the measure_id in the database, cms is, hqmf id, and the hqmf set id.  Since the message is dynamic you could say "the CMS ID" or the "HQMF set ID" doesn't match.

lib/tasks/generate_cql_fixtures.rake
```diff
@@ -145,7 +145,7 @@ namespace :bonnie do
       CqlMeasure.by_user(user).each do |measure|
         if (cms_hqmf.downcase  == 'cms' && measure.cms_id == measure_id)
```
mulcahyk: Mongoid is case-sensitive.  Consider making this equality and the following one case-insensitive; CMS113**v**6 does not match CMS113**V**6.  Or if you don't want to do that update the error message to let the user know that they may want to double check the case of the text in their arguments.

-- This comment applies to every instance where you are relying on user input.

alexanderelliott121:
comparison set to use downcase, error message updated.

mulcahyk:
The changes look good to me.  Tested and they function as expected.

mulcahyk (approved):

edeyoung (approved):

PR #824, opened 2017-08-21
BONNIE-967 Implement excel export.

alexanderelliott121:
Exports a cql measure.

To test, load a cql measure into bonnie and hit the excel export button.


app/assets/javascripts/templates/measure.hbs
```diff
@@ -56,8 +56,10 @@
       <span class="settings-container">
         <div class="patients-settings">
           {{#button "exportQrdaPatients" class="btn btn-default export-patients"}}<i class="fa fa-download" aria-hidden="true"></i> QRDA <span class="sr-only">Patient Export</span>{{/button}}
-          {{#unless cql}}
+          {{#if cql}}
```
jbradl11, line 59: Just for clarification, are we not supporting excel export for QDM+HQMF measures anymore?

edeyoung: that's correct.

lib/patient_export.rb
```diff
@@ -64,172 +52,101 @@ def self.export_excel_file(measure, records, results)
...
-          
-            population_criteria.each do |pc|
-              if criteria_keys_by_population[population[pc]].length > 0
-                toplevel_headings[previous_length] = pc
-                heading_positions[pc] = previous_length + 1
-                previous_length += criteria_keys_by_population[population[pc]].length
+            
+
+            statement_to_column = {}
+            header_row = DISPLAYED_ATTRIBUTES + population_criteria*2
```
mayerm94: Why is population_criteria multiplied by two? (Also, add spaces around operator)

app/assets/javascripts/helpers/cql_measure_helpers.js.coffee
```diff
@@ -4,6 +4,27 @@
 ###
 class CQLMeasureHelpers
 
+  @buildDefineToFullStatement: (measure) ->
+    ret = {}
```
jbradl11: Do we not already have a structure that links the Define Statement name to the actual statement?  Or is that not what this is doing? Either way I would recommend comments for this function and the function below this.

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -115,18 +115,61 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
...
-      differences = []
-      @model.get('populations').each (population) ->
-        differences.push(_(population.calculationResults().toJSON()).extend(population.coverage().toJSON()))
-
-      $.fileDownload "patients/excel_export?hqmf_set_id=#{@model.get('hqmf_set_id')}",
-        successCallback: => @exportPatientsView.excelSuccess()
-        failCallback: => @exportPatientsView.fail()
-        httpMethod: "POST"
-        data: {authenticity_token: $("meta[name='csrf-token']").attr('content'), results: differences }
+    calc_results = {}
```
jbradl11: It looks like the exportPatientsView.exporting() loading screen was removed. I think we should add this back in.

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -115,18 +115,61 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
...
+    patient_details = {}
+    population_details = {}
+    statement_details = CQLMeasureHelpers.buildDefineToFullStatement(@model)
+    file_name = @model.get('cms_id')
+    
+    for pop in @model.get('populations').models
```
jbradl11: Could this for loop be moved to the backend? Instead of passing the patients from the frontend you should be able to grab the patients and the measure on the backend and do the processing there.

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -115,18 +115,61 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
...
+      data: {
+        authenticity_token: $("meta[name='csrf-token']").attr('content')
+        calc_results: JSON.stringify(calc_results)
+        patient_details: JSON.stringify(patient_details)
+        population_details: JSON.stringify(population_details)
+        statement_details: JSON.stringify(statement_details)
+        file_name: file_name
+      }
+
+  removeRaw: (results) ->
```
jbradl11: Please add comments to each function.

app/controllers/patients_controller.rb
```diff
@@ -111,19 +111,12 @@ def qrda_export
...
-    
     # Only generate excel document if there are patients for the given measure.
-    if records.length > 0
-      cookies[:fileDownload] = "true" # We need to set this cookie for jquery.fileDownload
-      package = PatientExport.export_excel_file(measure, records, params[:results])
-      send_data package.to_stream.read, type: "application/xlsx", filename: "#{measure.cms_id}.xlsx"
-    else
-      render nothing: true
-    end 
+    cookies[:fileDownload] = "true" # We need to set this cookie for jquery.fileDownload
```
jbradl11, line 114: Looks like you need to remove the left over comment. Also the export shouldn't fail if there are 0 patients.

lib/patient_export.rb
```diff
@@ -64,172 +52,101 @@ def self.export_excel_file(measure, records, results)
...
-          population_criteria = HQMF::PopulationCriteria::ALL_POPULATION_CODES & population.keys
-          # Find the data criteria keys that are associated with this particular population
-          population_criteria_keys = []
-          population_criteria.each do |pc|
-            population_criteria_keys.concat(criteria_keys_by_population[population[pc]])
-          end
-
-          # Sheet name cannot be longer than 31 characters long. Replace with generic name if too long or not present.
-          worksheet_title = if population['title'].blank? || "#{population['title']} Patients".length > 31
-                              "Population #{population_index+1}"
```
jbradl11: Why was this check for character length removed? I can't remember but I am fairly certain this can cause an error.

jbradl11 (changes requested):
I am seeing formatting discrepancies between the original excel export and the CQL excel export. I am also getting a few bugs when testing.

test/functional/patients_controller_test.rb
```diff
@@ -207,17 +207,16 @@ class PatientsControllerTest  < ActionController::TestCase
...
..."Initial Population":"TRUE","Measure Population":"UNHIT","Measure Population Exclusion":"TRUE","RelatedEDVisit":"UNHIT","MeasureObservation":"UNHIT"}},"criteria":{"STRAT":1,"IPP":1,"MSRPOPL":1,"values":[{"low":null,"high":null}],"MSRPOPLEX":1}}},"c12":{"c20":{"statement_results":{"MedianTimefromEDArrivaltoEDDepartureforAdmittedEDPatients":{"Patient":"NA","SDE Ethnicity":"NA","SDE Payer":"NA","SDE Race":"NA","SDE Sex":"NA","Inpatient Encounter":"TRUE","Startification1":"NA","Stratification2":"TRUE","Initial Population":"FALSE","Measure Population":"UNHIT","Measure Population Exclusion":"UNHIT","RelatedEDVisit":"UNHIT","MeasureObservation":"UNHIT"}},"criteria":{"STRAT":1,"IPP":0,"MSRPOPL":0,"values":0,"MSRPOPLEX":0}},"c39":{"statement_results":{"MedianTimefromEDArrivaltoEDDepartureforAdmittedEDPatients":{"Patient":"NA","SDE Ethnicity":"NA","SDE Payer":"NA","SDE Race":"NA","SDE Sex":"NA","Inpatient Encounter":"TRUE","Startification1":"NA","Stratification2":"FALSE","Initial Population":"UNHIT","Measure Population":"UNHIT","Measure Population Exclusion":"UNHIT","RelatedEDVisit":"UNHIT","MeasureObservation":"UNHIT"}},"criteria":{"STRAT":0,"IPP":0,"MSRPOPL":0,"values":0,"MSRPOPLEX":0}}}}'
```
mayerm94: These results are gigantic, you might want to store them as a fixture

app/controllers/patients_controller.rb
```diff
@@ -111,19 +111,12 @@ def qrda_export
   end
 
   def excel_export
-
-    # Grab all records for the given measure
```
mayerm94, line 115: I don't think you should have removed this logic that grabs the records for the given measure, right now when you do an Excel export on a measure, the export contains the records from ALL measures

mayerm94 (changes requested):

test/functional/patients_controller_test.rb
mulcahyk: You don't need to include **calc_results** in the Excel export.  That field is there for caching reasons.

lib/patient_export.rb
alexanderelliott121: Pop criteria is doubled to represent actual vs. expected, will add spaces.

app/assets/javascripts/helpers/cql_measure_helpers.js.coffee
alexanderelliott121: The structure linking define name to full statement was removed when we went to granular highlighting.  I don't think we have the full define statements available at the point where we're building the call to the back end.

Adding comments.

app/assets/javascripts/views/measure_view.js.coffee
alexanderelliott121: Restoring.

test/functional/patients_controller_test.rb
alexanderelliott121: Moving them to fixtures is something I'll try.

@mulcahyk I'm not sure I understand what you mean.  The calc_results variable is the core of the excel export.

app/assets/javascripts/views/measure_view.js.coffee
alexanderelliott121: The loop actually needs to run the calculations, so it has to happen on the front end.
Since all the data is there, I don't see the value of doing the lookups again on the back end.

I can understand the desire to reduce the need for front-end/back-end agreement that back-end lookup would allow, but I like the idea of the flow to the excel export being largely isolated.  If a future dev needs to change what is exported or change the way records, populations and/or measures are implemented, it reduces the places that need to be changed to the front end.

app/controllers/patients_controller.rb
alexanderelliott121: Fix implemented in front end.

lib/patient_export.rb
alexanderelliott121: Check restored.

test/functional/patients_controller_test.rb
mulcahyk: I was thrown off by the name.  I was thinking about the calc_results we used with the QDM logic.  You can ignore my comment.

mayerm94 (changes requested):
Excel exports work for as expected for most measures, but I am getting an error when I upload this file, create a patient and try to export as excel. The error is ArgumentError (There is already a worksheet in this "workbook named 'Population Criteria Section'. Please use a unique name):"
[IETCQL_v5_0_Artifacts.zip](https://github.com/projecttacoma/bonnie/files/1249895/IETCQL_v5_0_Artifacts.zip)

jbradl11:
@mayerm94 Without looking at the zip you added, I suspect that error is caused by having multiple populations on a measure and naming each "tab" in the excel export the same name.  @alexanderelliott121  I recommend you name each "tab" the name of the given Population. ( I believe this is how I did it in the previous version).

app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -116,17 +116,65 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
...
-        data: {authenticity_token: $("meta[name='csrf-token']").attr('content'), results: differences }
+    
+    calc_results = {}
+    patient_details = {}
+    population_details = {}
+    statement_details = CQLMeasureHelpers.buildDefineToFullStatement(@model)
+    file_name = @model.get('cms_id')
+    # Loop iterates over the populations and gets the calculations for each population.
+    # From this it builds a map of pop_key->patient_key->results
+    for pop in @model.get('populations').models
```
jbradl11, line 127: Did you test the performance of this code on a large measure with lots of patients?

alexanderelliott121: Yes, worked well

alexanderelliott121:
@mayerm94 @jbradl11 
The measure has two populations with the same name.  I'm going to start incrementing the names.

jbradl11 (approved):
Tested locally, format looks good. Multiple populations work. Multiple libraries are supported. Measures with no patients export with message in export (eventually I think we should allow them to export the document with the headers populated, even if there are no patients). Approved.

mayerm94 (changes requested):
Everything looks right.

mayerm94 (approved):

alexanderelliott121:
JIRA Test added https://jira.mitre.org/browse/BONNIE-1003

c-monkey (approved):
https://jira.mitre.org/browse/BONNIE-1003 passed

edeyoung:
Also this JIRA test: https://jira.mitre.org/browse/BONNIE-1005

edeyoung:
all tests pass

PR #825, opened 2017-08-23
Cql4bonnie 549

mayerm94:


mayerm94:
This is a preliminary pull request, many of the tests removed from the measure_controller_test are in the process of being converted to CQL tests. Review should focus on the measure_controller

app/controllers/measures_controller.rb
```diff
@@ -131,7 +116,7 @@ def create
 
         if measure_details['episode_of_care'] && measure.data_criteria.values.select {|d| d['specific_occurrence']}.empty?
           measure.delete
-          flash[:error] = {title: "Error Loading Measure", summary: "An episode of care measure requires at least one specific occurrence for the episode of care.", body: "You have loaded the measure as an episode of care measure.  Episode of care measures require at lease one data element that is a specific occurrence.  Please add a specific occurrence data element to the measure logic."}
+          flash[:error] = {title: "Error Loading Measure1", summary: "An episode of care measure requires at least one specific occurrence for the episode of care.", body: "You have loaded the measure as an episode of care measure.  Episode of care measures require at lease one data element that is a specific occurrence.  Please add a specific occurrence data element to the measure logic."}
```
hossenlopp: Why was a '1' added to Measure? Also, I thought specific occurrences are no longer a thing for CQL measures.

app/controllers/measures_controller.rb
```diff
@@ -144,39 +129,30 @@ def create
...
+      clean_email = File.basename(current_user.email) # Prevent path traversal
+      filename = "#{clean_email}_#{Time.now.strftime('%Y-%m-%dT%H%M%S')}#{extension}"
+
+      operator_error = false # certain types of errors are operator errors and do not need to be emailed out.
+      FileUtils.cp(params[:measure_file].tempfile, File.join(errors_dir, filename))
+      File.chmod(0644, File.join(errors_dir, filename))
+      File.open(File.join(errors_dir, "#{clean_email}_#{Time.now.strftime('%Y-%m-%dT%H%M%S')}.error"), 'w') {|f| f.write(e.to_s + "\n" + e.backtrace.join("\n")) }
+      if e.is_a? Measures::ValueSetException
+        operator_error = true
+        flash[:error] = {title: "Error Loading Measure", summary: "The measure value sets could not be found.", body: "Please re-package the measure in the MAT and make sure &quot;VSAC Value Sets&quot; are included in the package, then re-export the MAT Measure bundle."}
```
hossenlopp: Value sets are no longer being included in the package. This check is probably not relevant.

app/controllers/measures_controller.rb
```diff
@@ -41,19 +41,29 @@ def value_sets
...
         redirect_to "#{root_path}##{params[:redirect_route]}"
         return
     elsif extension == '.zip'
       if !Measures::MATLoader.mat_export?(params[:measure_file])
-        flash[:error] = {title: "Error Uploading Measure", summary: "The uploaded zip file is not a Measure Authoring Tool export.", body: "You have uploaded a zip file that does not appear to be a Measure Authoring Tool zip file. If the zip file contains HQMF XML, please unzip the file and upload the HQMF XML file instead of the zip file. Otherwise, please re-export your measure from the MAT and select the 'eMeasure Package' option"}
+        flash[:error] = {title: "Error Uploading Measure", summary: "The uploaded zip file is not a Measure Authoring Tool export.", body: "You have uploaded a zip file that does not appear to be a Measure Authoring Tool zip file please re-export your measure from the MAT and select the 'eMeasure Package' option"}
```
c-monkey: Should add back in a period and capitalize "Please", or put in a comma.

As for redirection, we have a task on the backlog for redirecting to the new server, probably put in a comment with TODO here but skip on changing the wording because it should be consistent between all the places in the code that we change for redirection.

app/controllers/measures_controller.rb
mayerm94: The 1 that was added was a fat finger addition, good catch. But I think you're right about the specific occurrences, Ill just remove the error logic for this.

hossenlopp: Maybe they still are relevant, just serve a different purpose. @holmesie and/ @jbradl11, have any thoughts on the this?

app/controllers/measures_controller.rb
mayerm94: Good point, I just checked all instances where this exception is thrown and it looks like it makes sense to remove this error.

app/controllers/measures_controller.rb
```diff
@@ -41,19 +41,29 @@ def value_sets
...
      'type'=>params[:measure_type],
      'episode_of_care'=>params[:calculation_type] == 'episode'
     }
-
+    
     extension = File.extname(params[:measure_file].original_filename).downcase if params[:measure_file]
-    if extension && !['.zip', '.xml'].include?(extension)
-        flash[:error] = {title: "Error Loading Measure", summary: "Incorrect Upload Format.", body: "The file you have uploaded does not appear to be a Measure Authoring Tool zip export of a measure or HQMF XML measure file. Please re-export your measure from the MAT and select the 'eMeasure Package' option, or select the correct HQMF XML file."}
+    if extension && !['.zip'].include?(extension)
+        flash[:error] = {title: "Error Loading Measure", summary: "Incorrect Upload Format.", body: "The file you have uploaded does not appear to be a Measure Authoring Tool zip export of a measure Please re-export your measure from the MAT and select the 'eMeasure Package'."}
```
c-monkey: Should add back in a period, or else make "Please" lowercase and put in a comma.

app/controllers/measures_controller.rb
```diff
@@ -41,19 +41,29 @@ def value_sets
...
     measure_details = {
      'type'=>params[:measure_type],
      'episode_of_care'=>params[:calculation_type] == 'episode'
     }
-
+    
     extension = File.extname(params[:measure_file].original_filename).downcase if params[:measure_file]
-    if extension && !['.zip', '.xml'].include?(extension)
-        flash[:error] = {title: "Error Loading Measure", summary: "Incorrect Upload Format.", body: "The file you have uploaded does not appear to be a Measure Authoring Tool zip export of a measure or HQMF XML measure file. Please re-export your measure from the MAT and select the 'eMeasure Package' option, or select the correct HQMF XML file."}
+    if extension && !['.zip'].include?(extension)
```
c-monkey: I think we should slightly change the logic of this if/else block. The first line should be:
`if !extension || extension != '.zip'`
and the `elsif` changed to a simple `else`

app/controllers/measures_controller.rb
holmesie: No, @hossenlopp you are correct, we don't care about those anymore.

app/controllers/measures_controller.rb
```diff
@@ -41,19 +41,29 @@ def value_sets
...
+        flash[:error] = {title: "Error Loading Measure", summary: "Incorrect Upload Format.", body: "The file you have uploaded does not appear to be a Measure Authoring Tool zip export of a measure Please re-export your measure from the MAT and select the 'eMeasure Package'."}
         redirect_to "#{root_path}##{params[:redirect_route]}"
         return
     elsif extension == '.zip'
       if !Measures::MATLoader.mat_export?(params[:measure_file])
```
c-monkey: Let's change this to be:
`if Measures::HQMFLoader.mat_hqmf_export?(params[:measure_file])`
and error indicating this is a MAT package that is no longer supported.

app/controllers/measures_controller.rb
```diff
@@ -72,49 +82,24 @@ def create
           effectiveDate = Date.strptime(params[:vsac_date],'%m/%d/%Y').strftime('%Y%m%d')
         end
       end
-      # If file extension is a zip and a CQL MAT export
-      is_cql = false
-      if extension == '.zip' && Measures::CqlLoader.mat_cql_export?(params[:measure_file])
-        is_cql = true
 
+      if extension == '.zip' && Measures::CqlLoader.mat_cql_export?(params[:measure_file])
```
c-monkey: With my above suggested changes, you now can't get here unless uploaded file is a CQL-based measure. Therefore this if can be removed. You don't handle the `else` situation, but that's okay, it isn't necessary with my changes.

That being said, for clarity, I think above in the line before this `begin` block (_original numbering line_ `61`), we should add a comment, something like:
`# if we get here, uploaded file is a CQL-based measure in .zip format`

app/controllers/measures_controller.rb
```diff
@@ -72,49 +82,24 @@ def create
...
-        existing = CqlMeasure.by_user(current_user).where(hqmf_set_id: measure.hqmf_set_id)
-        qdm_existing = Measure.by_user(current_user).where(hqmf_set_id: measure.hqmf_set_id)
-        # Check if there is already a CQL measure with the given hqmf_set_id, this is intentionally different than checking if qdm based is already uploaded (>0 vs >1)
-        if existing.count > 0 || qdm_existing.count > 0
-          flash[:error] = {title: "Error Loading Measure", summary: "A version of this measure is already loaded.", body: "You have a version of this measure loaded already.  Try deleting that measure and re-uploading it."}
-          redirect_to "#{root_path}##{params[:redirect_route]}"
-          return
-        end
-      else
+        existing = CqlMeasure.by_user(current_user).where(hqmf_set_id: measure.hqmf_set_id).first
```
c-monkey: This value of `existing` is unused. If is_update is true, then `existing` is set by the code on line 91. If false, then the variable `cql_existing` is used.

c-monkey:
- You should re-enable the `updateMeasure` button on app/assets/javascripts/templates/measure.hbs

- When you upgrade, the type field on the upload view is empty. This can be fixed in bonnie_bundler in `lib/measures/loading/cql_loader.rb` in the function `load_mat_cql_exports` after the call to `Measures::Loader.load_hqmf_cql_model_json`. I did so with this one line:
`measure['type'] = measure_details['type']`
This is because `measure_details` is not passed into `load_hqmf_cql_model_json`

c-monkey (changes requested):

app/assets/javascripts/templates/measure.hbs
```diff
@@ -7,10 +7,9 @@
           {{#unless cql}}
             {{#button "toggleVisualization" class="btn btn-default btn-measure-viz btn-viz-chords"}}<i class="fa fa-sitemap" aria-hidden="true"></i> Chords View{{/button}}
             {{#button "toggleVisualization" class="btn btn-default btn-measure-viz btn-viz-text"}}<i class="fa fa-list" aria-hidden="true"></i> Text View{{/button}}
-            {{#button "showCQL" class="btn btn-default"}}Learn CQL{{/button}}
-            {{#button "updateMeasure" class="btn btn-default update-measure"}}<i class="fa fa-upload" aria-hidden="true"></i> Update{{/button}}
           {{/unless}}
-
+          {{#button "showCQL" class="btn btn-default"}}Learn CQL{{/button}}
```
c-monkey: I think this button should be removed altogether. I tried messing around with it and I don't think it is helpful to be able to edit CQL this way anymore.

c-monkey (changes requested):
Minor change requested. Everything else is good.

c-monkey (approved):

holmesie (approved):
Approved

mayerm94:
Test: https://jira.mitre.org/browse/BONNIE-1001

c-monkey (approved):
https://jira.mitre.org/browse/BONNIE-1001 passed

PR #827, opened 2017-08-24
If there is a Result just allow one Result

mulcahyk:
https://jira.mitre.org/browse/BONNIE-853

Align the behavior of Bonnie with the QDM spec (5.02, 5.3) that for certain data criteria there can be up to 1 Result per instance of that criteria.

This change will have the side effect that if a user selects a DateTime that this message will display:
screen shot 2017-08-24 at 6 10 59 pm
(where the measure id will be for whatever measure the user is on.)

### DIFF
--- a/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
+++ b/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
@@ -389,8 +389,7 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
       codes: @measure?.valueSets().map((vs) -> vs.toJSON()) or []
       # QDM say that per instance of a data criteria there can be only 1 Result
       # The function Thorax.Models.PatientDataCriteria.canHaveResult determines which criteria those are
-      onlySingleResultAllowed =  @values.models.length > 0
-      hideEditValueView: onlySingleResultAllowed
+      hideEditValueView: @values.models.length > 0
 
   # When we serialize the form, we want to put the description for any CD codes into the submission
   events:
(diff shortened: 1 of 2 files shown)


c-monkey (dismissed):
tests pass, looks good

One question: for existing patients with multiple results, do we need to do some sort of mongo query and report to the MDs so they can fix them?

edeyoung:
good idea! kristian, can you pull that together?

app/assets/javascripts/views/patient_builder/data_criteria.js.coffee
```diff
@@ -387,7 +387,10 @@ class Thorax.Views.EditCriteriaValueView extends Thorax.Views.BuilderChildView
   context: ->
     _(super).extend
       codes: @measure?.valueSets().map((vs) -> vs.toJSON()) or []
-      hideEditValueView: @criteriaType == 'risk_category_assessments' && @values.models.length > 0
+      # QDM say that per instance of a data criteria there can be only 1 Result
+      # The function Thorax.Models.PatientDataCriteria.canHaveResult determines which criteria those are
+      onlySingleResultAllowed =  @values.models.length > 0
+      hideEditValueView: onlySingleResultAllowed
```
edeyoung: Consider making this just be:

```
hideEditValueView: @values.models.length > 0
```

edeyoung (changes requested):

edeyoung (approved):

edeyoung:
i've approved *code* changes. @c-monkey can you re-run the tests for this to confirm it all still works? Thanks!

c-monkey (approved):
Tests pass,  TestCMS167 (CMS21v0) in the epecqmama@gmail.com works as expected.

PR #830, opened 2017-08-30
updated gemfile lock for v18

edeyoung:


c-monkey (approved):
SHA is correct, matching latest cql_qdm_patientapi.

Ruby and Jasmine tests pass.

PR #832, opened 2017-08-31
Cql4bonnie 1028

mayerm94:
Test with health-data-standards cql4bonnie_1028

### DIFF
--- a/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
@@ -222,12 +222,11 @@
               {{#if start_date}}
                 {{start_date}} {{start_time}}
               {{else}}
-                {{#if value}}
-                  {{value}} {{unit}}
-                  {{title}}
-                {{/if}}
                 {{#if numerator}}
                   {{numerator}}/{{denominator}}
+                  {{title}}
+                {{else}}
+                  {{value}} {{unit}}
                   {{title}}
                 {{/if}}
               {{/if}}
--- a/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
+++ b/app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
@@ -222,8 +222,14 @@
               {{#if start_date}}
                 {{start_date}} {{start_time}}
               {{else}}
-                {{value}} {{unit}}
-                {{title}}
+                {{#if value}}
+                  {{value}} {{unit}}
+                  {{title}}
+                {{/if}}
+                {{#if numerator}}
+                  {{numerator}}/{{denominator}}
+                  {{title}}
+                {{/if}}
               {{/if}}
               {{#if codes}} - Codes:
                 {{#each codes}}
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -222,8 +222,13 @@
               {{#if start_date}}
                 {{start_date}} {{start_time}}
               {{else}}
-                {{value}} {{unit}}
-                {{title}}
+                {{#if numerator}}
+                  {{numerator}}/{{denominator}}
```
lukoktonos: Should there be an #if to make sure there is a denominator?

app/assets/javascripts/templates/patient_builder/edit_criteria.hbs
```diff
@@ -222,8 +222,13 @@
...
+                  {{title}}
+                {{else}}
+                  {{value}} {{unit}}
```
lukoktonos: I think this will add an extra space if there are no units--add an #if block around the space and unit to just have 1 space in that case.

app/assets/javascripts/templates/patient_builder/edit_value.hbs
```diff
@@ -103,6 +103,7 @@
             <option value="PQ" data-icon="fa fa-pencil-square-o">Scalar</option>
             <option value="CD" data-icon="fa fa-list">Coded</option>
             <option value="TS" data-icon="fa fa-calendar">Time</option>
+            <option value="RT" data-icon="fa fa-pencil-square-o">Ratio</option>
```
holmesie: This is using the same icon as `Scalar`. Instead of `fa-pencil-square-o`, why not something like `fa-balance-scale`? Take a look at [this](http://fontawesome.io/cheatsheet/).

holmesie (changes requested):

app/assets/javascripts/templates/patient_builder/edit_value.hbs
hossenlopp: Why not `fa-percent`?

app/assets/javascripts/templates/patient_builder/edit_value.hbs
```diff
@@ -140,6 +141,17 @@
...
       {{/ifCond}}
 
+      {{#ifCond type '==' 'RT'}}
+        <div class="col-md-4">
+          <label for="numerator_{{@cid}}" class="sr-only">numerator</label>
+          <input type="number" name="numerator" id="numerator_{{@cid}}" class="form-control" placeholder="Numerator">
+        </div>
+        <div class="col-md-4">
+          <label for="denominator_{{@cid}}" class="sr-only">denominator</label>
+          <input type="number" name="denominator" id="denominator{{@cid}}" class="form-control" placeholder="Denominator">
```
lukoktonos: id should be `denominator_{{@cid}}` to be consistent with numerator and other ids

lukoktonos (changes requested):
Just have one more suggestion--I think there needs to be a validation added to `bonnie/app/assets/javascripts/views/patient_builder/data_criteria.js.coffee` (validateForAddition, line 497) to validate that both a numerator and a denominator have been filled in before allowing the user to add.

(Note the boolean logic for components is wrong for that currently, so don't copy it. That will be fixed with my BONNIE-453 PR).

I think the correct logic should look like:
```
(attributes.type == 'RT' && (!attributes.denominator || !attributes.numerator)) || ...
```

edeyoung:
RATIO WORK NO LONGER BEING USED AS OF 9/1 DISCUSSION IN HQMF CQL TOOLING CALL

PR #834, opened 2017-08-31
Cql4bonnie 1002

edeyoung:
Enables loading of MAT 5.4  measures as well as previous versions of MAT measures.

Test currently fails because needs to reference cql4bonnie_1002 in HDS and bonnie_bundler.

No JIRA test because automated test takes care of this.


holmesie (dismissed):

c-monkey (dismissed):
tests pass

c-monkey (approved):
tests passed locally

PR #835, opened 2017-09-02
Cql4bonine 1049 expected values

edeyoung:
This uses the code for updated expected value structure from the measure history code.

JIRA: https://jira.mitre.org/browse/BONNIE-1055

### DIFF
--- a/app/controllers/measures_controller.rb
+++ b/app/controllers/measures_controller.rb
@@ -193,6 +193,11 @@ class MeasuresController < ApplicationController
       end
     end
 
+    # ensure expected values on patient match those in the measure's populations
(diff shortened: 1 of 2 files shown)


mulcahyk:
Are there JIRA tests that go along with this?  Do you have a sample measure to work with?

c-monkey:
@mulcahyk Links are not in this github PR, but the referenced internal JIRA ticket has a lot of info, and the referenced external ticket has an attached measure.

mulcahyk (approved):
Performing as expected.

c-monkey (approved):
Verified bug with cql4bonnie branch, switch to this branch and bug fixed.

hossenlopp:
Regression test script passes with no changes.

PR #826, opened 2017-08-24
Cql4bonnie qdm5 3 453

lukoktonos:
JIRA Ticket: https://jira.mitre.org/browse/BONNIE-453

Unit tests included so no JIRA tests.  

UI Demo:
component_ui

NOTE:
`patient_builder_spec.js.coffee` looks like it has a lot of changes, but it is just an indentation change to separate QDM and CQL tests.

PR #836, opened 2017-09-04
Add support for ReferenceRange for LaboratoryTestPerformed, for QDM5.3

lukoktonos:
Also pulled in UI changes from bonnie branch cql4bonnie-QDM5_3_453 needed for visual inspection.
(PR for those changes here: https://github.com/projecttacoma/bonnie/pull/826)

JIRA Ticket: https://jira.mitre.org/browse/BONNIE-1000

*NOTE: For CI tests to pass, need to point health-data-standards and cql_qdm_patientapi to branches cql4bonnie-QDM5_3_1000.*


app/assets/javascripts/templates/patient_builder/edit_value.hbs
```diff
@@ -26,6 +26,34 @@
       </select>
     </div>    
   </div>
+  {{#ifCond this.fullCriteriaType '==' 'laboratory_test_performed'}}
+  <div class="row">
```
mayerm94: Tab in for ifCond block

app/assets/javascripts/templates/patient_builder/edit_value.hbs
```diff
@@ -26,6 +26,34 @@
       </select>
     </div>    
   </div>
+  {{#ifCond this.fullCriteriaType '==' 'laboratory_test_performed'}}
+    <div class="row">
+      <div class="col-md-6">
+        <label for="referenceRangeLow" class="control-label">Reference Range - Low</label>
+      </div> 
+      <div class="col-md-3">
```
mayerm94, line 34: col-md should sum to 12

lukoktonos: I think it looks better as is, because the input fields are vertically aligned.  If I change it to sum to 12, they are no longer aligned:
Sum to 12:
image
Sum to 11:
image

mayerm94 (dismissed):
Changes look good

app/assets/javascripts/templates/patient_builder/edit_value.hbs
holmesie: I don't think that would always be the case (between browsers now and into the future/tablets), these should _really_ add up to 12. Even if you used offsets, or had a blank div, that would be preferable imo.

lib/measures/patient_builder.rb
```diff
@@ -250,9 +250,12 @@ def self.recursive_field_value_derivation(value, value_sets, name, entry)
         end
         field_value["title"] = Measures::PatientBuilder.select_value_sets(field.code_list_id, value_sets)["display_name"] if field_value
       elsif field.type == "CMP"
-        field_value["code"] = Measures::PatientBuilder.select_code(field.code.code_list_id, value_sets)
+        field_value["code"] = Measures::PatientBuilder.select_code(field.code_list_id, value_sets)
         field_value["result"] =  {"scalar"=>value["value"], "units"=>value["unit"]}
-        # TODO: will have to add code here if range exists
+        if value["referenceRangeLow_value"]
```
holmesie, line 255: Likely dumb question, but are we always assuming there will be a `referenceRangeHigh_value` if there is a `referenceRangeLow_value`? Do we care if if the value was ever just a blank string or something? I guess what I'm getting at is maybe a `present?` here couldn't hurt.

holmesie (changes requested):

app/assets/javascripts/templates/patient_builder/edit_value.hbs
lukoktonos: I added a blank div at the end and it looks like it is still aligned vertically, so that should work.

lib/measures/patient_builder.rb
edeyoung: good point - i think you could have a referencehigh without a referenclow
i think these are meant to give you an idea if a lab result is too high or is too low.

lukoktonos: This `if value["referenceRangeLow_value"]` is meant to determine if this "CMP" is coming from a Component or a ResultComponent. It is OK if `value["referenceRangeLow_value"]` or `value["referenceRangeHigh_value"]` are `""`, because in rubyland that evaluates to true.  When the ResultComponent constructor gets called in cql_qdm_patientapi, the `""` evaluates to `false` in the if statement and so the range Interval has an undefined edge for that case.

There is a similar check to see if it is a Component vs ResultComponent in HDS:
```if json['referenceRangeLow_value']```

Do you think a comment clarifying this behavior would be sufficient? Or do you still think there needs to be some other change with the logic?

holmesie: No, things are clearer to me now.

holmesie (dismissed):

edeyoung (approved):

PR #837, opened 2017-09-05
Cql4bonnie value sets 1050

edeyoung:
JIRA: https://jira.mitre.org/browse/BONNIE-1056


lib/tasks/cql.rake
```diff
@@ -22,25 +22,20 @@ namespace :bonnie do
...
             main_cql_library = elms.first['library']['identifier']['id']
           end
-          # Build the definition dependency structure for this measure
-          cql_definition_dependency_structure = Measures::CqlLoader.populate_cql_definition_dependency_structure(main_cql_library, elms)
-          cql_definition_dependency_structure = Measures::CqlLoader.populate_used_library_dependencies(cql_definition_dependency_structure, main_cql_library, elms)
 
-          Measures::CqlLoader.replace_codesystem_oids_with_names(elms)
-          Measures::CqlLoader.modify_value_set_versions(elms)
+          cql_artifacts = Measures::CqlLoader.process_cql(cql, main_cql_library, user)
+          elms = cql_artifacts[:elms].instance_of?(Array) ? cql_artifacts[:elms] : [cql_artifacts[:elms]]
```
hossenlopp: We should no longer need to coalesce a singular elm to an elm array. That was an artifact of switching to library support. We should only encounter arrays.

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -34,11 +34,12 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
...
 
-        if bonnie.valueSetsByOid[oid]?
-          version = bonnie.valueSetsByOid[oid].version
+        oid_version = @model.get('value_set_oid_version_objects').find (oid_version) -> oid_version.oid == oid
+        if oid_version? && bonnie.valueSetsByOid[oid]?
+          version = oid_version.version
           # if it's a date (rather than "Draft"), format the version
-          version = moment(version, "YYYYMMDD").format("MM/DD/YYYY") if moment(version, "YYYYMMDD").isValid()
-          codeConcepts = bonnie.valueSetsByOid[oid].concepts ? []
+          #version = moment(version, "YYYYMMDD").format("MM/DD/YYYY") if moment(version, "YYYYMMDD").isValid()
```
jbradl11: Should this be removed? Or uncommented?

lib/tasks/cql.rake
```diff
@@ -22,25 +22,20 @@ namespace :bonnie do
...
           else
             main_cql_library = elms.first['library']['identifier']['id']
           end
-          # Build the definition dependency structure for this measure
-          cql_definition_dependency_structure = Measures::CqlLoader.populate_cql_definition_dependency_structure(main_cql_library, elms)
-          cql_definition_dependency_structure = Measures::CqlLoader.populate_used_library_dependencies(cql_definition_dependency_structure, main_cql_library, elms)
 
-          Measures::CqlLoader.replace_codesystem_oids_with_names(elms)
-          Measures::CqlLoader.modify_value_set_versions(elms)
+          cql_artifacts = Measures::CqlLoader.process_cql(cql, main_cql_library, user)
```
hossenlopp, line 33: Can this reliably be done without VSAC credentials?

edeyoung: Yes, if VSAC creds aren't added, then it retrieves values from the DB and uses those. Because this is a "rebuild", that information should be present.

hossenlopp: _Should_ be present. I'm just worried at some point we will run into issues if change the way we interpret value set versions differently.  Also, if they uploaded with the option to get draft, how do we know that's what they wanted when  rebuild_elm happens.

edeyoung: we discussed the draft status, and a comment was added

lukoktonos:
Manual regression test pass--all results the same.

jbradl11 (dismissed):
Tests pass. Looks good.

hossenlopp (approved):

PR #839, opened 2017-09-05
BONNIE-976 Based on cql4bonnie changes.

alexanderelliott121:
BONNIE-976, based off of changed cql4browsers.

When testing, please remember to change gemfile to point to BONNIE-976_Profile for bonnie bundler and hds.


c-monkey:
Unit tests pass. Discussed w/Alex, we need to create new test fixture so we can have include_draft be false in a test.

c-monkey:
This PR is related to:
https://github.com/projecttacoma/bonnie_bundler/pull/106
https://github.com/projectcypress/health-data-standards/pull/535

alexanderelliott121:
Test added

c-monkey (dismissed):
Unit tests here, in bonnie_bunder, and in HDS pass.

https://jira.mitre.org/browse/BONNIE-1030 passed.

app/controllers/measures_controller.rb
```diff
@@ -84,20 +84,16 @@ def create
...
         # If the measure is published (includesDraft = false)
         # EffectiveDate is specified to determine a value set version.
         includeDraft = params[:include_draft] == 'true'
-        unless includeDraft
-          effectiveDate = Date.strptime(params[:vsac_date],'%m/%d/%Y').strftime('%Y%m%d')
-        end
       end
 
-      measure = Measures::MATLoader.load(params[:measure_file], current_user, measure_details, params[:vsac_username], params[:vsac_password], false, false, effectiveDate, includeDraft, get_ticket_granting_ticket) # Note: overwrite_valuesets=false, cache=false
+      measure = Measures::MATLoader.load(params[:measure_file], current_user, measure_details, params[:vsac_username], params[:vsac_password], false, false, includeDraft, get_ticket_granting_ticket) # Note: overwrite_valuesets=true, cache=false
```
edeyoung: Looks like this may not have gotten updated properly - overwrite_valuesets = true in the comments but is passed in as false. It should be false.

edeyoung (dismissed):
From Bonnie Bundler PR:

Need to make sure that client-side valueset retrieves appropriately get the correct value set. Currently, I don't believe it will break because `version` is null so it just grabs the latest value set. However, it should actually grab the value set with the correct version. This isn't just for processing, but also for populating the data criteria as well as the value set list. See where the value set changes touched things in the recent value set change commit.

c-monkey:
bundle exec rake teaspoon broken: `.TypeError: undefined is not an object (evaluating 'ref1.length')`

jbradl11:
@c-monkey I am not getting that issue. And I believe that is related to having QDM + HQMF Measures in the account.

c-monkey:
Interesting. So Alex and I *are* getting issues. I logged into my account and deleted the 2 existing measures (both CQL-based and uploaded in the past couple days). To fix another issue I had run the `bonnie:cql:rebuild_elm` rake task, maybe that introduced the issue? Anyway, with no measures in my account the teaspoon ran w/out error.

c-monkey:
@edeyoung @jbradl11 Do we need to change the code added here https://github.com/projecttacoma/bonnie/commit/68a7d6aacb90bcae8bc1e9467969a55dd8371056#diff-13a5266159215058cea443be68fbb24fL57 to check for the existence of the value_set_oid_version_objects?

jbradl11:
What do you mean check for the existence? From now on, this structure will always be present.

c-monkey (approved):
the unit tests in bonnie, bonnie_bundler, and hds pass; https://jira.mitre.org/browse/BONNIE-1030 passes

jbradl11 (dismissed):
Tests pass. Jira Tests pass. Looks good.

c-monkey (dismissed):
As discussed in the Slack thread:

We feel that the JIRA test is sufficient. It doesn’t check what the value sets actually are and ensure they’re populated. It should probably also check to ensure you can build a patient. I’d expect “draft” selected value sets to have a different version than those done with a “profile”

lukoktonos:
Regression script passed -- all results are the same.

alexanderelliott121:
Improved test can be found at https://jira.mitre.org/browse/BONNIE-1067

jbradl11:
I am seeing failures for Unit tests here as well.

holmesie (dismissed):
All tests passing here, looks good.

jbradl11:
I am still getting errors in the unit tests but @alexanderelliott121  and @holmesie  have confirmed they have passing tests.

jbradl11 (dismissed):
All code looks good. Jira Test passes.

edeyoung (approved):
approved changes based on reviews and previous approvals from bradley and holmes

PR #841, opened 2017-09-08
Bonnie 1061

lukoktonos:
JIRA Test:
https://jira.mitre.org/browse/BONNIE-1069

JIRA Task:
https://jira.mitre.org/browse/BONNIE-1061

### DIFF
--- a/lib/tasks/cql.rake
+++ b/lib/tasks/cql.rake
@@ -217,6 +217,12 @@ namespace :bonnie do
                   new_facility_location['values'][0]['code_list_id'] = source_data_criterium['field_values']['FACILITY_LOCATION']['code_list_id']
                   new_facility_location['values'][0]['field_title'] = source_data_criterium['field_values']['FACILITY_LOCATION']['field_title']
                   new_facility_location['values'][0]['title'] = source_data_criterium['field_values']['FACILITY_LOCATION']['title']
+                  if source_data_criterium['field_values']['FACILITY_LOCATION']['value']
+                    new_facility_location['values'][0]['value'] = source_data_criterium['field_values']['FACILITY_LOCATION']['value']
+                  end
+                  if source_data_criterium['field_values']['FACILITY_LOCATION']['end_value']
+                    new_facility_location['values'][0]['end_value'] = source_data_criterium['field_values']['FACILITY_LOCATION']['end_value']
+                  end
 
                   # Convert times
                   converted_start_date = nil
--- a/lib/tasks/cql.rake
+++ b/lib/tasks/cql.rake
@@ -167,6 +167,48 @@ namespace :bonnie do
     task :update_facilities_and_diagnoses => :environment do
       # For any relevant datatypes, update old facilities and diagnoses to be collections with single elements
       Record.all.each do |patient|
+        if patient.source_data_criteria
+          patient.source_data_criteria.each_with_index do |source_data_criterium, index|
+            new_source_data_criterium_field_values = {}
+            if source_data_criterium['field_values']
+              source_data_criterium['field_values'].each do |field_value_key, field_value_value|
+                # update any 'DIAGNOSIS' field values that aren't collections
+                if field_value_key == 'DIAGNOSIS' && !(source_data_criterium['field_values']['DIAGNOSIS']['type'] == 'COL')
+                  new_diagnosis = {}
+                  new_diagnosis['type'] = 'COL'
+                  new_diagnosis['values'] = [{}]
+                  new_diagnosis['values'][0]['type'] = source_data_criterium['field_values']['DIAGNOSIS']['type']
+                  new_diagnosis['values'][0]['key'] = source_data_criterium['field_values']['DIAGNOSIS']['key']
+                  new_diagnosis['values'][0]['code_list_id'] = source_data_criterium['field_values']['DIAGNOSIS']['code_list_id']
+                  new_diagnosis['values'][0]['field_title'] = source_data_criterium['field_values']['DIAGNOSIS']['field_title']
+                  new_diagnosis['values'][0]['title'] = source_data_criterium['field_values']['DIAGNOSIS']['title']
+                  new_source_data_criterium_field_values['DIAGNOSIS'] = new_diagnosis 
+
+                # update any 'FACILITY_LOCATION' field values that aren't collections
+                elsif field_value_key == 'FACILITY_LOCATION' && !(source_data_criterium['field_values']['FACILITY_LOCATION']['type'] == 'COL')
+                  new_facility_location = {}
+                  new_facility_location['type'] = 'COL'
+                  new_facility_location['values'] = [{}]
+                  new_facility_location['values'][0]['type'] = source_data_criterium['field_values']['FACILITY_LOCATION']['type']
+                  new_facility_location['values'][0]['key'] = source_data_criterium['field_values']['FACILITY_LOCATION']['key']
+                  new_facility_location['values'][0]['code_list_id'] = source_data_criterium['field_values']['FACILITY_LOCATION']['code_list_id']
+                  new_facility_location['values'][0]['field_title'] = source_data_criterium['field_values']['FACILITY_LOCATION']['field_title']
+                  new_facility_location['values'][0]['start_date'] = source_data_criterium['field_values']['FACILITY_LOCATION']['start_date']
+                  new_facility_location['values'][0]['start_time'] = source_data_criterium['field_values']['FACILITY_LOCATION']['start_time']
+                  new_facility_location['values'][0]['end_date'] = source_data_criterium['field_values']['FACILITY_LOCATION']['end_date']
+                  new_facility_location['values'][0]['end_time'] = source_data_criterium['field_values']['FACILITY_LOCATION']['end_time']
+                  new_facility_location['values'][0]['title'] = source_data_criterium['field_values']['FACILITY_LOCATION']['title']
+                  new_source_data_criterium_field_values['FACILITY_LOCATION'] = new_facility_location 
+                else
+                  # add unaltered field value to new structure
+                  new_source_data_criterium_field_values[field_value_key] = field_value_value
+                end
+              end
+              source_data_criterium['field_values'] = new_source_data_criterium_field_values
+            end
+          end
+        end
+
         if patient.encounters
           patient.encounters.each do |encounter|
             if encounter['facility'] && !encounter['facility']['type']
@@ -183,7 +225,6 @@ namespace :bonnie do
               new_encounter_diagnosis['values'][0]['title'] = encounter['diagnosis']['title']
               new_encounter_diagnosis['values'][0]['code'] = encounter['diagnosis']['code']
               new_encounter_diagnosis['values'][0]['code_system'] = encounter['diagnosis']['code_system']
-
               encounter['diagnosis'] = new_encounter_diagnosis
             end
           end
--- a/lib/tasks/cql.rake
+++ b/lib/tasks/cql.rake
@@ -114,5 +114,100 @@ namespace :bonnie do
       end
     end
 
+    def update_facility(patient, datatype)
+      if datatype['facility'] && !datatype['facility']['type']
+        print "\e[#{32}m#{"[Facility]"}\e[0m"
+        puts "#{patient.first} #{patient.last}"
+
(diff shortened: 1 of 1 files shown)


lib/tasks/cql.rake
```diff
@@ -114,5 +114,100 @@ namespace :bonnie do
...
+    def update_facility(patient, datatype)
+      if datatype['facility'] && !datatype['facility']['type']
+        print "\e[#{32}m#{"[Facility]"}\e[0m"
+        puts "#{patient.first} #{patient.last}"
+
+        # Need to build new facility and assign it in order to actually save it in DB
+        new_datatype_facility = {}
+
+        # Assign type to be 'COL' for collection
+        new_datatype_facility['type'] = 'COL'
```
mulcahyk, line 139: I understand the need for a `type`.  Is there something other than `COL` that could be used?  It just looks a lot like C**Q**L.

lib/tasks/cql.rake
```diff
@@ -114,5 +114,100 @@ namespace :bonnie do
...
+          code = nil
+        end
+        
+        new_datatype_facility['values'][0]['code'] = {'code_system'=>code_system, 'code'=>code}
+        datatype['facility'] = new_datatype_facility 
+      end
+    end
+
+    task :update_facilities_and_diagnoses => :environment do
+      # Update old facilities and diagnoses to be collections with single elements
```
mulcahyk: Comments needs to be updated. The comment says you are only doing this for facilities and diagnoses but the code is also checking for adverse events and procedures.

lib/tasks/cql.rake
```diff
@@ -114,5 +114,100 @@ namespace :bonnie do
...
+            if encounter['facility'] && !encounter['facility']['type']
+              update_facility(patient, encounter)
+            end
+
+            # Diagnosis is only for encounter
+            if encounter['diagnosis'] && !encounter['diagnosis']['type']
+              print "\e[#{32}m#{"[Encounter Diagnosis]"}\e[0m"
+              puts "#{patient.first} #{patient.last}"
+              new_encounter_diagnosis = {}
+              new_encounter_diagnosis
```
mulcahyk: extra line??

lib/tasks/cql.rake
```diff
@@ -114,5 +114,100 @@ namespace :bonnie do
...
+        new_datatype_facility['values'][0]['code'] = {'code_system'=>code_system, 'code'=>code}
+        datatype['facility'] = new_datatype_facility 
+      end
+    end
+
+    task :update_facilities_and_diagnoses => :environment do
+      # Update old facilities and diagnoses to be collections with single elements
+      Record.all.each do |patient|
+        if patient.encounters
+          patient.encounters.each do |encounter, index|
```
mulcahyk: If you want to get the index use `.each_with_index`.  I don't see where you are using the index though.

lib/tasks/cql.rake
```diff
@@ -114,5 +114,100 @@ namespace :bonnie do
       end
     end
 
+    def update_facility(patient, datatype)
+      if datatype['facility'] && !datatype['facility']['type']
+        print "\e[#{32}m#{"[Facility]"}\e[0m"
```
mulcahyk: I like the coloring!

mulcahyk (changes requested):
The rake task is not updating the source_data_criteria.  This needs to be addressed.

lib/tasks/cql.rake
lukoktonos: That's true--this change would need to be reflected in cql_qdm_patientapi, health-data-standards, and bonnie, however, so I don't think it's in the scope of this PR to modify.

lib/tasks/cql.rake
lukoktonos: Facility is a possible sub-field of {encounter, adverse event, procedure}, and diagnosis is a sub-field of encounter.

lib/tasks/cql.rake
lukoktonos: Good catch--I had refactored the loop but forgot to remove the 'index'.

lib/tasks/cql.rake
lukoktonos: Good catch--don't know what that's doing there.

mulcahyk:
This shows an example of a patient converted by the rake task with their source_data_criteria and the new encounter['facility'].  
image
This still has the old field_values but not the new structure for facility location.

Here is a patient added using the latest cql4bonnie.
image

alexanderelliott121:
JIRA test has a typo.
bundle exec **rails** bonnie:cql:update_facilities_and_diagnoses

should be

bundle exec **rake** bonnie:cql:update_facilities_and_diagnoses

alexanderelliott121:
I've run the test and had it pass.

alexanderelliott121 (dismissed):
Test passed

lib/tasks/cql.rake
```diff
@@ -114,5 +114,140 @@ namespace :bonnie do
...
+                  new_facility_location['values'][0]['code_list_id'] = source_data_criterium['field_values']['FACILITY_LOCATION']['code_list_id']
+                  new_facility_location['values'][0]['field_title'] = source_data_criterium['field_values']['FACILITY_LOCATION']['field_title']
+                  new_facility_location['values'][0]['start_date'] = source_data_criterium['field_values']['FACILITY_LOCATION']['start_date']
+                  new_facility_location['values'][0]['start_time'] = source_data_criterium['field_values']['FACILITY_LOCATION']['start_time']
+                  new_facility_location['values'][0]['end_date'] = source_data_criterium['field_values']['FACILITY_LOCATION']['end_date']
+                  new_facility_location['values'][0]['end_time'] = source_data_criterium['field_values']['FACILITY_LOCATION']['end_time']
+                  new_facility_location['values'][0]['title'] = source_data_criterium['field_values']['FACILITY_LOCATION']['title']
+                  new_source_data_criterium_field_values['FACILITY_LOCATION'] = new_facility_location
```
mulcahyk: source_data_criteria['field_values]['FACILITY_LOCATION'] in the old version does not have any time information.  Time information is in the fields FACILITY_LOCATION_ARRIVAL_DATETIME and FACILTIY_LOCATION_DEPARTURE_DATETIME.

lib/tasks/cql.rake
```diff
@@ -114,5 +114,140 @@ namespace :bonnie do
...
+        new_datatype_facility['values'][0]['code'] = {'code_system'=>code_system, 'code'=>code}
+        datatype['facility'] = new_datatype_facility 
+      end
+    end
+
+    task :update_facilities_and_diagnoses => :environment do
+      # For any relevant datatypes, update old facilities and diagnoses to be collections with single elements
+      Record.all.each do |patient|
+        if patient.source_data_criteria
+          patient.source_data_criteria.each_with_index do |source_data_criterium, index|
```
mulcahyk: I don't see where you are using `index` - consider removing.

mulcahyk (changes requested):

edeyoung:
Looks like "key" may be  missing from source data criteria conversion. Top is a converted value. Bottom is a value built afterwards.

image

edeyoung:
Looks like 'key' is missing and the 'cd' type needs to be modified to 'fac'. Top is old converted one. Bottom is new one.

image

Not in the scope of this PR, but the time likely should not be set if the date is set for a facility location. Will create ticket.

lib/tasks/cql.rake
```diff
@@ -114,5 +114,159 @@ namespace :bonnie do
...
+                  new_diagnosis['values'] = [{}]
+                  new_diagnosis['values'][0]['type'] = 'CD'
+                  new_diagnosis['values'][0]['key'] = 'DIAGNOSIS'
+                  new_diagnosis['values'][0]['code_list_id'] = source_data_criterium['field_values']['DIAGNOSIS']['code_list_id']
+                  new_diagnosis['values'][0]['field_title'] = source_data_criterium['field_values']['DIAGNOSIS']['field_title']
+                  new_diagnosis['values'][0]['title'] = source_data_criterium['field_values']['DIAGNOSIS']['title']
+                  new_source_data_criterium_field_values['DIAGNOSIS'] = new_diagnosis 
+
+                # update any 'FACILITY_LOCATION' field values that aren't collections
+                elsif field_value_key == 'FACILITY_LOCATION' && !(source_data_criterium['field_values']['FACILITY_LOCATION']['type'] == 'COL')
```
mulcahyk, line 214: This is assuming that all encounters will have a FACILITY_LOCATION.  Most of the current patients don't have a FACILITY_LOCATION; only 14 of 90 do.  The rest have only the times.  When this runs it is deleting the time only soruce_data_criteria.

mulcahyk (changes requested):

mulcahyk (changes requested):
The issue I'm seeing still isn't fixed.
Look at this test patient:
First: "InEDPsychDiagPrincipal"
Last: "IPStrat1Pass"
Measure_ids: "28AC347D-2F91-4A0C-9395-2602134BAA89"
This test patient in the original has facility arrival and departure times only; no location.  After running the rake task the source_criteria with those field_values only has the PRINCIPAL_DIAGNOSIS left.  When you look at the encounter there is no facility information.

lukoktonos:
@mulcahyk I think this is the correct behavior.  For the new version, in order to add a facility arrival/departure, the UI validates that a code has been entered.  It is impossible to add a time without a code.  

We purposefully remove any facility objects (and the corresponding source_data_critieria arrival/departure times) for any patients that have time information without a location code.

lib/tasks/cql.rake
lukoktonos: I believe this is the intended behavior of the script.  @edeyoung is this the case?  If not, I can remove the part of the script that deletes these items.

edeyoung:
@mulcahyk yes, this is the correct behavior. The Location Period needs to be linked to a Location code.

edeyoung:
@losborne can you make the log show what patients in what measures in what accounts have this removed?

lib/tasks/cql.rake
```diff
@@ -114,5 +114,175 @@ namespace :bonnie do
...
 
+    def print_helper(title, patient)
+      if title == 'Removing'
+        printf "%-22s", "\e[#{31}m#{"[#{title}] "}\e[0m"
+      else
+        printf "%-22s", "\e[#{32}m#{"[#{title}] "}\e[0m"
+      end
+      printf "%-80s", "\e[#{36}m#{"#{patient.first} #{patient.last}"}\e[0m"
+      puts "#{patient.measure_ids[0]}"
+    end
```
mulcahyk: More cool logging.  Love it!

lukoktonos:
@edeyoung log updated

lib/tasks/cql.rake
```diff
@@ -114,5 +114,175 @@ namespace :bonnie do
...
+
+        # code
+        if datatype.facility['code']
+          code_system = datatype.facility['code']['code_system']
+          code = datatype.facility['code']['code']
+          new_datatype_facility['values'][0]['code'] = {'code_system'=>code_system, 'code'=>code}
+          datatype.facility = new_datatype_facility
+        else
+          print_helper("Removing", patient)
+          datatype.remove_attribute(:facility)
```
mulcahyk: This is a small issue.
It looks like every patient that has a `[Facilty]` also has a `[Removing]`.

lib/tasks/cql.rake
```diff
@@ -114,5 +114,175 @@ namespace :bonnie do
...
+                  new_facility_location['values'][0]['end_time'] = converted_end_time
+
+                  # Reassign
+                  new_source_data_criterium_field_values['FACILITY_LOCATION'] = new_facility_location 
+                elsif !(field_value_key == 'FACILITY_LOCATION_ARRIVAL_DATETIME' || field_value_key == 'FACILITY_LOCATION_DEPARTURE_DATETIME')
+                  # add unaltered field value to new structure, unless it's a time we already used above
+                  new_source_data_criterium_field_values[field_value_key] = field_value_value
+                else
+                  # There was an arrival/depature time without a code, remove them
+                  new_source_data_criterium_field_values.delete(field_value_key)
```
mulcahyk: Can the `Removing` logging be done here as well?

mulcahyk (dismissed):

mulcahyk (approved):

PR #847, opened 2017-09-19
Added warning message for outdated (pre-QDM 5.3) measures

c-monkey:
JIRA: https://jira.mitre.org/browse/BONNIE-1074
TESTS: added unit tests in cql_logic_view_spec.js.coffee

I recommend comparing the test file with the `?w=1` url suffix because I had to indent most of the file to split into 2 test fixtures

If you wish to see this in a manual test, download a package (or use any you already have) prior to MAT 5.4 from: https://gitlab.mitre.org/bonnie/internal-documentation/wikis/cql-measures-for-testing, upload to Bonnie, and open the measure details screen.


app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -67,8 +68,20 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
     @unusedStatementViews = []
     
     # Look through all elm library structures, and check for CQL errors noted by the translation service.
+    # Also finds if using old versions of QDM
     if Array.isArray @model.get('elm')
       _.each @model.get('elm'), (elm, elm_index) =>
+        _.each elm.library.usings.def, (id) =>
+          if (id.localIdentifier == "QDM")
```
holmesie: It's not really Coffeescript style to use parens in If, Else, Unless, and Conditional Assignments, e.g. `if (stuff)`, but since I'm seeing it in other places in this file, I don't think it's an approval stopper.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -67,8 +68,20 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
...
+            versionParts = id.version.split('.')
```
holmesie: Is it possible at any point for either `id` itself being null, or in this case, `version` being null (which causes the `.split` to raise an error)?

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -67,8 +68,20 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
...
+            if (versionParts.length <2)
+              @hasOutdatedQDM = true
+            else
+              # explicitly checking for versions of QDM prior to 5.3
+              major = parseInt(versionParts[0], 10)
+              minor = parseInt(versionParts[1], 10)
+              if (major < 5 || (major == 5 && minor < 3))
+                @hasOutdatedQDM = true
```
holmesie: I think this all (previous 9 lines) can be simplified to:
```
version = parseFloat(id?.version)
unless isNaN(version)
   @hasOutdatedQDM = version < 5.3
```
parseFloat will ignore the second potential `.<number>`, so if we had `"5.0.2"`, it would return `5.0`.

NOTE: this would only make sense if we didn't ever care about patch numbers in QDM versions, which I am unsure of.

holmesie (changes requested):

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
c-monkey: Although extremely appealing, I think removing the flexibility isn't appropriate in the general case.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
c-monkey: Nice catch. I was following the existing style as opposed to doing the right thing. Thanks!

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
c-monkey: I added a null check just in case.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
c-monkey: fixed this and some of the accompanying code

holmesie (dismissed):
Tests pass, my comments have been addressed, approved.

alexanderelliott121 (dismissed):

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -59,18 +59,31 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
...
-    if Array.isArray @model.get('elm')
-      _.each @model.get('elm'), (elm, elm_index) =>
+    # Also finds if using old versions of QDM
+    if Array.isArray @model.get 'elm'
+      _.each @model.get('elm'), (elm) =>
+        _.each elm.library.usings.def, (id) =>
+          if id?.localIdentifier == "QDM"
+            versionParts = id.version.split '.'
+            if versionParts.length <2
+              @hasOutdatedQDM = true
```
edeyoung: I doubt this will happen, but what if we go to QDM 5.3.1?

This would falsely fire.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
```diff
@@ -59,18 +59,31 @@ class Thorax.Views.CqlPopulationLogic extends Thorax.Views.BonnieView
...
+            else
+              # explicitly checking for versions of QDM prior to 5.3
+              major = parseInt versionParts[0], 10
+              minor = parseInt versionParts[1], 10
+              if major < 5 || (major == 5 && minor < 3)
+                @hasOutdatedQDM = true
```
edeyoung: What about something like this: 

```
Gem::Version.new('0.4.1') > Gem::Version.new('0.10.1')
```

from here: https://stackoverflow.com/questions/2051229/how-to-compare-versions-in-ruby

This would make it more generalized and we can then add the version we care about as a config variable (which we should do)

edeyoung (dismissed):
See comments

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
c-monkey: I consent to this proposal.

c-monkey: This would be perfect if this code were in Ruby. I'll figure out an equivalent for front-end.

edeyoung: ha :) yes, thanks!

alexanderelliott121:
I'm going to hold off on approval until the branch has been merged/rebased

app/assets/javascripts/helpers/compare_version.js.coffee
```diff
@@ -0,0 +1,15 @@
+class CompareVersion
```
holmesie, line 1: I like this!

config/bonnie.yml
```diff
@@ -11,6 +11,8 @@ defaults: &defaults
     profiles_url: https://vsac.nlm.nih.gov/vsac/profiles
     profile: Latest eCQM
   enable_client_error_email: true
+  # oldest version of QDM that can be loaded
+  support_qdm_version: 5.3
```
holmesie, line 15: This will be nice to have

holmesie (changes requested):
What's going on with the Travis failures?

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
c-monkey: I have added a unit test to show QDM 5.3.1 is equal to or newer than QDM 5.3 and will now pass.

app/assets/javascripts/views/logic/cql_logic_view.js.coffee
c-monkey: The new CompareVersion class should handle this.

c-monkey:
Still investigating 1 travis failure. I fixed some BonnieRouter global variable assignment to work both in the App and in Jasmine tests, but am still investigating the other failure.

c-monkey:
This test failure is similar to one we fixed in the master branch. The test fixture is being changed by other tests execution. One can reproduce this test failure in the `cql4bonnie` branch in the web-teaspoon runner by running just the MeasureView suite. Regardless, this PR is going to be held up until I can fix that test. I also noticed, while debugging this, that the new suite I added `CompareVersion` always runs, for example when you select a suite to debug.

holmesie (dismissed):

c-monkey:
rebased

alexanderelliott121 (approved):

PR #848, opened 2017-09-19
initial pull request template

c-monkey:
Testing if committing this to `cql4bonnie` works, or if we have to add them to the `master` branch.

After merged, should populate PR with the template file.


c-monkey:
@edeyoung I'm going to merge into cql4bonnie and test if this works. I assume it will not, as detailed in https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/, but would be happy if it does work.

PR #870, opened 2017-10-16
Cql4bonnie 357

mayerm94:
Check out bonnie_bundler and bonnie branches cql4bonnie_357. The idea here is to add a value to the valueset model that indicates whether the valueset in question is a direct reference code. If it is a direct reference code, it will not have an oid so we need to change the view to say "Direct Reference Code" instead of the guid that is currently being displayed instead of the oid.
https://oncprojectracking.healthit.gov/support/browse/BONNIE-357
https://jira.mitre.org/browse/BONNIE-1090

**Submitter:**
- [x] This pull request describes why these changes were made.
- [x] JIRA ticket for this PR:
- [x] Code diff has been done and been reviewed

- [x] JIRA test links: N/A

**Reviewer 1.5:** @losborne

Name:
- [ ] Code is maintainable and reusable, reuses existing code and infrastructure where appropriate, and accomplishes the task’s purpose
- [ ] You have tried to break the code

**Reviewer 2:**

Name:
- [ ] Code is maintainable and reusable, reuses existing code and infrastructure where appropriate, and accomplishes the task’s purpose
- [ ] The tests appropriately test the new code, including edge cases
- [ ] You have tried to break the code

- [ ] JIRA tests have been run and pass
- [ ] You agree with the justification for use of JIRA tests or have provided input on why you disagree


test/functional/measures_controller_test.rb
```diff
@@ -217,11 +241,11 @@ class << measure_file
     post :create, {measure_file: measure_file, measure_type: 'eh', calculation_type: 'episode'}
     assert_equal "Error Uploading Measure", flash[:error][:title]
     assert_equal "The uploaded zip file is not a valid Measure Authoring Tool export of a CQL Measure.", flash[:error][:summary]
-    assert_equal "You have uploaded a zip file that does not appear to be a Measure Authoring Tool CQL zip file please re-export your measure from the MAT and select the 'eMeasure Package' option", flash[:error][:body]
+    assert_equal "You have uploaded a zip file that does not appear to be a Measure Authoring Tool CQL zip file please re-export your measure from the MAT and select the 'eMeasure Package' option. Please use https://bonnie.healthit.gov/ for HQMF based measures.", flash[:error][:body]
```
holmesie: The "Please use https://bonnie.healthit.gov/ for HQMF based measures." is a bit misleading, as the CQL measures still use HQMF. What about something like, "Please use https://bonnie.healthit.gov/ for QDM logic based HQMF measures."?

lib/tasks/cql.rake
```diff
@@ -114,6 +114,35 @@ namespace :bonnie do
...
+              vs["direct_reference"] = false
+            else
+              vs["direct_reference"] = true
+            end
+          end
+          measure.save
+          successful += 1
+        rescue Exception => ex
+          errorString = ex.to_s + " in measure " + measure.title + " " + measure._id.to_s
+          puts errorString
```
holmesie: Why not just `puts ex.to_s + " in measure " + measure.title + " " + measure._id.to_s`?

lib/tasks/cql.rake
```diff
@@ -114,6 +114,35 @@ namespace :bonnie do
...
+        end
+      end
+
+      successString = "Successfully updated " + successful.to_s + "/" + total.to_s
+      puts successString
```
holmesie: Why not just `puts "Successfully updated " + successful.to_s + "/" + total.to_s`?

holmesie (changes requested):

test/functional/measures_controller_test.rb
mayerm94: There is a branch out now https://github.com/projecttacoma/bonnie/commits/BONNIE-987-qdm-redirect that is redoing all of this text

lib/tasks/cql.rake
mayerm94: Personal preference/style/Artifact from when I was testing the rake task.

test/functional/measures_controller_test.rb
holmesie: Good to know, I'll make this comment there instead.

holmesie (approved):

holmesie (changes requested):
Reverting approval for the time being

alexanderelliott121:
Quick note: pull request does not actually have description of why the changes were made.
Or jira link

app/assets/javascripts/views/measure_value_sets.js.coffee
```diff
@@ -33,21 +33,19 @@ class Thorax.Views.MeasureValueSets extends Thorax.Views.BonnieView
...
             code.hasLongDisplayName = code.display_name.length > 160
         else
           version = ''
           codeConcepts = []
 
         codes = new Backbone.PageableCollection(@sortAndFilterCodes(codeConcepts), @pagination_options)
-        if version.match(/^Draft/)
-          version = "Draft"
-        valueSet = { name: name, oid: oid, version: version, codes: codes, cid: cid }
+        valueSet = { name: name, oid: oid, version: version, codes: codes, cid: cid, direct_reference: direct_reference }
```
alexanderelliott121: Is there a reason you're nixing the check to see if the version is a draft?

holmesie:
Looks like some issues with the tests failing, is that expected? Also has some conflicts that need to be addressed.

alexanderelliott121 (changes requested):
The check to determine if a version is draft or not still needs to be addressed.

Was the draft check removed for a reason?

app/assets/javascripts/views/measure_value_sets.js.coffee
mayerm94: Accidental deletion, good catch

lib/tasks/cql.rake
```diff
@@ -114,6 +114,33 @@ namespace :bonnie do
...
+      marked as a direct reference code
+    $ rake bonnie:cql:update_value_set_direct_reference_field }
+    task :update_value_set_direct_reference_field => :environment do
+      successful = 0
+      total = 0
+      CqlMeasure.all.each do |measure|
+        begin
+          total += 1
+          value_sets = measure.value_set_oid_version_objects
+          value_sets.each do |vs|
```
edeyoung, line 139: suggest printing out all modified value sets so we can visually confirm that they are all direct reference codes

lib/tasks/cql.rake
```diff
@@ -55,7 +55,7 @@ namespace :bonnie do
     end
 
     desc %{Outputs user accounts that have cql measures and which measures are cql in their accounts.
-      Example test@test.com  
+      Example test@test.com
```
edeyoung, line 70: several white space related changes. please revert

app/assets/javascripts/templates/measure/value_sets/value_sets.hbs
```diff
@@ -6,8 +6,14 @@
       <a data-toggle="collapse" href="#codes_{{cid}}" class="collapsed value_sets">
         <div class="row">
           <div class="col-sm-5"><strong>{{name}}</strong></div>
-          <div class="col-sm-5">{{oid}}</div>
-          <div class="col-sm-2">{{version}}</div>
+          {{#ifCond direct_reference "==" false}}
```
edeyoung, line 9: can this just be `{{#if direct_reference}}` because direct_reference is a boolean?

edeyoung:
I would like to see some sort of test of the script itself (JIRA?) to help confirm that we are not doing anything we don't want to do. @holmesie @alexanderelliott121 did you both test the script and confirm it works as expected?

lib/tasks/cql.rake
```diff
@@ -114,6 +114,33 @@ namespace :bonnie do
       end
     end
 
+    desc %{Iterate over value sets, if the oid is not in the dotted format, it should be
+      marked as a direct reference code
+    $ rake bonnie:cql:update_value_set_direct_reference_field }
+    task :update_value_set_direct_reference_field => :environment do
```
holmesie, line 132: I've just had a thought about this task being in this file. We have a `bonnie_run_once.rake`, which is, in it's own words, "This rakefile is for tasks that are designed to be run once to address a specific problem". Perhaps we should start that tradition for CQL for now on, perhaps putting them in a `cql_run_once.rake`? I kind of like the sound of that seperation... Thoughts?

lib/tasks/cql.rake
mayerm94: These whitespace changes were made before I enabled the feature in atom to not automatically fix trailing whitespace. I agree that the white space changes make the PR a little gross but I don't feel right deliberately making the codebase worse by going back and re-inserting meaningless whitespace (I can't just revert the commit because the commit includes actual non-whitespace changes)

okeefm: You can amend the commit if need be (though it would get tricky if this isn't the last commit on this PR).

lib/tasks/cql.rake
mayerm94: Perhaps.....

mayerm94: I like this idea, or we should start considering what belongs in bonnie vs what belongs in the server_scripts repo

alexanderelliott121 (dismissed):

lib/tasks/cql.rake
```diff
@@ -114,6 +114,33 @@ namespace :bonnie do
...
+          total += 1
+          value_sets = measure.value_set_oid_version_objects
+          value_sets.each do |vs|
+            if vs["oid"].include? "."
+              vs["direct_reference"] = false
+            else
+              vs["direct_reference"] = true
+            end
+          end
+          measure.save
```
lukoktonos, line 146: Either `measure.save!` or 
```
if measure.save
  successful += 1
end
```
would be better here, in case it fails on saving the measure.

lib/tasks/cql.rake
lukoktonos: I agree that this printout would be helpful.

lukoktonos (changes requested):
The fixtures in `/spec/javascripts/fixtures/json/measure_data/CQL/*` also need to be updated.

mayerm94:
JIRA Test for the rake task https://jira.mitre.org/browse/BONNIE-1158

app/assets/javascripts/templates/measure/value_sets/value_sets.hbs
mayerm94: Handlebars doesn't seem to support writing this like that

alexanderelliott121:
I'm trying to run through the test for the rake task.
Unfortunately, the screenshots for expected results are too small for me to actually be able to compare them against my results.

alexanderelliott121:
I've run the JIRA test, result appears mixed.

Upside, the db fields have 3 values (oid, version, direct_reference).

On the downside, steps 6 failed: the oid and version fields were populated, which the test indicates should not be the case.

I thought the issue might have been a result of the pre-existing state of my db, but they persisted after I cleared my db and re-ran the tests.

I suspect that this is a result of changes to the cql4bonnie branch.

app/assets/javascripts/templates/measure/value_sets/value_sets.hbs
lukoktonos: I remember struggling with this for the special Components view.  It will work if (and probably only if) you flip the logic like this:

```
            {{#if direct_reference}}
              <div class="col-sm-5">Direct Reference Code</div>
              <div class="col-sm-2">N/A</div>
            {{else}}
              <div class="col-sm-5">{{oid}}</div>
              <div class="col-sm-2">{{version}}</div>
            {{/if}}
```

You can't do something like `{{#if !direct_reference}}` though, annoyingly.

hossenlopp: Handlebars has an `{{#unless}}` block.

PR #877, opened 2017-10-19
Pr template update

edeyoung:
Updates to the PR template


c-monkey (approved):
We need to clearly document how one can get before and after code coverage reports. For example, for our Bonnie coffescript teaspoon doesn't automatically generate reports. I don't think this is appropriate in this PR template, but ratther in the https://github.com/projecttacoma/bonnie/blob/master/.github/CONTRIBUTING.md or on the internally linked page.

PR #878, opened 2017-10-20
gemfile updates

edeyoung:
Gemfile update sanity check


mayerm94 (approved):
These are the most recent commit hashes on the respective branches

PR #883, opened 2017-10-26
Cql4bonnie use mat artifacts

jbradl11:

Purpose: Is to support the new MAT packages that contain the JSON artifacts. In this PR rebuild_elm has been updated to support regenerating the elm using the new stored MAT packages if there is one, otherwise it will call the translation service for older measures.

JIRA TICKET: https://jira.mitre.org/browse/BONNIE-1153

**Submitter:**
- [x] JIRA ticket for this PR:

- [x] JIRA test links:
- [x] Justification for using JIRA tests:
- [x] JIRA tests have been added to sprint

**Reviewer 1:**

Name:

**Reviewer 2:**

Name: @losborne

- [x] JIRA tests have been run and pass NA
- [x] You agree with the justification for use of JIRA tests or have provided input on why you disagree NA


test/unit/rebuild_elm_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    # Modify some of the measure model
+    measure.title = 'No mat package'
+    measure.elm_annotations = nil
+
+    # Remove the stored MAT package from the DB
+    package = CqlMeasurePackage.all.first
+    package.delete
+    measure.save
+
+    # Confirm measure model saved corectly.
```
lukoktonos: Trailing whitespace

test/unit/rebuild_elm_test.rb
```diff
@@ -0,0 +1,92 @@
...
+    VCR.use_cassette("valid_translation_response") do
+      # Run rake task on all cql measures
+      Rake::Task['bonnie:cql:rebuild_elm'].execute
+    end
+    
+    measure = CqlMeasure.where({hqmf_set_id: "7B2A9277-43DA-4D99-9BEE-6AC271A07747"}).first
+    # Confirm that measure title did not update.
+    assert_equal measure.title, 'No mat package'
+
+    # Confirm there is still now associated MAT package.
```
lukoktonos: "now associated" -> "no associated"

okeefm:
Are there JIRA tests on this PR? I'm not sure how to find them if there are.

lib/tasks/cql.rake
```diff
@@ -16,22 +16,34 @@ namespace :bonnie do
...
+              File.open(File.join(dir, measure.measure_id + '.zip'), 'wb') do |zip_file|
+                # Write the package binary to a zip file.
+                zip_file.write(measure.package.file.data)
+                files = Measures::CqlLoader.get_files_from_zip(zip_file, dir)
+                cql_artifacts = Measures::CqlLoader.process_cql(files, main_cql_library, user)
+                cql = files[:CQL]
+              end
+            end
+          # If the measure does not have a MAT package stored, continue as we have in the past using the cql to elm service
           else
```
losborne: This else statement isn't covered by tests according to the coverage tool:
image

test/unit/rebuild_elm_test.rb
```diff
@@ -0,0 +1,92 @@
...
+
+      # Run rake task on all cql measures
+      Rake::Task['bonnie:cql:rebuild_elm'].execute
+      measure = CqlMeasure.where({hqmf_set_id: "7B2A9277-43DA-4D99-9BEE-6AC271A07747"}).first
+      # Confirm that measure title did not update.
+      assert_equal measure.title, 'foo'
+      
+      # Confirm that the CQL and ELM annotations were updated.
+      assert_not_equal nil, measure.cql
+      assert_not_equal nil, measure.elm_annotations
```
lukoktonos, line 42: Should there be an assert here related to `measure.package`?

lukoktonos (changes requested):

test/unit/rebuild_elm_test.rb
jbradl11: We have a unit test to check this in Bonnie_bundler. I don't see a need to re-test, but I can if you think its worth it.

lukoktonos (dismissed):
Regression test passes with a rebuild_elm in-between.  

Confirmed file stored in db matches original zip package.

Everything looks good--approved.

okeefm (dismissed):
Looks good to me. rake task runs successfully, Bonnie seems to work afterwards, automated tests pass.

lukoktonos (approved):
Still looks good with updated hash.

lib/tasks/cql.rake
```diff
@@ -16,22 +16,34 @@ namespace :bonnie do
...
-          # Grab the measure cql
-          if measure[:cql].instance_of?(Array) # New type of measure
-            cql = measure[:cql]
-          elsif measure[:cql].instance_of?(String) # Old type of measure
-            cql = [measure[:cql]]
-          end
+          cql = nil
+          cql_artifacts = nil
           # Grab the name of the main cql library
-          if measure[:main_cql_library].present?
```
edeyoung, line 26: should we be making the assumption that this is present? are we not worried about older measures? should we delete all old measures? (i assume they're on mitre accounts, but we should confirm of course).

lib/tasks/cql.rake
```diff
@@ -16,22 +16,34 @@ namespace :bonnie do
...
+                cql = files[:CQL]
+              end
+            end
+          # If the measure does not have a MAT package stored, continue as we have in the past using the cql to elm service
           else
-            main_cql_library = elms.first['library']['identifier']['id']
+            # Grab the measure cql
+            cql = measure[:cql]
+             # Use the CQL-TO-ELM Translation Service to regenerate elm for older measures.
+            elm_json, elm_xml = CqlElm::CqlToElmHelper.translate_cql_to_elm(cql)
```
edeyoung, line 42: just want to confirm: if we have no packages using the CQL-to-ELM translator, and we don't have it running, this *will not* break, right? It would only break if a package needed the CQL-to-ELM translator?

lib/tasks/cql.rake
jbradl11: Yes I believe we can make this assumption, because if you look in the 'else' clause of this, there is a variable that is never set ("elms"). So I am pretty confident that this task would have crashed previously if measure[:main_cql_library] was not present.

lib/tasks/cql.rake
jbradl11: Correct, this will task will work properly with the translator off if all measures being updated do not require it. I just ran my unit test without the service running, and it worked properly.

okeefm (approved):
Still looks good; rake task still works

PR #884, opened 2017-10-27
Rake task that adds JSON ELM to MAT packages.

hossenlopp:
bonnie:cql:add_json_to_package task takes a MAT package zip and builds a new one in the same directory that has the JSON ELM added. It handles the odd situations of older packages without clause level annotations by replacing the XML ELM with stuff from the translation service.

**Submitter:**
- [x] JIRA ticket for this PR: https://jira.mitre.org/browse/BONNIE-1156
- [x] If UI changes have been made, google WAVE plug-in has been executed to ensure no 508 issues were introduced. N/A
- [x] Tests are included and test edge cases (N/A, This is a utility rake task that would be immensely burdensome to write tests for.)
         1. Note why this could not be done or is not applicable here: This is a utility rake task that would be immensely burdensome to write tests for.
- [x] Automated regression test(s) pass N/A, doesn't change calculation code.

- [x] JIRA test links: N/A
- [x] JIRA tests have been added to sprint N/A, but this was used to update JIRA tests.

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lukoktonos (dismissed):
Code looks good.
New .zip file has identical files except for additional json.  
Measure behaves appropriately after uploading it into Bonnie.
Practice deployment used these measures and all selected tests passed.

Approved.

lukoktonos (dismissed):
Changes look good.  Tested with both bash and zsh--working as expected.

lib/tasks/cql.rake
```diff
@@ -333,5 +333,134 @@ namespace :bonnie do
...
+      files = Measures::CqlLoader.get_files_from_zip(File.new(input_package_path), temp_path)
+      raise Exception.new("Package already has ELM JSON!") if files[:ELM_JSON].length > 0
+
+      # translate_cql_to_elm
+      elm_jsons, elm_xmls = CqlElm::CqlToElmHelper.translate_cql_to_elm(files[:CQL])
+
+      # older packages don't have annotations or clause level annotations, if they dont have them wipe out the existing
+      # XML ELM and use the ones from the translation server
+      # start by assuming there are annotations, then set this to false
+      annotations_exist = true
```
jbradl11, line 374: Should we just always update the XML so we know it matches the JSON? That way we know both artifacts were generated with the same version of the translation service.

hossenlopp: The rake task can take an optional parameter to force XML replacement. Should I remove or flip that option and make it default to replace XML?

jbradl11: I think that is the correct way to go. But if you have a good reason not to change it, let me know.

hossenlopp: Which is the correct way to go? Is it removing the ability to keep the original XML ELM?

I'd argue it should be kept. So we can optionally test the XML ELM because we need to support it anyways.

hossenlopp:
Updated branch and made it so replacing the XML ELM is the default operation.

lukoktonos (approved):
Latest changes look good.

jbradl11 (approved):
Code looks good, ran a few times using different packages while passing in different flags. .

PR #895, opened 2017-11-09
updated user guides

edeyoung:
Updates the user guide.

Need to confirm that the correct user guide shows when a user clicks the user guide links in the splash page and in the help drop down.


alexanderelliott121 (changes requested):
Valueset description, page 13, does not reflect addition of Terminology.

On another note, when viewing the guide in chrome the tab title is "Bonnie_user_guide_v2.0_no_draft".
Don't know if this really matters.

alexanderelliott121 (dismissed):

mayerm94 (dismissed):

alexanderelliott121 (approved):

mayerm94 (approved):

PR #896, opened 2017-11-14
Accesible user guide

edeyoung:
Added editor's accessible user guide.


edeyoung:
something odd with this PR... redoing.

PR #897, opened 2017-11-14
updated user guides to 508 version.

edeyoung:
Confirm user guide is updated as expected with tool tips and shows as expected when accessed from the website.


mayerm94 (approved):
Looks good, mouse over text works

PR #898, opened 2017-11-20
new users guides from the editor

edeyoung:
Sanity check - new user guide from the editor. Need to confirm it loads properly, etc.


edeyoung:
This does not have the "XML" change in it.

mayerm94 (approved):
Can download locally, mouseover text works

PR #923, opened 2018-01-04
Modified rebuild_elm to update the measure's source_data_criteria and…

jbradl11:

Running the Rebuild Elm task would not update the data criteria and source data criteria code_list_ids. This is an issue for direct reference codes because of the GUIDs we use to act as their value set. The DRC are given new GUIDS in the Rebuild elm but the reference (code_list_id) on the data criteria never gets updated. This PR adds a new function call in the rebuild elm task to update the source data criteria and data criteria. 

*NOTE* You need to point the Gemfile to the Bonnie_Bundler branch 'rebuild_elm_value_set_drc_fix'

**Submitter:**
- [x] JIRA ticket for this PR: https://jira.mitre.org/browse/BONNIE-1242

- [x] JIRA test links: N/A

**Reviewer 1:** @hossenlopp 

Name:

**Reviewer 2:**

Name: @mayerm94


lukoktonos (dismissed):
Issue is fixed after rebuilding elm, code and tests look good, tests pass.

Only note is that existing patients may need to be rebuilt if they were built with the broken measure (already discussed with JB & lizzie).

mayerm94 (commented):
I am unsure about the consequences of this, but when you run the rebuild elm rake task, the code_list_id on source_data_criteria on the patient is not updated to the new GUID. It doesn't seem to have any calculation effects and the code_list_id field on the patient may not even be used. Attached are two screenshots, the first with index 0 is of a data criteria that was dragged on before rebuild_elm was run and the second, with index 1, is of the data criteria that was dragged on after rebuild_elm was run. The difference between the two is the code_list_id field. The code_list_id is equal to the direct reference code GUID that the data_criteria on the measure used at the time the data_criteria was dragged onto the patient. @jbradl11 @edeyoung 
drc_sdc_before
drc_sdc_after

lukoktonos:
Regression test passes as of 62b7c70e8c55bb68d8d807987aa86c0ac96415e4

lukoktonos (dismissed):
Most recent changes look good.

Tests and regression test pass, running update_source_data_criteria multiple times shows that all patients were hit the first run (no errors) and are not updated after the first run (0 updated).

Updating a measure after running these tasks shows expected behavior.

mayerm94 (dismissed):
DB comparison matches rake task outputs, single run and double run come back clean, no differences.

hossenlopp (changes requested):
The `bonnie:patients:rebuild_elm_update_drc_code_list_ids` task does not clean up the replaced DRC valueset references and entries.

There are no tests for the updates to the `bonnie:patients:update_source_data_criteria`.

hossenlopp:
It also looks like the `bonnie:patients:update_source_data_criteria` task is replacing some direct reference value set codes with real value set OIDs on the patients. This doesn't quite make sense.

jbradl11:
The `bonnie:patients:rebuild_elm_update_drc_code_list_ids` purposefully didn't remove the existing DRC valueset references. If you believe its worth cleaning it up, i am fine with that. Also, the `bonnie:patients:update_source_data_criteria` is a "run once" rake task, which typically doesn't have tests. @losborne would know more about that rake tasks specifically.

hossenlopp:
@losborne and I looked at the patients that had their DRC GUID changed to an OID. Turns out it was on a measure that had a DRC previously that was replaced with a value set for one of the source data criteria. (Or so we think...)

@jbradl11 Having duplicated DRC on the measure and in the value sets collection make me worried that odd duplications could popup in the UI or affect calculation results.

hossenlopp (dismissed):

mayerm94 (dismissed):
old GUID removal since last review makes sense. Re-ran rake task and results look as expected

hossenlopp (approved):

PR #962, opened 2018-03-16
Bonnie 1199 update patient import

lukoktonos:
This PR integrates the update_facilities_and_diagnoses rake task with the patient import rake task, so that it doesn't need to be run for every patient import.

NOTE: After this PR is integrated, the patient export instructions need to be edited to remove the step to update facilities and diagnoses.

Pull requests into Bonnie require the following. Submitter and reviewer should :white_check_mark: when done. For items that are not-applicable, note it's not-applicable ("N/A") and :white_check_mark:.

**Submitter:**
- [x] This pull request describes why these changes were made.
- [x] This PR is into the correct branch.
- [x] JIRA ticket for this PR: https://jira.mitre.org/browse/BONNIE-1199
- [x] JIRA ticket links to this PR
- [x] Code diff has been done and been reviewed (it **does not** contain: additional white space, not applicable code changes, debug statements, etc.)
- [x] If UI changes have been made, google WAVE plug-in has been executed to ensure no 508 issues were introduced. N/A
- [x] Tests are included and test edge cases
- [x] Tests have been run locally and pass (remember to update Gemfile when applicable)
- [x] Code coverage has not gone down and all code touched or added is covered. 

Branch | Coverage
-- | --
master | 1566 / 1783 LOC (87.83%) covered.
bonnie-1199-update-patient-import | 1690 / 1919 LOC (88.07%) covered.

- [x] Automated regression test(s) pass

If JIRA tests were used to supplement or replace automated tests:
- [x] JIRA test links: N/A
- [x] Justification for using JIRA tests: N/A
- [x] JIRA tests have been added to sprint N/A

**Reviewer 1:** @mayerm94

Name:
- [x] Code is maintainable and reusable, reuses existing code and infrastructure where appropriate, and accomplishes the task’s purpose
- [x] The tests appropriately test the new code, including edge cases
- [x] You have tried to break the code

If JIRA tests were used to supplement or replace automated tests:
- [x] JIRA tests have been run and pass N/A
- [x] You agree with the justification for use of JIRA tests or have provided input on why you disagree N/A

**Reviewer 2:**

Name:
- [x] Code is maintainable and reusable, reuses existing code and infrastructure where appropriate, and accomplishes the task’s purpose
- [x] The tests appropriately test the new code, including edge cases
- [x] You have tried to break the code

If JIRA tests were used to supplement or replace automated tests:
- [x] JIRA tests have been run and pass N/A
- [x] You agree with the justification for use of JIRA tests or have provided input on why you disagree N/A

PR #972, opened 2018-03-28
Bonnie v2.1 VSAC Refactor

hossenlopp:
Light refactoring of to make use of new VSACAPI class in bonnie_bundler. Primarily refactors the way ticket granting ticket is stored in the session for reuse.

~❗️ Currently points to corresponding bonnie_bundler branch. This should be reverted after bonnie_bundler PR is merged.~
- [x] Gemfile fixed before merge.

**Submitter:**
- [x] JIRA ticket for this PR: https://jira.mitre.org/browse/BONNIE-1339
- [x] If UI changes have been made, google WAVE plug-in has been executed to ensure no 508 issues were introduced. N/A
         1. Note why this could not be done or is not applicable here: N/A
         2. Add TODOs in the code noting that it requires a test N/A
         3. Add a JIRA task to add the test and link it here:  N/A
Before - 1567 / 1784 LOC (87.84%) covered.
After -  1579 / 1791 LOC (88.16%) covered.
- [x] Automated regression test(s) pass N/A Does not affect calculation.

If JIRA tests were used to supplement or replace automated tests: N/A
- [x] JIRA test links: N/A

**Reviewer 1:**

Name:

If JIRA tests were used to supplement or replace automated tests: N/A

**Reviewer 2:**

Name: @losborne

If JIRA tests were used to supplement or replace automated tests: N/A


config/bonnie.yml
```diff
@@ -5,11 +5,12 @@ defaults: &defaults
...
-    ticket_url: https://vsac.nlm.nih.gov/vsac/ws/Ticket
-    api_url: https://vsac.nlm.nih.gov/vsac/svs/RetrieveMultipleValueSets
-    profiles_url: https://vsac.nlm.nih.gov/vsac/profiles
-    profile: Latest eCQM
+  vsac:
+    auth_url: https://vsac.nlm.nih.gov/vsac/ws
+    content_url: https://vsac.nlm.nih.gov/vsac/svs
+    utility_url: https://vsac.nlm.nih.gov/vsac
+    default_profile: Latest eCQM
+    default_release: eCQM Update 2018 EP-EC and EH
```
edeyoung: is there a way to just have it always default to the latest release?

I feel like we should have a default *program* and then always just try to get the latest release from that program. Thoughts?

app/controllers/measures_controller.rb
```diff
@@ -150,7 +150,7 @@ def create
         f.write("Original Filename was #{params[:measure_file].original_filename}\n")
         f.write(e.to_s + "\n" + e.backtrace.join("\n"))
       end
-      if e.is_a? Measures::VSACException
+      if e.is_a?(Measures::VSACException) || e.is_a?(Util::VSAC::VSACError)
```
edeyoung, line 153: does it make sense for `VSACException` to live in `Measures` but the rest to live in `Util::VSAC`?

lib/tasks/vsac.rake
```diff
@@ -23,30 +27,34 @@ namespace :bonnie do
...
-      # TODO: This is a temporary thing. This call is actually calling the profiles url. HDS VSApiV2 
-      # should have a get_profiles call added.
-      result = api.get_valueset("0")
-      
-      doc = Nokogiri::XML(result)
-      profile_list_element = doc.at_xpath("/ProfileList")
-      
-      profile_list_element.xpath("//profile").each do |profile|
-        puts "  #{profile.text}"
+      api = Util::VSAC::VSACAPI.new(config: APP_CONFIG["vsac"])
```
edeyoung: so these functions do more than print things? If so, can you add a comment? If not, are they place holder functions? If so, can you add a comment?

config/bonnie.yml
hossenlopp: Defaulting to the latest release would be work that would happen in the following ticket. https://jira.mitre.org/browse/BONNIE-1297

I do have the API class look for a default program when getting releases, first in this config, second in a constant.

app/controllers/measures_controller.rb
hossenlopp: I don't think it makes sense either. But Measures::VSACException is thrown by the value set parsing code and sometimes isn't directly an issue with VSAC. I didn't want to rock the boat with changes to how the parameters for VSAC are passed and the flow of value sets from the api to database. Removing or renaming Measures::VSACException could be work in the following ticket that does change the flow to handle the release parameter. https://jira.mitre.org/browse/BONNIE-1297

lib/tasks/vsac.rake
hossenlopp: These are just utility rake tasks that exist to play with the VSAC API. They have been useful in the past, so I expanded them to handle the the utility calls. I will add a comment to explain their existence.

config/bonnie.yml
edeyoung: i'm feeling blind - i don't see it in the config. is it not yet populated in the config even though it can be? If that's the case, I'd say populate it here

hossenlopp: I completely forgot to put it in the config. I will fix it. 

This is where it checks the config for a profile, but defaults to the constant in the class.
https://github.com/projecttacoma/bonnie_bundler/pull/140/files#diff-4f7e551733f73d385a07bdd4f118cc77R144

In this config, default_profile could definitely be cleaned up in the next iteration. I think some of the flow at the moment uses it.

test/functional/measures_controller_test.rb
```diff
@@ -420,14 +420,16 @@ class << update_measure_file
     assert_equal true, measure.episode_of_care?
     assert_equal 'eh', measure.type
     assert_nil measure.episode_ids
+    measure_id_before = measure._id
 
     # Update the measure
     VCR.use_cassette("update_response") do
-      post :create, {vsac_date: '09/24/2017', include_draft: false, measure_file: update_measure_file, measure_type: 'eh', calculation_type: 'episode', vsac_username: ENV['VSAC_USERNAME'], vsac_password: ENV['VSAC_PASSWORD']}
+      post :create, {vsac_date: '09/24/2017', include_draft: false, measure_file: update_measure_file, measure_type: 'eh', calculation_type: 'episode', hqmf_set_id: "762B1B52-40BF-4596-B34F-4963188E7FF7", vsac_username: ENV['VSAC_USERNAME'], vsac_password: ENV['VSAC_PASSWORD']}
```
hossenlopp, line 443: It should be noted that this change fixes this test to actually do the update it says it intends to do. On line 432 it adds a check to make sure the measure was actually updated.

There actually is an underlying issue with measure uploads that this test did exploit. The value sets will be updated for a measure but not the measure itself if the measure is uploaded as a new measure.

app/controllers/measures_controller.rb
```diff
@@ -289,29 +302,25 @@ def cql_to_elm
...
-      # 7.5hrs from now
-      if !ticket.nil? && !ticket.empty?
-        session[:tgt] = {ticket: ticket, expires: Time.now + 27000}
-        ticket
+    if ticket_granting_ticket.nil?
+      # Retrieve a new ticket granting ticket by creating the api class.
+      api = Util::VSAC::VSACAPI.new(config: APP_CONFIG['vsac'], username: params[:vsac_username], password: params[:vsac_password])
+      ticket_granting_ticket = api.ticket_granting_ticket
+      if ticket_granting_ticket.nil?
+        return nil # this means no credentials were provided
```
lukoktonos: This line is untested and I believe unreachable via the Bonnie UI.  The 'Load' button is not activated until credentials have been provided.  Should this throw an error instead of returning `nil`?

Trying to set `ticket_granting_ticket` to `nil` in `measures_controller.rb` when loading the measure works because credentials are still there.

Trying to set both `ticket_granting_ticket` and credentials to `nil` results in this error message, which seems wrong:
image

hossenlopp: Good call. I'll look into resolving this...

.rubocop_todo.yml
```diff
@@ -540,7 +540,7 @@ Lint/Void:
 
 # Offense count: 42
 Metrics/AbcSize:
-  Max: 213
+  Max: 213.6
```
lukoktonos, line 543: I think rather than increasing the limits, we should add `measures_controller.rb` as an excluded file for these categories.  The limits were chosen somewhat arbitrarily based on the state of the repo, but it makes sense to draw the line somewhere.

hossenlopp: This appeared to be an auto generated file. That limit matched the exact result for `measures_controller.rb` so I upped it. I do believe I can lower it in the UI and parameter refactor work.

lukoktonos: In https://github.com/projecttacoma/bonnie/pull/962 I originally increased the limits as well but I think we decided that adding the file to the Exclude list was a better route.

hossenlopp: @losborne and I discussed this in person and decided to keep this change for now. Excluding the file will let the MeasuresController#create function become worse without warning. Soon to follow work on VSAC UI changes will likely improve this metric and we will lower this max value to match the improvement.

lukoktonos (approved):
Looks good.  Just needs gemfile update before merge.

jbradl11 (approved):
Changes make sense. Excited to see it being used with the new UI.

hossenlopp:
Gemfile has been sorted for merge to point to latest `bonnie_bundler` `bonnie-v2.1`. Only failing travis errors are loofah and rails sanitizer.

PR #994, opened 2018-05-14
Excel export now uses toString for datatype results

holmesie:
Excel export now uses toString for datatype results. This means that results in the excel export show the data criteria (description, start/stop time/ code) that resulted when a define statement was calculated against a patient.

Example:

screen shot 2018-05-22 at 11 03 36 am

---

**Submitter:**
- [x] JIRA ticket for this PR: https://jira.mitre.org/browse/BONNIE-1442
- [x] Test fixtures updated and documented as necessary ( see [internal wiki](https://gitlab.mitre.org/bonnie/internal-documentation/wikis/testing#test-fixtures) )

Branch | Back End Coverage | Front End Coverage
-- | -- | --
master | 1746 / 1964 LOC (88.9%) covered | screen shot 2018-05-14 at 11 52 44 am
excel_datatype_toString | 1746 / 1964 LOC (88.9%) covered | screen shot 2018-05-24 at 2 58 32 pm

- [x] JIRA test links: N/A

**Reviewer 1:**

Name: @losborne 

**Reviewer 2:**

Name: @mayerm94 

### DIFF
--- a/app/assets/javascripts/helpers/cql_results_helpers.js.coffee
+++ b/app/assets/javascripts/helpers/cql_results_helpers.js.coffee
@@ -100,7 +100,8 @@ class CQLResultsHelpers
   # relevance in determining the result. This is a two level map just like `statement_relevance`. The first level key is
   # the library name and the second key level is the statement name. The value is an object that has three properties,
   # 'raw', 'final' and 'pretty'. 'raw' is the raw result from the execution engine for that statement. 'final' is the final
-  # result that takes into account the relevance in this calculation. 'pretty' is a human readable description of the result.
+  # result that takes into account the relevance in this calculation. 'pretty' is a human readable description of the result
+  # that is only generated if doPretty is true.
   # The value of 'final' will be one of the following strings:
   # 'NA', 'UNHIT', 'TRUE', 'FALSE'.
   #
@@ -151,9 +152,10 @@ class CQLResultsHelpers
   # @param {Measure} measure - The measure.
   # @param {object} rawClauseResults - The raw clause results from the calculation engine.
   # @param {object} statementRelevance - The `statement_relevance` map. Used to determine if they were hit or not.
+  # @param {boolean} doPretty - If true, also generate pretty versions of result.
   # @returns {object} Object with the statement_results and clause_results structures, keyed as such.
   ###
-  @buildStatementAndClauseResults: (measure, rawClauseResults, statementRelevance) ->
+  @buildStatementAndClauseResults: (measure, rawClauseResults, statementRelevance, doPretty = false) ->
     statementResults = {}
     clauseResults = {}
     emptyResultClauses = []
@@ -165,21 +167,21 @@ class CQLResultsHelpers
         statementResults[lib][statementName] = { raw: rawStatementResult}
         if _.indexOf(Thorax.Models.Measure.cqlSkipStatements, statementName) >= 0 || statementRelevance[lib][statementName] == 'NA'
           statementResults[lib][statementName].final = 'NA'
-          statementResults[lib][statementName].pretty = 'NA'
+          statementResults[lib][statementName].pretty = 'NA' if doPretty
         else if statementRelevance[lib][statementName] == 'FALSE' || !rawClauseResults[lib]?
           statementResults[lib][statementName].final = 'UNHIT'
-          statementResults[lib][statementName].pretty = 'UNHIT'
+          statementResults[lib][statementName].pretty = 'UNHIT' if doPretty
         else
           if @_doesResultPass(rawStatementResult)
             statementResults[lib][statementName].final = 'TRUE'
-            statementResults[lib][statementName].pretty = @_prettyResult(rawStatementResult)
+            statementResults[lib][statementName].pretty = @prettyResult(rawStatementResult) if doPretty
           else
             statementResults[lib][statementName].final = 'FALSE'
             if rawStatementResult instanceof Array && rawStatementResult.length == 0
               # Special case, handle empty array.
-              statementResults[lib][statementName].pretty = "FALSE ([])"
+              statementResults[lib][statementName].pretty = "FALSE ([])" if doPretty
             else
-              statementResults[lib][statementName].pretty = "FALSE (#{rawStatementResult})"
+              statementResults[lib][statementName].pretty = "FALSE (#{rawStatementResult})" if doPretty
 
         # create clause results for all localIds in this statement
         localIds = measure.findAllLocalIdsInStatementByName(lib, statementName)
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/views/measure_view.js.coffee
```diff
@@ -169,13 +169,15 @@ class Thorax.Views.Measure extends Thorax.Views.BonnieView
...
+
   # Iterates through the results to remove extraneous fields.
   removeRaw: (results) ->
     ret = {}
     for libKey of results
       ret[libKey] = {}
       for statementKey of results[libKey]
-        ret[libKey][statementKey] = results[libKey][statementKey].final
+        result = results[libKey][statementKey]
+        ret[libKey][statementKey] = if result.raw && result.raw.length > 0 then result.raw.join('\n').replace(/\n$/, '') else result.final
```
lukoktonos: For arrays, joining on `\n` looks pretty clean.  

Another option is to join on `,\n`, or add `[` and `]` (or perhaps `Array{}` to match CQL syntax) surrounding this string in order to be more explicit that an array is being returned.  

I think I prefer it as-is, but putting some examples of each here for discussion:

As-is:
```
Encounter, Performed: Encounter Inpatient
START: 01/01/2012 8:00 AM
STOP: 01/02/2012 8:00 AM
CODE: SNOMED-CT 8715000
Encounter, Performed: Encounter Inpatient
START: 01/01/2012 8:00 AM
STOP: 01/02/2012 8:00 AM
CODE: SNOMED-CT 183452005
```

Comma/newline join:
```
Encounter, Performed: Encounter Inpatient
START: 01/01/2012 8:00 AM
STOP: 01/02/2012 8:00 AM
CODE: SNOMED-CT 8715000,
Encounter, Performed: Encounter Inpatient
START: 01/01/2012 8:00 AM
STOP: 01/02/2012 8:00 AM
CODE: SNOMED-CT 183452005
```

Comma/newline + brackets:
```
[Encounter, Performed: Encounter Inpatient
START: 01/01/2012 8:00 AM
STOP: 01/02/2012 8:00 AM
CODE: SNOMED-CT 8715000,
Encounter, Performed: Encounter Inpatient
START: 01/01/2012 8:00 AM
STOP: 01/02/2012 8:00 AM
CODE: SNOMED-CT 183452005]
```

holmesie: I also prefer as-is. The only alternative I think might be nicer is an empty line between them, but I don't think that's too much of an improvement.

lukoktonos (changes requested):
The current code works for QDM datatypes but not other cql types like Code, Quantity, DateTime, etc.  See below for example output for those:

DateTime (this one somewhat works but doesn't match the JIRA task, which would expect something more similar to the CQL syntax)
image
image

Code
image
image

Diagnoses
image

Quantity
image

Interval
image

I think we should enumerate the supported types and syntax for each that needs to be supported and have tests for each of them.

holmesie:
@losborne 
Agreed about Interval, DateTime, and Code. I'll have to investigate a way to deal with those in a different way, because overriding toString is not an option for DateTime.

I think Quantity is fine as is.

I'm confused what you are doing with Diagnoses, it looks you constructed an artificial raw? Those look like the internal Bonnie diagnoses, not the actual Diagnosis<*> type from the cql_qdm_patientapi.

lukoktonos:
@holmesie Ah right, here is the right Diagnosis result:

image

holmesie:
@losborne Okay, I see.

Is that a special type? Looks like just an array of codes.

lukoktonos:
@holmesie It is an array of Codes--I'm not sure where `CQL_QDM.Diagnosis` would be used, but that extends from Datatype so should be OK.

edeyoung:
Some things that came out of yesterday's sprint burndown:

Have some way to represent groupings. We'll get into differentiating tuples/arrays/etc. at a later time.
Differentiate the different "falsey" types - false, null, empty

app/assets/javascripts/helpers/cql_results_helpers.js.coffee
```diff
@@ -261,3 +272,33 @@ class CQLResultsHelpers
...
       return true
+
+  ###*
+  # Generates a pretty human readable representation of a result.
+  #
+  # @private
+  # @param {(Array|object|boolean|???)} result - The result from the calculation engine.
+  # @returns {String} a pretty version of the given result
+  ###
+  @_prettyResult: (result) ->
```
lukoktonos: This function could use some more coverage for most of the cases.  The current tests only hit `Array` and `CQL_QDM.QDMDatatype`.

lukoktonos (changes requested):
The excel export no longer prints out the datatypes.

Commit 0bde914c:
image

Commit 951fbd14 (latest as of this comment):
image

app/assets/javascripts/helpers/cql_results_helpers.js.coffee
```diff
@@ -261,3 +276,33 @@ class CQLResultsHelpers
...
+      moment.utc(result.toString()).format('MM/DD/YYYY h:mm A')
+    else if result instanceof cql.Interval
+      "Interval: #{@_prettyResult(result['low'])} - #{@_prettyResult(result['high'])}"
+    else if result instanceof cql.Code
+      "Code: #{result['system']}: #{result['code']}"
+    else if result instanceof cql.Quantity
+      "Quantity: #{result['unit']}: #{result['value']}"
+    else if result instanceof CQL_QDM.QDMDatatype
+      result.toString()
+    else if result instanceof String
```
lukoktonos: String literals fall through to `JSON.stringify(result, null, 2)`, which produces this:
image
image

I think there needs to be an `or typeof(result) == 'string'` here.   

It would also be good to add a test to cover this case (it's the only case still uncovered).

lukoktonos (changes requested):

app/assets/javascripts/helpers/cql_results_helpers.js.coffee
holmesie: Good call! I think with that additional `or`, the existing tests should hit that branch. Just need to add another `expect`.

app/assets/javascripts/helpers/cql_results_helpers.js.coffee
holmesie: All lines now covered.

lukoktonos (approved):
Looks good!

app/assets/javascripts/helpers/cql_results_helpers.js.coffee
```diff
@@ -118,21 +121,22 @@ class CQLResultsHelpers
...
-  #     "SDE Ethnicity": { "raw": "???", "final": "NA" },
-  #     "SDE Payer": { "raw": "???", "final": "NA" },
-  #     "SDE Race": { "raw": "???", "final": "NA" },
-  #     "SDE Sex": { "raw": "???", "final": "NA" },
-  #     "Most Recent Delivery": { "raw": "???", "final": "TRUE" },
-  #     "Most Recent Delivery Overlaps Diagnosis": { "raw": "???", "final": "TRUE" },
-  #     "Initial Population": { "raw": "???", "final": "TRUE" },
-  #     "Numerator": { "raw": "???", "final": "TRUE" },
-  #     "Denominator Exceptions": { "raw": "???", "final": "UNHIT" }
+  #     "Patient": { "raw": "???", "final": "NA", "pretty": "NA" },
```
hossenlopp, line 125: Adding this pretty result generation to this function could have some serious performance impacts since this pretty data will be generated on every single calculation. Has the performance impact of this been analyzed? I expect the Opioid measure at least will become considerably slower as a result of this, but I bet the overall initial calculations run on page load will be noticeably affected.

The pretty formatted data isn't needed for every calculation. It is only used for excel export and potential debugging or hybrid measure result views. Perhaps a helper method could be called after calculation, only before the pretty format is needed, that will generate and add it to the existing structure.

mayerm94: I ran profiler on a measure with 3 strats with 44 patients. Master = 4220.5 ms Scripting.  This branch 4830.2 ms scripting.

hossenlopp: Was that for loading an account with that single measure? Or just changing tabs on that measure?

app/assets/javascripts/helpers/cql_results_helpers.js.coffee
hossenlopp: This function should be public so it can be used to generate pretty results for other calculation results on demand.

holmesie: Right

mayerm94 (approved):

PR #1005, opened 2018-05-31
MAT Integration Review

hossenlopp:
This is a PR for reviewing (ask Hoss anything) about the MAT Integration work.


test/controllers/api_v1/measures_controller_test.rb
```diff
@@ -0,0 +1,406 @@
+require 'test_helper'
```
lukoktonos, line 1: There are many controller tests in test/ rather than test/controllers, including a `measure_controller_test.rb`.  Should those be moved into this directory? Should this file be combined with the other `measures_controller_test.rb` at some point?

Gemfile
```diff
@@ -46,7 +46,11 @@ gem 'roo'
 
 gem 'oj' # Faster JSON
 
-gem 'mongoid-history'    # For adding versions
+gem 'mongoid-history'
+gem 'apipie-rails', :git => 'https://github.com/hossenlopp/apipie-rails', :branch => 'master'  # API documentation generation
```
c-monkey, line 50: Is reverting back to a mainline version on our list pre-integrating with the cql version of Bonnie? Feels weird to be using a version forked to your personal github.

hossenlopp: Well... they never pulled in my PR...

hossenlopp: https://github.com/Apipie/apipie-rails/pull/440

test/controllers/api_v1/measures_controller_test.rb
hossenlopp: I don't see said tests in test/. If we move forward continuing with the separate measures controller for the API then we should continue with the separate test file and it should be put in a folder to clearly identify it as an API controller.

hossenlopp: I do agree that this file should be in test/functional/api_v1/.

hossenlopp:
Since this work has been merged into CQL master, im going to close it.

PR #1021, opened 2018-06-20
Merge master into v2.2

c-monkey:
Resolved a few merge conflicts and updated CMS460 fixture and tests.

Pull requests into Bonnie require the following. Submitter and reviewer should :white_check_mark: when done. For items that are not-applicable, note it's not-applicable ("N/A") and :white_check_mark:.

**Submitter:**
- [x] This pull request describes why these changes were made.
- [x] This PR is into the correct branch.
- [x] ~~JIRA ticket for this PR:~~ n/a
- [x] ~~JIRA ticket links to this PR~~ n/a
- [x] Code diff has been done and been reviewed (it **does not** contain: additional white space, not applicable code changes, debug statements, etc.)
- [x] If UI changes have been made, google WAVE plug-in has been executed to ensure no 508 issues were introduced.
- [x] Tests are included and test edge cases
- [x] Tests have been run locally and pass (remember to update Gemfile when applicable)
- [x] Test fixtures updated and documented as necessary ( see [internal wiki](https://gitlab.mitre.org/bonnie/internal-documentation/wikis/testing#test-fixtures) )
- [x] Code coverage has not gone down and all code touched or added is covered. 
     * In rare situations, this may not be possible or applicable to a PR. In those situations:
         1. Note why this could not be done or is not applicable here: 
         2. Add TODOs in the code noting that it requires a test
         3. Add a JIRA task to add the test and link it here: 

Branch | Back End Coverage | Front End Coverage
-- | -- | --
bonnie-v2.2 | 1738 / 1950 LOC (89.13%) | Statements   : 58.94% (7134/12104)<br>Branches     : 42.79% ( 1935/4522 )<br>Functions    : 49.19% ( 1178/2395 )<br>Lines        : 56.81% ( 6408/11279 )
this-pr | 1738 / 1950 LOC (89.13%) covered. | Statements   : 58.97% (7141/12109)<br>Branches     : 42.85% ( 1940/4527 )<br>Functions    : 49.19% ( 1178/2395 )<br>Lines        : 56.85% ( 6415/11284 )

- [x] Automated regression test(s) pass

If JIRA tests were used to supplement or replace automated tests:
- [X] JIRA test links: N/A
- [X] Justification for using JIRA tests: N/A
- [X] JIRA tests have been added to sprint N/A

**Reviewer 1:**

Name: @losborne
- [x] Code is maintainable and reusable, reuses existing code and infrastructure where appropriate, and accomplishes the task’s purpose
[...]

PR #1022, opened 2018-06-20
Revert "Merge master into v2.2"

edeyoung:
Reverts projecttacoma/bonnie#1021


lukoktonos (approved):
State after revert matches state of e37d9bfb on `bonnie-v2.2`.

PR #1208, opened 2019-04-19
Verification function for the convert measures rake task

hackrm:
A validation function was created to verify the measure converter rake task properly converted the measures into CQM format. All attributes were verified in detail except for "cql_libraries" which was only validated based on the number of elements, and "population_sets" which wasn't verified at all.

After running the rake task with the validator on the database collection of CQL measures, there were two validation notes that came up due to value sets. Here are the messages:

Conversion Difference
Measure CMS645v2: Bone density evaluation for patients with prostate cancer and receiving androgen deprivation therapy with _id 5a97f802b848467fb6a92e3c in account edeyoung@mitre.org
--- value_sets --- Is different from CQL measure
Conversion Difference
Measure CMSv0: mesg12 with _id 5bfe3673b84846165cd1d1fe in account rreddy@hcareis.com
--- value_sets --- Is different from CQL measure

**Submitter:**
- [x] JIRA ticket for this PR: https://jira.mitre.org/browse/BONNIE-1837
- [x] If UI changes have been made, google WAVE plug-in has been executed to ensure no 508 issues were introduced. - N/A
- [x] Tests are included and test edge cases - N/A
- [x] Test fixtures updated and documented as necessary ( see [internal wiki](https://gitlab.mitre.org/bonnie/internal-documentation/wikis/testing#test-fixtures) ) - N/A
- [x] Automated regression test(s) pass - N/A

**Reviewer 1:**

Name: @losborne 

**Reviewer 2:**

Name:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -36,7 +36,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/cqm-converter
-  revision: 143cc16fadce48166b6bc074bb01b1d4ecf469e9
+  revision: c1ac4080fe027335ef0ae82a5f8749cadb8b0ef8
   branch: master
   specs:
     cqm-converter (1.0.3)
--- a/lib/tasks/cql.rake
+++ b/lib/tasks/cql.rake
@@ -142,23 +142,26 @@ namespace :bonnie do
           if diff.empty?
             print ".".green
           else
-            fail_count += 1
-            puts "Conversion Difference".yellow
+            puts "\nConversion Difference".yellow
             measure_user = User.find_by(_id: measure[:user_id])
             puts "Measure #{measure.cms_id}: #{measure.title} with id #{measure._id} in account #{measure_user.email}".light_blue
             diff.each_entry do |element|
               puts "--- #{element} --- Is different from CQL measure".light_blue
             end
+            fail_count += 1
           end
+        rescue StandardError => e
+          fail_count += 1
+          puts "\nMeasure  #{measure.title} #{measure.cms_id} with id #{measure._id} failed with message: #{e.message}".red
         rescue ExecJS::ProgramError => e
+          fail_count += 1
           # if there was a conversion failure we should record the resulting failure message with the measure
-          puts "Measure  #{measure.title} #{measure.cms_id} failed with message: #{e.message}".red
+          puts "\nMeasure  #{measure.title} #{measure.cms_id} with id #{measure._id} failed with message: #{e.message}".red
         end
       end
-      puts '**** Done converting ****'
+      puts "\n**** Done converting ****"
       puts "Successful Conversions: #{bonnie_cql_measures.count - fail_count}"
-      puts "Unsuccessful Conversion*: #{fail_count}"
-      puts "* - the converted measure exists, but the data is off in some way from the original CQL measure"
+      puts "Unsuccessful/Failed Conversions: #{fail_count}"
     end
 
     def self.measure_conversion_diff(cql_measure, cqm_measure)
--- a/lib/tasks/cql.rake
+++ b/lib/tasks/cql.rake
@@ -186,7 +186,9 @@ namespace :bonnie do
       calc_method = cql_measure[:episode_of_care] ? 'EPISODE_OF_CARE' : 'PATIENT'
       differences.push('calculation_method') if calc_method != cqm_measure['calculation_method']
 
-      differences.push('source_data_criteria') if cql_measure.source_data_criteria.count != cqm_measure.source_data_criteria.count
+      # Duplicate source data criteria are removed from the measure when converted, so this verification ensures that was done properly
+      unique_sdc = cql_measure.source_data_criteria.values.index_by { |sdc| [sdc['code_list_id'], sdc['description']] }
+      differences.push('source_data_criteria') if unique_sdc.length != cqm_measure.source_data_criteria.count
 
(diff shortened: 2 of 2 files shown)


lib/tasks/cql.rake
```diff
@@ -126,18 +128,83 @@ namespace :bonnie do
...
+        differences['component_hqmf_set_ids'] = 'component hqmf set ids' if Digest::MD5.hexdigest(cql_measure[:component_hqmf_set_ids].to_json) != Digest::MD5.hexdigest(cqm_measure['component_hqmf_set_ids'].to_json)
+      end
+      if cql_measure[:component]
+        differences['component'] = 'component' if !cqm_measure['component']
+        differences['composite_hqmf_set_id'] = 'composite hqmf set id' if Digest::MD5.hexdigest(cql_measure.composite_hqmf_set_id().to_json) != Digest::MD5.hexdigest(cqm_measure['composite_hqmf_set_id'].to_json)
+      end
+
+      cql_measure[:value_set_oid_version_objects].each do |cql_val_set|
+        if cql_val_set[:version] == ""
+          differences['value_sets'] = 'value sets' if (cqm_measure.value_sets.where(oid: cql_val_set[:oid], version: "").count < 1 && cqm_measure.value_sets.where(oid: cql_val_set[:oid], version: "N/A").count < 1)
```
hackrm: During the conversion, if a value_set_oid_version_object had an empty string version, it could either contain an empty string or "N/A" in the converted CQM measure, which is the reason for this if statement structure.

hackrm:
If you want to run the rake task and only convert one of the measures listed in the description, temporarily replace line 131 in cql.rake with one of the following: 
      bonnie_cql_measures = CqlMeasure.where(_id: BSON::ObjectId("5a97f802b848467fb6a92e3c"))
      bonnie_cql_measures = CqlMeasure.where(_id: BSON::ObjectId("5bfe3673b84846165cd1d1fe"))

Gemfile.lock
```diff
@@ -36,7 +36,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/cqm-converter
-  revision: bc3885c3789bc316428bd47ba2c828c134e166ab
+  revision: 24d197e354bcfe6dcf94f47dbed50144b4dc1f8f
```
lukoktonos: This commit is behind master, should update and possibly need to verify the changes contained in the most recent commit(s)

lib/tasks/cql.rake
```diff
@@ -126,18 +128,83 @@ namespace :bonnie do
...
       bonnie_cql_measures.each do |measure|
         begin
           cqm_measure = CQM::Converter::BonnieMeasure.to_cqm(measure)
           cqm_measure.value_sets.map{ |value_set| value_set.save!}
           cqm_measure.user = measure.user
           cqm_measure.save!
-          puts measure.title + ' ' +  measure.cms_id
+          # Verify Measure was converted properly
+          diff = measure_conversion_diff(measure, cqm_measure)
+          unless diff.empty?
```
lukoktonos: What about adding a success message if there are no errors found and a summary at the bottom?

Currently it looks like this:
```
**** Starting to convert measures! ****

**** Done converting ****
```

What about:
```
**** Starting to convert measures! ****
lwosborne@mitre.org CMS123 converted successfully (green text)
...
**** Done converting ****
     Successful Conversions: 42
     Unsuccessful Conversions: 2
```

lib/tasks/cql.rake
```diff
@@ -126,18 +128,83 @@ namespace :bonnie do
...
+          puts "Measure  #{measure.title} #{measure.cms_id} failed with message: #{e.message}".red
+        end
+      end
+      puts '**** Done converting ****'
+    end
+
+    def self.measure_conversion_diff(cql_measure, cqm_measure)
+      differences = {}
+
+      differences['user id'] = 'user_id' if Digest::MD5.hexdigest(cql_measure.user_id.to_json) != Digest::MD5.hexdigest(cqm_measure['user_id'].to_json)
```
lukoktonos: Since we are only using each of these values once, doing a `Digest::MD5` increases the runtime rather than speeding things up.  If we were going to be re-using the strings for multiple comparisons, then it might be faster to convert once (O(n) for hashing) and compare multiple times (O(1) for hash comparison) rather than doing multiple string comparisons (O(n) for each string comparison).  As it is now, these `Digest::MD5`s seem like they can be removed to both simplify the code and speed it up slightly.

We also don't need the `to_json` for these either, we can just do
```
differences['hqmf id'] = 'hqmf_id' if cql_measure.hqmf_id != cqm_measure['hqmf_id']
```

Finally, instead of using a hash with identical key/values, you can just use an array to store `differences`.  The values are not used in the printout loop above:
```
            diff.each_key do |key|
              puts "--- #{key} --- Is different from CQL measure".light_blue
            end
```

So the final version can look something like:
```
differences = []
...
differences.push 'hqmf_id' if cql_measure.hqmf_id != cqm_measure['hqmf_id']
```

lukoktonos (changes requested):

Gemfile.lock
hackrm: This has been addressed and will push the update by tomorrow. This PR is also a number of commits behind for it's cqm-models reference so that will be updated too.

lukoktonos (approved):
Looks good!

jbradl11 (approved):

PR #1218, opened 2019-05-10
updated rake task

zlister:
Updates convert patient rake task to prevent it from stopping due to errors found.


lukoktonos (approved):
looks good

PR #1242, opened 2019-06-07
Shift years

zlister:
This PR is adding the `shift_years` function so that the dates on the patients will be adjusted according to the new measurement period.

- [x] Update cqm-models in the Gemfile once merged

**Submitter:**
- [x] JIRA ticket for this PR: https://jira.mitre.org/browse/BONNIE-1848
- [x] Test fixtures updated and documented as necessary ( see [internal wiki](https://gitlab.mitre.org/bonnie/internal-documentation/wikis/testing#test-fixtures) )

**Reviewer 1:**

Name:

**Reviewer 2:**

Name: @losborne 

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -49,7 +49,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/cqm-models
-  revision: ef59618fa137ba6fa27c057b2a57cea4977ba0fa
+  revision: 2688b71f537e97d4b9b126f453816b0c7fffda60
   branch: master
   specs:
     cqm-models (1.1.1.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -49,8 +49,8 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/cqm-models
-  revision: 469c6b18bf2c8a1f2024a6f2daa0ef44c1b2cce0
-  branch: nil_check
+  revision: 4a4f6c7f371eea55d521354963a1bf661a2216f2
+  branch: master
   specs:
     cqm-models (1.1.1.0)
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -49,7 +49,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/cqm-models
-  revision: 2688b71f537e97d4b9b126f453816b0c7fffda60
+  revision: ef59618fa137ba6fa27c057b2a57cea4977ba0fa
   branch: master
   specs:
     cqm-models (1.1.1.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -49,8 +49,8 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/cqm-models
-  revision: 2688b71f537e97d4b9b126f453816b0c7fffda60
-  branch: master
+  revision: 469c6b18bf2c8a1f2024a6f2daa0ef44c1b2cce0
+  branch: nil_check
   specs:
     cqm-models (1.1.1.0)
 
--- a/app/controllers/measures_controller.rb
+++ b/app/controllers/measures_controller.rb
@@ -128,20 +128,23 @@ class MeasuresController < ApplicationController
   def shift_years(measure, year_shift)
     # Copy the patients to make sure there are no errors before saving every patient
     patients = CQM::Patient.by_user_and_hqmf_set_id(current_user, measure.hqmf_set_id).all.entries
-    begin
       patients.each do |patient|
-        patient_birthdate_year = patient.qdmPatient.birthDatetime.year
-        patient.qdmPatient.birthDatetime.change(year: year_shift + patient_birthdate_year)
-        patient.qdmPatient.dataElements.each do |data_element|
-          data_element.shift_years(year_shift)
+        begin
+          patient_birthdate_year = patient.qdmPatient.birthDatetime.year
+          if year_shift + patient_birthdate_year > 9999 || year_shift + patient_birthdate_year < 0001
+            raise RangeError
+          end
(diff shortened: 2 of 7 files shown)


app/controllers/measures_controller.rb
```diff
@@ -117,4 +120,12 @@ def retrieve_measure_details(params)
     }
   end
 
+  def shift_years(measure, year)
+    CQM::Patient.by_user_and_hqmf_set_id(current_user, measure.hqmf_set_id).each do |patient|
+      patient.qdmPatient.dataElements.each do |data_element|
```
lukoktonos: In addition to this loop, we should also update the birthDatetime of the qdmpatient (which is in the data elements array but also outside of it:

image

test/functional/measures_controller_test.rb
```diff
@@ -760,15 +760,20 @@ class << update_measure_file
...
       year: '1984',
-      id: measure.id.to_s
+      id: measure.id.to_s,
+      measurement_period_shift_dates: "true"
     }
     measure = CQM::Measure.where(id: measure_id).first
     assert_equal '1984', measure.measure_period['low']['value'].slice(0,4)
+    patient = CQM::Patient.by_user(@user).first
+    assert_equal 1984, patient.qdmPatient.dataElements.first.authorDatetime.year
+    assert_equal 1984, patient.qdmPatient.dataElements.first.relevantPeriod.low.year
```
lukoktonos: Should add a test where the data element had a year other than 2012 originally and check that it has shifted appropriately by the correct number of years

One edge case that is unlikely but worth considering is if the shift in dates would produce an invalid CQL Year,
for example if a data element has year 0001 but then the MP is changed from 2012 to 2011.  We might want to throw an error in that case

lukoktonos (changes requested):

app/controllers/measures_controller.rb
jbradl11: In addition to this, would a death Datetime be handled in a similar way?

lib/tasks/fixture_helper.rb
```diff
@@ -45,12 +45,27 @@ def load_fixture_file(file, collection_name, user = nil)
...
   return nil unless json.is_a?(Hash)
   json.each_pair do |k,v|
-    json[k] = Time.parse(v) if k.ends_with?("_at")
+    if v.is_a?(Array)
+      v.each do |val| 
+        if val.is_a?(Hash)
+          convert_times(val)
+        elsif val.is_a?(String)
+          val.to_datetime if val.match?(/\d{4}-\d{2}-\d{2}T/)
+        end
```
jbradl11, line 60: Is there a possibility of an Array inside of an Array?

app/controllers/measures_controller.rb
lukoktonos: Yes, good point--the users said that they would want all dates on the patient shifted.

zlister: Is the death Datetime a data element? It might already be handled then

lib/tasks/fixture_helper.rb
zlister: I haven't seen one, usually it is an array of Hashes (an array containing other classes tend to be a Hash). I can add another check just so that it is covered in case though

app/controllers/measures_controller.rb
zlister: `expiredDatetime` is a data_element

lukoktonos (commented):
It looks like CQM Patients' QDM Patients have both a `dataElements` array as well as `extendedData.source_data_criteria`.  I'm not sure what the role of the latter is--should that even be on the qdm patient?  If so, I think we may need to adjust those dates as well, as they have start/end dates.

It does not look like the qdmpatient's birthdate is being updated with the change in measurementperiod.  (the one in dataElements is, but not the other one.  The date picker in the patient builder view doesnt show a different birthdate)

It would be better for the error upon failure to shift an element (due to <1 or >9999) to give some information about which element is causing the issue if possible:
```Element on Patient could not be shifted. Please make sure shift will keep all years between 1 and 9999```

I think it may make sense to hold off on merging this PR until the patient saving and patient builder timing section get in, because those allow for more comprehensive manual testing.

File: Gemfile.lock [hunk 4]
```diff
@@ -49,7 +49,7 @@ GIT
 
 GIT
   remote: https://github.com/projecttacoma/cqm-models
-  revision: 2688b71f537e97d4b9b126f453816b0c7fffda60
+  revision: ef59618fa137ba6fa27c057b2a57cea4977ba0fa
```
lukoktonos: `2688b71` is the most up-to-date, `ef59618f` is 1 commit behind now

lib/tasks/fixture_helper.rb
```diff
@@ -45,12 +45,36 @@ def load_fixture_file(file, collection_name, user = nil)
...
+# Some date fields can be nested so that is why there is recursion
 def convert_times(json)
   return nil unless json.is_a?(Hash)
   json.each_pair do |k,v|
-    json[k] = Time.parse(v) if k.ends_with?("_at")
+    if v.is_a?(Array)
+      v.each do |val|
+        if val.is_a?(Hash)
+          vals = [val]
+          vals.each { |vall| convert_times(vall) }
```
jbradl11: You should be able to just to do this
```
if val.is_a?(Hash)
  convert_times(val)
```
Since you have a base check for `Hash` this will be handled.

lib/tasks/fixture_helper.rb
```diff
@@ -45,12 +45,36 @@ def load_fixture_file(file, collection_name, user = nil)
...
+        elsif val.is_a?(String)
+          val.to_datetime if val.match?(/\d{4}-\d{2}-\d{2}T/)
+        elsif val.is_a?(Array)
+          val.each do |value|
```
jbradl11: Similarly to the previous comment. Since you have a base check for `Array` you can just pass val back into the function.

lukoktonos (approved):
Looks good!

jbradl11 (approved):

PR #1336, opened 2019-10-01
Bumping bonnie production version from 4.0.0. to 4.0.1

AndrewBird81:
Just increasing the version number

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bonnie",
-  "version": "3.1.0",
+  "version": "4.0.1",
   "description": "A tool for testing electronic clinical quality measures",
   "main": "index.js",
   "repository": "git@github.com:projecttacoma/bonnie.git",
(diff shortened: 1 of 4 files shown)


lukoktonos (commented):
It looks like `package.json` version could be updated as well (though that's not really used for anything currently)

lukoktonos (dismissed):

ferret1964 (dismissed):

adongare (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bonnie",
-  "version": "3.1.0",
+  "version": "4.0.1",
```
adongare, line 3: is it necessary to update CURRENT VERSION and CAPABILITIES Vxx version in show  _show.html.erb_?

jbradl11: In the past we haven't listed out the 3rd digit on the `show.html.erb`, so I don't have an opinion. I Believe users don't usually care about the minor version update.

jbradl11 (approved):

PR #1337, opened 2019-10-03
Bumping the version of bonnie-prior

AndrewBird81:
Just bumping the version of bonnie-prior from 3.2.1 to 3.2.2


jbradl11:
Looks like the build is failing.

PR #1338, opened 2019-10-04
Replacing hard-coded email address with configuration value equivalent

AndrewBird81:
This work was lost when we fixed/replaced the bonnie-prior branch earlier


lukoktonos (approved):
Looks good--travis just failing due to bundle-audit (unrelated to this change)

adongare (approved):

AndrewBird81:
The build is failing because of an unrelated security vulnerability. We will address that in a separate release.

PR #1950, opened 2021-07-19
Add Additional Attributes to Procedure Resource: reasonCode

greenemcg:


app/assets/javascripts/helpers/fhir_value_sets.js.coffee
```diff
@@ -636,7 +636,8061 @@
       }
     }
 
-
+  # Description   https://hl7.org/fhir/R4/valueset-procedure-reason.html
+  @PROCEDURE_REASON_VS =
```
ghost: If possible I would suggest moving the value set into a separate file. fhir_value_sets is too big to maintain. Examples can be seen in **helpers/value_sets** files

greenemcg: I will try also from a UI perspective having 2000 items in a list that is not sorted is a joke

greenemcg: Can u tell me what file or do i have to look through all 22 of them

ghost: Might make sense to trim the list in that case. Users can enter a custom code anyway.

greenemcg: that would have to be a separate story by sone one knowing which codes are more important than others.

adongare (approved):

